PR #1, opened 2016-04-07
usdpaa-apps: fix broken clean task

ikoveshnikov:
usdpaa and usdpaa-apps depends on linux kernel and need to be
rebuilt after it. usdpaa-apps build system cannot complete
'make clean' target, so let bitbake handle clean task in order
to avoid need to manually clean usdpaa-apps recipe on every
kernel rebuild

Signed-off-by: Ivan Koveshnikov <ikoveshnik@gmail.com>


otavio:
I agree with the issue and it is fine  for merging, however we need to follow the procedure so please send it to the mailing list.

ikoveshnikov:
Ok, thank you. Just have sent to the mail list

PR #9, opened 2018-06-29
nxp 4.9.88 2.0.0 without mx8

otavio:


recipes-graphics/wayland/weston_4.0.0.imx.bb
```diff
@@ -84,6 +85,8 @@ PACKAGECONFIG[colord] = "--enable-colord,--disable-colord,colord"
 PACKAGECONFIG[clients] = "--enable-clients --enable-simple-clients --enable-demo-clients-install,--disable-clients --disable-simple-clients"
 # Weston with PAM support
 PACKAGECONFIG[pam] = "--with-pam,--without-pam,libpam"
+# Weston with i.MX G2D renderer
+PACKAGECONFIG[imxg2d] = "--enable-imxg2d,--disable-imxg2d,virtual/libg2d"
```
gibsson: Is this only working with imxgpu2d or any g2d provider?

recipes-graphics/wayland/weston_4.0.0.imx.bb
```diff
@@ -19,8 +19,10 @@ S = "${WORKDIR}/git"
 UPSTREAM_CHECK_URI = "https://wayland.freedesktop.org/releases.html"
 
 inherit autotools pkgconfig useradd distro_features_check
-# depends on virtual/egl
-REQUIRED_DISTRO_FEATURES = "opengl"
+# Disable OpenGL for parts with GPU support for 2D but not 3D
+REQUIRED_DISTRO_FEATURES          = "opengl"
+REQUIRED_DISTRO_FEATURES_imxgpu2d = ""
+REQUIRED_DISTRO_FEATURES_imxgpu3d = "opengl"
```
gibsson: Is this working for imx7 which doesn't have any GPU?
I'm asking because previous patch adds Weston for all packages:
https://github.com/Freescale/meta-freescale/pull/9/commits/52364a8346dc008489442b750020ba43da3774c8

recipes-graphics/clutter/clutter-1.0_%.bbappend
```diff
@@ -1,5 +1,5 @@
-PACKAGECONFIG_imxgpu2d ??= " \
+PACKAGECONFIG_imxgpu3d ??= " \
```
gibsson: Can you explain why we should use it with gpu3d instead of gpu2d before?
Are both working? If so, can't we have both options available with 1 preferred option?

recipes-graphics/cogl/cogl-1.0_%.bbappend
```diff
@@ -1,5 +1,5 @@
-PACKAGECONFIG_imxgpu2d ??= " \
+PACKAGECONFIG_imxgpu3d ??= " \
```
gibsson: Same question as clutter

recipes-graphics/wayland/weston_4.0.0.imx.bb
otavio: From what I understand, it works with any G2D provider.

recipes-graphics/wayland/weston_4.0.0.imx.bb
otavio: Yes. Using software rendering I think.

recipes-graphics/clutter/clutter-1.0_%.bbappend
otavio: I think it ought to use imxgpu which implies it has any GPU. Will rework.

recipes-graphics/cogl/cogl-1.0_%.bbappend
otavio: Same as previous question.

recipes-graphics/wayland/weston_4.0.0.imx.bb
gibsson: So shouldn't it be something like:
PACKAGECONFIG_append_virtual/libg2d = " imxg2d"
(the syntax is most likely wrong)

gibsson: Or just adding something like:
PACKAGECONFIG_append_imxdpu = " imxg2d"

recipes-graphics/wayland/weston_4.0.0.imx.bb
gibsson: Who provides opengl then? Is it using the mesa swrast?

recipes-graphics/wayland/weston_4.0.0.imx.bb
otavio: The syntax adds support for it but it does not /enable/ it.

recipes-graphics/clutter/clutter-1.0_%.bbappend
gibsson: Clutter now depends on imxgpu3d since imxgpu2d is no longer used as "any GPU" selector.
See following comment for details:
imx-base.inc: Add imxgpu to MACHINEOVERRIDES_EXTENDER
https://github.com/OSSystems/meta-freescale/commit/98c1cde6beb7183f84ef2e247e1e2b5ba14a1ede

otavio:
Initial patchset merged.

PR #15, opened 2018-08-31
L4.9.88 2.2.0 v5

thochstein:


recipes-graphics/imx-gpu-g2d/imx-gpu-g2d_6.2.4.p1.6.bb
```diff
@@ -35,4 +35,5 @@ FILES_${PN} = "${libdir}/libg2d* /opt"
 FILES_${PN}-dev = "${includedir}"
 INSANE_SKIP_${PN} = "ldflags"
 
-COMPATIBLE_MACHINE = "(mx6|mx7ulp)"
+COMPATIBLE_MACHINE = "(^$)"
+COMPATIBLE_MACHINE_imxgpu2d = "(mx6|mx7)"
```
otavio: Why change it? it adds an indirection and it is harder to read.

recipes-graphics/wayland/wayland-protocols_1.13.bb
```diff
@@ -0,0 +1,20 @@
+SUMMARY = "Collection of additional Wayland protocols"
```
otavio: Please use .imx as suffixes for the version. This allows to spot it is specific to the imx support.

conf/machine/include/imx-base.inc
```diff
@@ -281,6 +281,7 @@ SERIAL_CONSOLE = "115200 ttymxc0"
 SERIAL_CONSOLE_mxs = "115200 ttyAMA0"
 
 KERNEL_IMAGETYPE = "zImage"
+KERNEL_IMAGETYPE_mx8 = "Image"
```
otavio: Split this on a new commit.

recipes-kernel/linux/linux-imx_4.9.88.bb
```diff
@@ -11,10 +11,14 @@ require recipes-kernel/linux/linux-imx.inc
...
+SRCBRANCH = "imx_4.9.88_imx8qxp_beta2"
+LOCALVERSION = "-${SRCBRANCH}"
+SRCREV = "05f46d3f9b4a8987edac62a52528d52f47cfa42d"
 
 DEFAULT_PREFERENCE = "1"
 
-COMPATIBLE_MACHINE = "(mx6|mx7)"
+EXTRA_OEMAKE_append_mx6 = " ARCH=arm"
+EXTRA_OEMAKE_append_mx7 = " ARCH=arm"
+EXTRA_OEMAKE_append_mx8 = " ARCH=arm64"
```
otavio: Why? This is handled automatically by the system.

conf/machine/imx8mqevk.conf
```diff
@@ -0,0 +1,53 @@
+#@TYPE: Machine
```
otavio: One commit per machine file, please.

recipes-graphics/imx-dpu-g2d/imx-dpu-g2d_1.4.6.bb
```diff
@@ -34,4 +34,5 @@ INSANE_SKIP_${PN} += "ldflags"
 FILES_${PN} = "${libdir}/libg2d* /opt"
 FILES_${PN}-dev = "${libdir}/libg2d${SOLIBSDEV} ${includedir}"
 
-COMPATIBLE_MACHINE = "(mx8mm)"
+COMPATIBLE_MACHINE = "(^$)"
+COMPATIBLE_MACHINE_imxdpu = "${MACHINE}"
```
otavio: You can write this as:

```
COMPATIBLE_MACHINE = "(imxdpu)"
```

recipes-graphics/imx-gpu-viv/imx-gpu-viv-6.inc
```diff
@@ -131,8 +133,11 @@ PACKAGE_FP_TYPE = "hardfp"
 HAS_GBM     = "false"
 HAS_GBM_mx8 = "true"
 
-GLES3_HEADER_REMOVALS       = "gl31.h gl32.h"
-GLES3_HEADER_REMOVALS_mx8mq = "gl32.h"
+# GLES 3.1 is supported for i.MX 8 Series only
+# GLES 3.2 is supported for i.MX 8 Family only
```
otavio: The commit log is confusing. 

I think you could instead set it for each SoC as it would be clearer and easier to understand.

recipes-graphics/imx-gpu-g2d/imx-gpu-g2d_6.2.4.p1.6.bb
otavio: I see. In this case, we have two options:

1. list all compatible SoCs
2. use the imxgpu2d override feature. 

If we take the second, which is more or less what you did, it can be simplified as:

```
COMPATIBLE_MACHINE = "(imxgpu2d)"
```

thochstein: Right, good idea.

thochstein:
I don't have a good way to retrieve Carol's emailed patches because I only have Outlook and Carol was not in today. I saved the files from Outlook, but found there was much manual updating in order to remove changes introduced by Outlook. After doing this for a while, I realized I could be introducing bugs, so best to wait for the patch files. If you can zip them and send them to me I will apply them to the fork.

otavio:
> I don't have a good way to retrieve Carol's emailed patches because I only have Outlook and Carol was not in today. I saved the files from Outlook, but found there was much manual updating in order to remove changes introduced by Outlook. After doing this for a while, I realized I could be introducing bugs, so best to wait for the patch files. If you can zip them and send them to me I will apply them to the fork

I applied all ones which were good to go. Please rebase this branch on top of current master and see if there are still items from comments to address and please let me know.

thochstein:
I have addressed all of your comments.

classes/fsl-eula.bbclass
```diff
@@ -0,0 +1,29 @@
+# fsl-eula.bbclass defines a common naming scheme used by the
```
otavio: I really dislike this rework of the class. It puts a lot of things hidden and it is hard for me to buy the idea of this extra complexity.

Couldn't we use the class we have now?

recipes-bsp/u-boot/u-boot-imx_2017.03.bb
```diff
@@ -20,5 +20,29 @@ inherit dtc-145 fsl-u-boot-localversion
 
 LOCALVERSION ?= "-${SRCBRANCH}"
 
+BOOT_TOOLS = "imx-boot-tools"
+
+do_deploy_append_mx8mq () {
+    # Deploy the mkimage, u-boot-nodtb.bin and fsl-imx8mq-XX.dtb for mkimage to generate boot binary
```
otavio: Why would someone need to recreate the U-Boot blob from the deploy files? I fail to see the use-case.

This is an important point to be discussed as multiple SoCs already make use of DeviceTree in U-Boot but none has had the need to put them inside the deploy directory. 

Is it possible for you to explain the use-case of it?

classes/fsl-eula.bbclass
thochstein: This is not a rework, this is an extension. The existing class is inherited, and the existing recipes will continue to use the existing class as-is.

The existing class does not handle the need to produce multiple binaries along multiple axes like architecture and graphics backend. Rather than duplicate the code for that in multiple recipes, I chose to capture it in these classes for maximum reuse and actually for maximum hiding, as that is the point of abstraction. All the recipes using these classes are much simpler.

You're not seeing the full usage of the extension, as that is from a demo layer we created:
https://source.codeaurora.org/external/imx/meta-imx-gpu-sdk/tree/recipes-graphics/imx-gpu-sdk/imx-gpu-sdk-tesselation-rle_1.0.0.bb

That all said, there is actually an issue still with this implementation. Given the minimal usage by this upstreaming, it may be in our best interests to withdraw this from consideration for now.

recipes-bsp/u-boot/u-boot-imx_2017.03.bb
thochstein: I was not a part of this design. Can you ask Jun? I assume you see that these artifacts are gathered in the imx-boot compile stage. I had the impression it was something to do about gathering files for the boot partition. I also had the impression that there might be existing mechanisms for accomplishing this instead of this design we have.

classes/fsl-eula.bbclass
thochstein: I'd have to copy these classes into multiple recipes, so I changed my mind about withdrawing this, It would be a lot of work for minimal gain, at the expense of complicating multiple recipes.

thochstein:
Rebased, no other changes.

PR #17, opened 2018-10-05
L4.9.123 2.3.0 Optee

thochstein:


recipes-bsp/imx-atf/imx-atf_1.4.1.bb
```diff
@@ -0,0 +1,65 @@
+# Copyright 2017-2018 NXP
+
+DESCRIPTION = "i.MX ARM Trusted Firmware"
+SECTION = "BSP"
+LICENSE = "BSD-3-Clause"
+LIC_FILES_CHKSUM = "file://${COREBASE}/meta/files/common-licenses/BSD-3-Clause;md5=550794465ba0ec5312d6919e203a55f9"
+
+PV = "1.4.1+git${SRCPV}"
+
+SRCBRANCH = "imx_4.9.88_imx8qxp_beta2"
```
gibsson: Why not using the 4.9.123 2.3.0 release tag (as the series title suggests)?
That would also mean adding SOC_ATF_mx8mm below.

recipes-bsp/u-boot/u-boot-imx_2017.03.bb
```diff
@@ -20,5 +20,29 @@ inherit dtc-145 fsl-u-boot-localversion
```
gibsson: Maybe the title should be "deploy boot image generation tools" as the patch doesn't deplay on "boot partition".

conf/machine/include/imx-base.inc
```diff
@@ -34,7 +34,7 @@ UBOOT_ENTRYPOINT_vf = "0x80008000"
 
 PREFERRED_PROVIDER_virtual/xserver = "xserver-xorg"
 XSERVER_DRIVER                  = "xf86-video-fbdev"
-XSERVER_DRIVER_imxgpu2d         = "xf86-video-imx-vivante"
+XSERVER_DRIVER_imxgpu           = "xf86-video-imx-vivante"
```
gibsson: Hi Tom, I disagree with that change since xf86-video-imx-vivante requires libg2d:
https://github.com/Freescale/meta-freescale/blob/master/recipes-graphics/xorg-driver/xf86-video-imx-vivante_6.2.4.p1.6.bb#L12

gibsson:
Yes I understand it is WIP, I just wanted to give some feedback anyway for the final series ;-)

conf/machine/include/imx-base.inc
gibsson: For instance i.MX8MQ has imxgpu but no g2d provider. But maybe we need to add something about dpu2d.

thochstein: Thanks, I deleted the patch. Note that for i.MX 8, we're not supporting X11 backend.

recipes-bsp/imx-atf/imx-atf_1.4.1.bb
```diff
@@ -0,0 +1,65 @@
...
+    install -m 0644 ${S}/build/${SOC_ATF}/release/bl31.bin ${DEPLOYDIR}/${BOOT_TOOLS}/bl31-${SOC_ATF}.bin
+}
+addtask deploy before do_install after do_compile
+
+do_install () {
+    install -d ${D}/boot
+    install -m 0644 ${S}/build/${SOC_ATF}/release/bl31.bin ${D}/boot/bl31-${SOC_ATF}.bin
+}
+
+FILES_${PN} = "/boot"
```
otavio: It does not make much sense to have a binary package with it as it cannot be loaded from the filesystem. It is needed for boot-stream generation only.

recipes-bsp/u-boot/u-boot-imx_2017.03.bb
```diff
@@ -20,5 +20,29 @@ inherit dtc-145 fsl-u-boot-localversion
...
+    then
+        for config in ${UBOOT_MACHINE}; do
+            i=$(expr $i + 1);
+            for type in ${UBOOT_CONFIG}; do
+                j=$(expr $j + 1);
+                if [ $j -eq $i ]
+                then
+                    install -d ${DEPLOYDIR}/${BOOT_TOOLS}
+                    install -m 0777 ${B}/${config}/arch/arm/dts/${UBOOT_DTB_NAME}  ${DEPLOYDIR}/${BOOT_TOOLS}
+                    install -m 0777 ${B}/${config}/tools/mkimage  ${DEPLOYDIR}/${BOOT_TOOLS}/mkimage_uboot
```
otavio: This should be moved to a `u-boot-imx-mx8boot` or similar, so it can be built as part of native and also installed in SDK.

recipes-bsp/imx-sc-firmware/imx-sc-firmware_0.7.bb
```diff
@@ -0,0 +1,50 @@
...
+SC_FIRMWARE_NAME_mx8qm = "mx8qm-${BOARD_TYPE}-scfw-tcm.bin"
+SC_FIRMWARE_NAME_mx8qxp = "mx8qx-${BOARD_TYPE}-scfw-tcm.bin"
+symlink_name = "scfw_tcm.bin"
+
+SYSROOT_DIRS += "/boot"
+
+do_install () {
+    install -d ${D}/boot
+    install -m 0644 ${S}/${SC_FIRMWARE_NAME} ${D}/boot/
+}
```
otavio: Same as ATF. Seems needed for boot stream, not runtime.

recipes-bsp/imx-atf/imx-atf_1.5.0.bb
```diff
@@ -0,0 +1,57 @@
...
+}
+
+do_deploy () {
+    install -d ${DEPLOYDIR}/${BOOT_TOOLS}
+    install -m 0644 ${S}/build/${SOC_ATF}/release/bl31.bin ${DEPLOYDIR}/${BOOT_TOOLS}/bl31-${SOC_ATF}.bin
+}
+addtask deploy before do_install after do_compile
+
+do_install () {
+}
```
otavio: You can use:

```
do_install[noexec] = "1"
```

So it ensures it is not run.

recipes-bsp/imx-atf/imx-atf_1.5.0.bb
```diff
@@ -0,0 +1,57 @@
...
+inherit pkgconfig deploy
+
+BOOT_TOOLS = "imx-boot-tools"
+
+SOC_ATF ?= "imx8qm"
+SOC_ATF_mx8qm = "imx8qm"
+SOC_ATF_mx8qxp = "imx8qxp"
+SOC_ATF_mx8mq = "imx8mq"
+SOC_ATF_mx8mm = "imx8mm"
+
```
otavio:
```
EXTRA_OEMAKE += " \
    CROSS_COMPILE="${TARGET_PREFIX}" \
    PLAT=${PLATFORM} \
"
```

recipes-bsp/imx-atf/imx-atf_1.5.0.bb
```diff
@@ -0,0 +1,57 @@
...
+do_compile () {
+    export CROSS_COMPILE="${TARGET_PREFIX}"
```
otavio: Drop.

recipes-bsp/imx-atf/imx-atf_1.5.0.bb
```diff
@@ -0,0 +1,57 @@
...
+    cd ${S}
```
otavio: This is the default.

recipes-bsp/imx-atf/imx-atf_1.5.0.bb
```diff
@@ -0,0 +1,57 @@
...
+    echo "-> Build ${SOC_ATF} bl31.bin"
+    # Set BUIL_STRING with the revision info
+    BUILD_STRING=""
+    if [ -e ${S}/.revision ]; then
+        cur_rev=`cat ${S}/.revision`
+        echo " Current revision is ${cur_rev} ."
+        BUILD_STRING="BUILD_STRING=${cur_rev}"
+    else
+        echo " No .revision found! "
+    fi
```
otavio: Couldn't this revision track to be moved to the Makefile?

recipes-bsp/imx-atf/imx-atf_1.5.0.bb
```diff
@@ -0,0 +1,57 @@
...
+
+inherit pkgconfig deploy
+
+BOOT_TOOLS = "imx-boot-tools"
+
+SOC_ATF ?= "imx8qm"
+SOC_ATF_mx8qm = "imx8qm"
+SOC_ATF_mx8qxp = "imx8qxp"
+SOC_ATF_mx8mq = "imx8mq"
+SOC_ATF_mx8mm = "imx8mm"
```
otavio: Please use:

```
PLATFORM ?= "Invalid"
PLATFORM_mx8qm = "imx8qm"
...
```

`SOC_ATF` doesn't communicates nothing useful. Instead, it is misleading. Using the `PLATFORM` allows for immediate recognition what it is about and also it is easy to grasp when looking below when it gets used.

recipes-bsp/imx-atf/imx-atf_1.5.0.bb
```diff
@@ -0,0 +1,57 @@
...
+    # Set BUIL_STRING with the revision info
+    BUILD_STRING=""
+    if [ -e ${S}/.revision ]; then
+        cur_rev=`cat ${S}/.revision`
+        echo " Current revision is ${cur_rev} ."
+        BUILD_STRING="BUILD_STRING=${cur_rev}"
+    else
+        echo " No .revision found! "
+    fi
+    oe_runmake clean PLAT=${SOC_ATF}
```
otavio: The clean is run automatically if rebuilding. So it does not need to be called here.

recipes-bsp/imx-atf/imx-atf_1.5.0.bb
```diff
@@ -0,0 +1,57 @@
...
+    oe_runmake ${BUILD_STRING} PLAT=${SOC_ATF} bl31
```
otavio: `oe_runmake ${BUILD_STRING} bl31`

recipes-bsp/imx-atf/imx-atf_1.5.0.bb
```diff
@@ -0,0 +1,57 @@
...
+    unset CROSS_COMPILE
+}
+
+do_deploy () {
+    install -d ${DEPLOYDIR}/${BOOT_TOOLS}
```
otavio: Drop.

recipes-bsp/imx-atf/imx-atf_1.5.0.bb
```diff
@@ -0,0 +1,57 @@
...
+    install -m 0644 ${S}/build/${SOC_ATF}/release/bl31.bin ${DEPLOYDIR}/${BOOT_TOOLS}/bl31-${SOC_ATF}.bin
```
otavio: `install -Dm 0644 ...`

recipes-bsp/imx-atf/imx-atf_1.5.0.bb
```diff
@@ -0,0 +1,57 @@
...
+}
+addtask deploy before do_install after do_compile
+
+do_install () {
+}
+
+PACKAGE_ARCH = "${MACHINE_ARCH}"
```
otavio: `PACKAGE_ARCH = "${MACHINE_SOCARCH}"`

recipes-bsp/imx-atf/imx-atf_1.5.0.bb
thochstein: Is this really better? One thing I don't like is I am now required to give a destination filename. I ran into this updating imx-sc-firmware. 

`install -Dm 0644 ${S}/${SC_FIRMWARE_NAME} ${DEPLOYDIR}/${BOOT_TOOLS}/${SC_FIRMWARE_NAME}`

Forcing a destination introduces an error path if my destination has a typo.

recipes-bsp/imx-atf/imx-atf_1.5.0.bb
```diff
@@ -0,0 +1,57 @@
...
+
+PV = "1.5.0+git${SRCPV}"
+
+SRCBRANCH = "imx_4.9.123_imx8mm_ga"
+SRC_URI = "git://source.codeaurora.org/external/imx/imx-atf.git;protocol=https;branch=${SRCBRANCH}"
+SRCREV = "af3554fc6ea8c17710ffd37a14a25834fad19271"
+
+S = "${WORKDIR}/git"
+
+inherit pkgconfig deploy
```
thochstein: Drop pkgconfig.

recipes-bsp/imx-atf/imx-atf_1.5.0.bb
thochstein: Done.

otavio:
@thochstein please rebase this and split another PR from this, so we can keep doing progress reducing the number of patches.

otavio:
Please rebase this as many changes are now merged :-)

conf/layer.conf
```diff
@@ -48,3 +48,10 @@ BBFILES_DYNAMIC += " \
     virtualization-layer:${LAYERDIR}/dynamic-layers/virtualization-layer/*/*/*.bb \
     virtualization-layer:${LAYERDIR}/dynamic-layers/virtualization-layer/*/*/*.bbappend \
 "
+
+# Enable optee for all builds using this layer, assuming the machine supports it.
+# To remove optee from the build, add the following line to local.conf:
+# DISTRO_FEATURES_remove = "optee"
+# DEPRECATED: The ability to remove optee from the build is deprecated and
+# will be removed in some future release.
+DISTRO_FEATURES_append = " optee"
```
otavio: Nack! This is something the distro needs to do.

recipes-bsp/imx-mkimage/imx-boot_0.2.bb
```diff
@@ -121,7 +125,9 @@ compile_mx8x() {
 }
 do_compile() {
     compile_${SOC_FAMILY}
-    # mkimage for i.MX8
+    if ${DEPLOY_OPTEE}; then
+        cp ${DEPLOY_DIR_IMAGE}/tee.bin ${BOOT_STAGING}
+    fi
```
otavio: This should be done on do_install.

recipes-bsp/imx-mkimage/imx-boot_0.2.bb
```diff
@@ -66,6 +68,8 @@ SOC_TARGET_mx8qxp = "iMX8QX"
 SOC_TARGET_mx8mq  = "iMX8M"
 SOC_TARGET_mx8mm  = "iMX8MM"
 
+DEPLOY_OPTEE = "${@bb.utils.contains('COMBINED_FEATURES', 'optee', 'true', 'false', d)}"
+
```
otavio: Not need. You can check if the file exists and copy it if it does.

thochstein: Consider the negative case where optee is expected but missing. It will be better to have the recipe fail instead of being silent, right?

recipes-bsp/imx-mkimage/imx-boot_0.2.bb
thochstein: I guess this includes all of the compile_mx* functions as well?

otavio: compile is fine to be on compile, but installation should be on the other task.

recipes-bsp/imx-mkimage/imx-boot_0.2.bb
thochstein: We use a distro feature here to allow the user to decide if he wants optee. This is the point of COMBINED_FEATURES. It is perfectly valid for the machine to support optee but the user to disable it.
My point is about the case where the machine supports optee, and the user wants it, but the file is missing. This should cause the recipe to fail.

otavio: This is not the use case of a distro feature. Instead, user can do:

```
MACHINE_FEATURES_append = " optee"
```

and enable it when desired.

thochstein: That's not what the mega-manual says. Here's the MACHINE_FEATURES definition, emphasis mine: Specifies the list of hardware features the MACHINE is _capable_ of supporting. For related information on _enabling_ features, see the DISTRO_FEATURES, COMBINED_FEATURES, and IMAGE_FEATURES variables.

thochstein: And from the COMBINED_FEATURES definition: For example, the "bluetooth" feature requires hardware support but should also be optional at the distribution level, in case the hardware supports Bluetooth but you do not ever intend to use it.

otavio: Yes but this is intended for use with things which has to enable support in multiple components as for example OpenGL, ALSA or as you mentioned Bluetooth. 

All those need to change all the system and add/remove build flags and dependencies. The `optee` specifically is a machine-only change and I will not pollute the `DISTRO_FEATURES` namespace with something which is i.MX specific.

The possible use of it inside the `MACHINE_FEATURES` allows for easy addition of it while leaves the decision to the user.

thochstein:
There is another part to this that I still don't know how to integrate:

https://source.codeaurora.org/external/imx/meta-fsl-bsp-release/commit/imx/meta-bsp/classes/image_types_fsl.bbclass?h=rocko-4.9.123-2.3.0_8mm_ga&id=ee58cb466a85dcfe3357849a7aed67565dc8f1ec

And for that matter, there are other additions in image_types_fsl.bbclass that I don't think are represented with the new wic-based implementation.

recipes-bsp/imx-mkimage/imx-boot_0.2.bb
thochstein: The copying needs to be done here since the make is done with BOOT_STAGING as input. All the copying is part of setting up the compilation.

recipes-bsp/imx-mkimage/imx-boot_0.2.bb
thochstein: imx-atf, imx-boot, and image_types_fsl.bbclass (which I haven't pushed yet) are the multiple components that need to enable support for optee.

Your suggestion uses MACHINE_FEATURES in an undocumented way. Is your use model missing from the manual?

As it is documented, MACHINE_FEATURES is an invariant. The MACHINE supports a feature or it doesn't. If the user needs to be able to enable it, then DISTRO_FEATURES and COMBINED_FEATURES come into play. Seems like a nice, clear use model. And this is what is documented. What is wrong with it?

otavio: Yes, all those components are meta-freescale ones. A distribution goes beyond the i.MX. If I am using an Intel or a Marvell machine, does the `optee` makes sense? of course not. So it is not a `DISTRO_FEATURE` but a `BSP_FEATURE` which does not exists. The closest is `MACHINE_FEATURES`.

thochstein: But I think what's important is that optee itself is not an i.MX-specific concept. Others could use it to control their particular implementation as well, whatever the distro, whatever the machine.

For Marvell and Intel, the optee distro feature would be fine because it just says that the user would like it if available. Without the machine support of course the image won't have it. Not only does it make sense, it's the main point of COMBINED_FEATURES.

otavio: Sure, but then it must be sent to OE-Core as this is where the generic distro features belongs.

thochstein: Okay, I'm still not convinced, but I'm not going to argue any more. I will remove the usage of COMBINED_FEATURES.

thochstein:
This is ready.

conf/machine/imx6qdlsabreauto.conf
```diff
@@ -42,4 +42,4 @@ SERIAL_CONSOLES = "115200;ttymxc3"
 
 MACHINE_FIRMWARE_append_mx6 = " linux-firmware-ath6k"
 
-MACHINE_FEATURES += " pci wifi bluetooth"
+MACHINE_FEATURES += "pci wifi bluetooth optee"
```
otavio: I wonder if this works when using the community distribution due to the difference between the U-Boot in use. Did you try it?

otavio: Please move the commit which enables the `optee` as the last commit so it does not breaks the tree on the commits in the PR.

recipes-security/optee/optee-client-imx_git.bb
```diff
@@ -0,0 +1,54 @@
...
+	install -D -p -m0755 ${S}/out/export/bin/tee-supplicant ${D}${bindir}/tee-supplicant
+
+	cp -a ${S}/out/export/include ${D}/usr/
+
+	sed -i -e s:/etc:${sysconfdir}:g -e s:/usr/bin:${bindir}:g ${WORKDIR}/tee-supplicant.service
+	install -D -p -m0644 ${WORKDIR}/tee-supplicant.service ${D}${systemd_system_unitdir}/tee-supplicant.service
+}
+
+PACKAGES += "tee-supplicant"
+FILES_${PN} += "${libdir}/* ${includedir}/*"
```
otavio: `includedir` belongs to the `-dev` package.

recipes-security/optee/optee-client-imx_git.bb
```diff
@@ -0,0 +1,54 @@
...
+	install -D -p -m0644 ${S}/out/export/lib/libteec.so.1.0 ${D}${libdir}/libteec.so.1.0
+	ln -sf libteec.so.1.0 ${D}${libdir}/libteec.so
+	ln -sf libteec.so.1.0 ${D}${libdir}/libteec.so.1
+
+	install -D -p -m0755 ${S}/out/export/bin/tee-supplicant ${D}${bindir}/tee-supplicant
+
+	cp -a ${S}/out/export/include ${D}/usr/
+
+	sed -i -e s:/etc:${sysconfdir}:g -e s:/usr/bin:${bindir}:g ${WORKDIR}/tee-supplicant.service
+	install -D -p -m0644 ${WORKDIR}/tee-supplicant.service ${D}${systemd_system_unitdir}/tee-supplicant.service
```
otavio: Could a init.d version of the service to be made? I know NXP uses systemd by default on their distro but many users don't due to a number of reasons...

recipes-security/optee/optee-os-imx_git.bb
```diff
@@ -0,0 +1,84 @@
+# Copyright (C) 2017-2018 NXP
+
+SUMMARY = "OPTEE OS"
+DESCRIPTION = "OPTEE OS"
+HOMEPAGE = "http://www.optee.org/"
+LICENSE = "BSD"
+LIC_FILES_CHKSUM = "file://LICENSE;md5=69663ab153298557a59c67a60a743e5b"
+
+inherit deploy pythonnative
+DEPENDS = "python-pycrypto-native u-boot-mkimage-native"
```
otavio: `DEPENDS` goes on the previous block of code.

recipes-security/optee/optee-os-imx_git.bb
```diff
@@ -0,0 +1,84 @@
...
+
+SRCBRANCH = "imx_4.9.123_imx8mm_ga"
+OPTEE_OS_SRC ?= "git://source.codeaurora.org/external/imx/imx-optee-os.git;protocol=https"
+SRC_URI = "${OPTEE_OS_SRC};branch=${SRCBRANCH}"
+SRCREV = "abcc7c2135986a6d7d377f2f53c967cda0281351" 
+
+S = "${WORKDIR}/git"
+
```
otavio: Here is where I'd put the `inherit`, so it makes the header is more easier to follow.

recipes-security/optee/optee-os-imx_git.bb
```diff
@@ -0,0 +1,84 @@
...
+        cp $f ${DEPLOYDIR}
+    done
+}
+
+addtask deploy after do_install before do_build
+
+
+FILES_${PN} = "/lib/firmware/"
+FILES_${PN}-dev = "/usr/include/optee"
+INSANE_SKIP_${PN}-dev = "staticdev"
```
otavio: Instead of moving the static to the `-dev`, please do:

```
RDEPENDS_${PN}-dev = "${PN}-staticdev"
```

recipes-security/optee/optee-test-imx_git.bb
```diff
@@ -0,0 +1,48 @@
+# Copyright (C) 2017-2018 NXP
+
+SUMMARY = "OPTEE test"
+HOMEPAGE = "http://www.optee.org/"
+
+LICENSE = "BSD"
+LIC_FILES_CHKSUM = "file://LICENSE.md;md5=daa2bcccc666345ab8940aab1315a4fa"
+
+inherit pythonnative
+DEPENDS = "optee-os-imx optee-client-imx python-pycrypto-native"
```
otavio: Same comments regarding the `DEPENDS` and `inherit` location.

classes/image_types_fsl.bbclass
```diff
@@ -42,3 +42,6 @@ IMAGE_CMD_wic_append_mxs() {
 	bbnote "Setting partition type to 0x53 as required for mxs' SoC family."
 	echo -n S | dd of=$out${IMAGE_NAME_SUFFIX}.wic bs=1 count=1 seek=450 conv=notrunc
 }
+
+# Include required software for optee
+IMAGE_INSTALL += "${@bb.utils.contains('MACHINE_FEATURES', 'optee', 'packagegroup-fsl-optee-imx', '', d)}"
```
otavio: Please move this to `imx-base.inc` and add it to `MACHINE_ESSENTIAL_RRECOMMENDS` or similar.

conf/machine/imx6qdlsabreauto.conf
thochstein: I did not.

otavio: I am afraid it won't work, as community U-Boot does not support optee as far as I know.

thochstein: Okay, shall I add it with use-fsl-bsp qualification?

otavio: Yep!

otavio:
Closing this as it is supperseeded.

PR #19, opened 2018-10-24
linux-imx-headers

thochstein:


recipes-kernel/linux/linux-imx-headers_4.9.123.bb
```diff
@@ -0,0 +1,47 @@
...
+
+SRCBRANCH = "imx_4.9.123_imx8mm_ga"
+LOCALVERSION = "-${SRCBRANCH}"
+SRC_URI = "git://source.codeaurora.org/external/imx/linux-imx.git;protocol=https;branch=${SRCBRANCH} \
+    file://0001-uapi-Add-dma-buf.h-and-hantrodec.h.patch"
+SRCREV = "6a71cbc089755afd6a86c005c22a1af6eab24a70"
+S = "${WORKDIR}/git"
+
+do_compile() {
+}
```
otavio: Please:

```
do_compile[noexec] = "1"
```

recipes-kernel/linux/linux-imx-headers_4.9.123.bb
```diff
@@ -0,0 +1,47 @@
...
+    videodev2.h \
+"
+
+do_install() {
+    oe_runmake headers_install INSTALL_HDR_PATH=${B}${exec_prefix}
+    src=${B}${includedir}/linux
+    dest=${D}${includedir}/imx/linux
+    install -d $dest
+    cd $src
+    install -m 0644 ${IMX_UAPI_HEADERS} $dest
```
otavio: We need to remove the generic headers so *only* i.MX specific ones are left.

recipes-kernel/linux/linux-imx-headers_4.9.123.bb
```diff
@@ -20,6 +20,7 @@ do_compile() {
 IMX_UAPI_HEADERS = " \
     dma-buf.h \
     hantrodec.h \
+    ion.h \
```
otavio: Squash this commit in the previous one.

recipes-kernel/linux/linux-imx-headers_4.9.123.bb
thochstein: Not sure what you mean. Only ${IMX_UAPI_HEADERS} is installed to ${D}.

otavio: It makes sense, sorry for the noise ...

recipes-multimedia/gstreamer/gstreamer1.0-plugins-base_1.14.imx.bb
```diff
@@ -63,12 +60,7 @@ PACKAGECONFIG[x11]          = "${X11ENABLEOPTS},${X11DISABLEOPTS},${X11DEPENDS}"
 
 EXTRA_OECONF += " \
     --enable-zlib \
-    CPPFLAGS=" \
-    -I${STAGING_KERNEL_DIR}/include/uapi \
-    -I${STAGING_KERNEL_DIR}/include \
-    -I${STAGING_KERNEL_DIR}/drivers/staging/android/uapi \
-    -I${STAGING_KERNEL_BUILDDIR}/include/generated/uapi \
-    " \
+    CPPFLAGS="-I${STAGING_DIR_TARGET}${includedir}/imx" \
```
otavio: Please use `${STAGING_INCDIR}/imx` as it is smaller.

recipes-kernel/linux/linux-imx-headers_4.9.123.bb
```diff
@@ -0,0 +1,49 @@
...
+    file://0001-uapi-Install-custom-headers.patch"
+SRCREV = "6a71cbc089755afd6a86c005c22a1af6eab24a70"
+S = "${WORKDIR}/git"
+
+do_compile[noexec] = "1"
+
+IMX_UAPI_HEADERS = " \
+    dma-buf.h \
+    hantrodec.h \
+    hx280enc.h \
```
thochstein: I added hx280enc.h in the latest pass.

otavio:
It seems good; we just need to double check what else we can move to depend on the new linux-imx-headers. Suspects are:

- [ ] imx-lib
- [ ] imx-test
- [ ] imx-uuc
- [ ] imx-vpu
- [ ] imx-alsa-plugins
- [ ] gstreamer1.0-plugins-imx
- [ ] imx-gst1.0-plugin

As those should be part of this PR and we can merge them all in one shot.

recipes-bsp/imx-vpu/imx-vpu_5.4.38.bb
```diff
@@ -22,14 +22,13 @@ inherit fsl-eula-unpack
...
+    INCLUDE_DIR="-I${STAGING_INCDIR}/imx"
     oe_runmake CROSS_COMPILE="${HOST_PREFIX}" PLATFORM="${PLATFORM}" INCLUDE="${INCLUDE_DIR}" all
 }
 
 do_install () {
     oe_runmake PLATFORM="${PLATFORM}" DEST_DIR="${D}" install
 }
 
-PACKAGE_ARCH = "${MACHINE_ARCH}"
+PACKAGE_ARCH = "${MACHINE_SOCARCH}"
```
thochstein: This line is redundant to MACHINE_SOCARCH_FILTER.

otavio: You're right ... however, it does not hurt. 

In fact, after we finish going over the recipes, I will rework them one more time, globally, to avoid duplication but first I wish to go over the minimal set of changes which allow use to prove it keeps all working.

recipes-multimedia/gstreamer/gstreamer1.0-plugins-imx_git.bb
```diff
@@ -35,7 +33,8 @@ EGLVIVSINK_PLATFORM = "${@bb.utils.contains('DISTRO_FEATURES', 'wayland', 'wayla
                           'fb', d),d)}"
 
 EGL_PLATFORM_CONF = "--egl-platform=${EGLVIVSINK_PLATFORM}"
-EXTRA_OECONF = "--kernel-headers=${STAGING_KERNEL_DIR}/include ${PACKAGECONFIG_CONFARGS}"
+
+EXTRA_OECONF = "CFLAGS="-I${STAGING_INCDIR}/imx""
```
otavio: You must keep the previous:

```
EXTRA_OECONF = "--kernel-headers=${STAGING_INCDIR}/imx ${PACKAGECONFIG_CONFARGS}"
```

otavio:
Hello Tom, 

I made a small rework on the recipes to simplify its use and also to allow for later improvements with easy. 

Take a look at the class and please try it at your side. Assuming it works, we can merge it.

gibsson:
> It seems good; we just need to double check what else we can move to depend on the new linux-imx-headers. Suspects are:
> 
> * [x]  imx-lib
> * [x]  imx-test
> * [x]  imx-uuc
> * [x]  imx-vpu
> * [x]  imx-alsa-plugins
> * [x]  gstreamer1.0-plugins-imx
> * [x]  imx-gst1.0-plugin
> 
> As those should be part of this PR and we can merge them all in one shot.

But in that case what's the plan for machines that use a kernel older than 4.9? Do they have to stick to sumo?

Having a look at meta-freescale-3rdparty, it seems that many vendors are still using 3.14/4.1:
https://github.com/Freescale/meta-freescale-3rdparty/tree/master/recipes-kernel/linux

Do we consider that every board vendor must be aligned with NXP's latest at all time or do we create as many linux-imx-headers recipes as there are kernel versions supported? Note that for i.MX6, nothing prevents people to stay on 4.1 or 3.14 but still use latest graphics/vpu libs.

That's for the above reason that I am not fond of this approach.

otavio:
Hello @gibsson, thanks for joining the review :-)

> But in that case what's the plan for machines that use a kernel older than 4.9? Do they have to stick to sumo?
> 
> Having a look at meta-freescale-3rdparty, it seems that many vendors are still using 3.14/4.1:
> https://github.com/Freescale/meta-freescale-3rdparty/tree/master/recipes-kernel/linux
> 
> Do we consider that every board vendor must be aligned with NXP's latest at all time or do we create as many linux-imx-headers recipes as there are kernel versions supported? Note that for i.MX6, nothing prevents people to stay on 4.1 or 3.14 but still use latest graphics/vpu libs.
> 
> That's for the above reason that I am not fond of this approach.

I understand your concern and yes, it is going to cause some problems. Many of those board are outdated because vendors are not paying attention to community BSP.

There many side benefits of having this in a general way, mainly:

* reduces build time for multiple boards
* provides a global ABI for all boards
* reduce dependencies

So we will need to ping vendors to update their machines and in case they don't do that, remove broken boards. This ultimately is not a bad thing as if they don't care why should we care?

gibsson:
> Hello @gibsson, thanks for joining the review :-)

Hello, also happy to be part of the discussion.

> > But in that case what's the plan for machines that use a kernel older than 4.9? Do they have to stick to sumo?
> > Having a look at meta-freescale-3rdparty, it seems that many vendors are still using 3.14/4.1:
> > https://github.com/Freescale/meta-freescale-3rdparty/tree/master/recipes-kernel/linux
> > Do we consider that every board vendor must be aligned with NXP's latest at all time or do we create as many linux-imx-headers recipes as there are kernel versions supported? Note that for i.MX6, nothing prevents people to stay on 4.1 or 3.14 but still use latest graphics/vpu libs.
> > That's for the above reason that I am not fond of this approach.
> 
> I understand your concern and yes, it is going to cause some problems. Many of those board are outdated because vendors are not paying attention to community BSP.

I understand but I wouldn't discard them though. Especially those using 4.1 kernels, it's still fairly recent in the NXP world, many vendors haven't switched to 4.9 yet (not just on the Yocto BSP).

> There many side benefits of having this in a general way, mainly:
> 
> * reduces build time for multiple boards

How so? You still need to build the kernel so I'm not sure to follow.
Actually it increases the build time for 3rd-party vendors as it downloads both the vendor kernel + the full NXP tree just for the headers.

> * provides a global ABI for all boards

But the ABI stays the same for all kernels, as long as you use 4.9 based on NXP fork you'll have the same.

> * reduce dependencies

How so? You still need to build the kernel for those packages to be useful.
Here you replacing the kernel dependency on kernel-header dependency (which in turns downloads the full kernel anyway).

> So we will need to ping vendors to update their machines and in case they don't do that, remove broken boards. This ultimately is not a bad thing as if they don't care why should we care?

Once again, I don't think it's only a matter of updating the recipe, some vendors haven't switched to 4.9 yet because they haven't run all their tests on it.

All in all, I think this approach locks things down for unnecessary reasons (as it works fine without that approach today in a more flexible manner).

otavio:
> > I understand your concern and yes, it is going to cause some problems. Many of those board are outdated because vendors are not paying attention to community BSP.
> 
> I understand but I wouldn't discard them though. Especially those using 4.1 kernels, it's still fairly recent in the NXP world, many vendors haven't switched to 4.9 yet (not just on the Yocto BSP).

They can still keep using old Yocto Project releases. I am sorry but their slowness cannot block things to improve.

> > There many side benefits of having this in a general way, mainly:
> > 
> > * reduces build time for multiple boards
> 
> How so? You still need to build the kernel so I'm not sure to follow.
> Actually it increases the build time for 3rd-party vendors as it downloads both the vendor kernel + the full NXP tree just for the headers.

We make more binaries SoC arch and as consequence, more recipes are skip when doing multiple machines.

For you, it may not be a big deal but for customers which use multiple machines and for vendors who have multiple machines, it has a huge impact.

> > * provides a global ABI for all boards
> 
> But the ABI stays the same for all kernels, as long as you use 4.9 based on NXP fork you'll have the same.

Yes but this is not guaranteed by the layer until we force it to stick to a static set of headers.

> > * reduce dependencies
> 
> How so? You still need to build the kernel for those packages to be useful.
> Here you replacing the kernel dependency on kernel-header dependency (which in turns downloads the full kernel anyway).
> 
> > So we will need to ping vendors to update their machines and in case they don't do that, remove broken boards. This ultimately is not a bad thing as if they don't care why should we care?
> 
> Once again, I don't think it's only a matter of updating the recipe, some vendors haven't switched to 4.9 yet because they haven't run all their tests on it.
> 
> All in all, I think this approach locks things down for unnecessary reasons (as it works fine without that approach today in a more flexible manner).

I respect your view @gibsson but vendors will need to upgrade or move to mainline Linux kernel.

In fact, many vendors do not upgrade at all. Those, I am not concerned as this is their decision.

All we can do to assist faster upgrades, we do. We have shared tree for Linux and U-Boot which allows for reducing work duplication and sooner or later they need to upgrade.

recipes-bsp/imx-test/imx-test_git.bb
```diff
@@ -56,8 +56,8 @@ do_compile() {
     oe_runmake V=1 VERBOSE='' \
                CROSS_COMPILE=${TARGET_PREFIX} \
                INC="-I${STAGING_INCDIR} \
+                    -I${STAGING_IMX_INCDIR}" \
```
thochstein: Isn't the order here wrong? The source headers should have a higher priority than the sysroot headers, in which case line 58 is also wrong.

classes/use-imx-headers.bbclass
```diff
@@ -0,0 +1,19 @@
...
+#
+# Please use the STAGING_IMX_INCDIR variable to refer to the installed
+# headers.
+#
+# Copyright 2018 (C) O.S. Systems Software LTDA.
+
+DEPENDS_imx_append = " linux-imx-headers"
+PACKAGE_ARCH_imx ?= "${MACHINE_SOCARCH}"
+
+STAGING_IMX_INCDIR = "${STAGING_INCDIR}/imx"
```
thochstein: STAGING_INCDIR_IMX seems a better name since the word order follows the directory order.

classes/use-imx-headers.bbclass
```diff
@@ -0,0 +1,19 @@
...
+#
+# All recipes using this class ought to depend on linux-imx-headers
+# and by default to use MACHINE_SOCARCH.
+#
+# Please use the STAGING_IMX_INCDIR variable to refer to the installed
+# headers.
+#
+# Copyright 2018 (C) O.S. Systems Software LTDA.
+
+DEPENDS_imx_append = " linux-imx-headers"
```
thochstein: This needs to be DEPENDS_append_imx, right?

recipes-bsp/imx-test/imx-test_git.bb
otavio: You are right, can you fix it?

classes/use-imx-headers.bbclass
otavio: Great, mind to fix it?

classes/use-imx-headers.bbclass
otavio: Yes, it is indeed better

classes/use-imx-headers.bbclass
thochstein: Will do.

PR #30, opened 2018-12-13
L4.9.123 2.3.0 181213

thochstein:


recipes-bsp/imx-vpu-hantro/imx-vpu-hantro_1.8.0.bb
```diff
@@ -32,4 +32,4 @@ do_install () {
 
 FILES_${PN} += "/unit_tests"
 
-COMPATIBLE_MACHINE = "(mx8mq|mx8mm)"
+COMPATIBLE_MACHINE_imxvpu = "(mx8m)"
```
otavio:
```
COMPATIBLE_MACHINE = "(mx8m)"
```

Avoid using an override for this as it is hard to understand.

PR #31, opened 2019-01-08
imx-base.inc: use SDMA RAM script for the whole i.MX 6 family

agners:
The Linux device trees of all members of the i.MX 6 family use the
same SDMA RAM script.

Signed-off-by: Stefan Agner <stefan.agner@toradex.com>


otavio:
Applied: https://github.com/Freescale/meta-freescale/commit/c9427b8da9a301c9f462c54986f592208ba161f8

PR #36, opened 2019-01-14
Series 15605

angolini:
It's only a PR for the patch series:
https://patchwork.openembedded.org/series/15605/


otavio:
Applied on master.

PR #37, opened 2019-01-14
Move libusb receipes to use libusb1

### LINKED ISSUE #35: Receipes that depend on libusb are broken
They should depend on "libusb1"

deinok:
Closes #35


deinok:
@otavio Can we backport this to thud also?

PR #45, opened 2019-02-02
L4.9.123 2.3.0 0201

thochstein:


recipes-graphics/imx-gpu-apitrace/imx-gpu-apitrace_7.1.0.bb
```diff
@@ -34,4 +34,4 @@ FILES_${PN} = "${bindir} ${libdir}"
 FILES_${PN}-dbg += "${libdir}/*/*/.debug"
 
 PACKAGE_ARCH = "${MACHINE_SOCARCH}"
-COMPATIBLE_MACHINE = "(mx6q|mx6dl|mx6sx|mx6sl|mx7ulp|mx8mq)"
+COMPATIBLE_MACHINE = "(mx6q|mx6dl|mx6sx|mx6sl|mx7ulp|mx8)"
```
otavio: This should be another commit ... also, I'd prefer to not use generic overrides if possible ...

thochstein:
This is ready.

PR #47, opened 2019-02-18
u-boot.imx: Add u-boot configuration for fsl

thochstein:
The fsl distro was broken for the following machine files
because they set u-boot configuration to fslc configuration:

- imx6qdlsabreauto
- imx6qdlsabresd
- imx6ulevk


thochstein:
Thanks, don't forget the meta-freescale-distro change as well.

PR #48, opened 2019-02-21
Update linux-fslc and fix U-Boot build

fbertux:


recipes-kernel/linux/linux-fslc/defconfig
```diff
@@ -166,6 +167,7 @@ CONFIG_WL18XX=m
 CONFIG_WLCORE_SDIO=m
 # CONFIG_WILINK_PLATFORM_DATA is not set
 CONFIG_INPUT_EVDEV=y
+CONFIG_INPUT_EVBUG=m
```
otavio: This should not be included. Please drop it.

PR #79, opened 2019-04-13
wic: Fix image generation for i.MX

texierp:
Some boards expects a specific binary name.

Signed-off-by: Pierre-Jean Texier <pjtexier@koncepto.io>
Signed-off-by: Joris Offouga <offougajoris@gmail.com>

### DIFF
--- a/conf/machine/include/imx-base.inc
+++ b/conf/machine/include/imx-base.inc
@@ -14,6 +14,7 @@ PREFERRED_PROVIDER_virtual/bootloader ??= "${IMX_DEFAULT_BOOTLOADER}"
 
 PREFERRED_PROVIDER_u-boot-mxsboot-native ??= "u-boot-fslc-mxsboot-native"
 
+UBOOT_BINARY ?= "u-boot.imx"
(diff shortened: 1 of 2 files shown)


otavio:
I understand the change but isn't it always: `u-boot.imx` and `SPL` + `u-boot.img`?

texierp:
In fact, with the WaRP7, we have to flash the following file : u-boot-dtb.imx [1]
So, without this change, wic image creation fail.

[1] http://git.denx.de/?p=u-boot.git;a=blob;f=board/warp7/README;h=579bb2de38a9d40fd5972193d4722b765b0b512c;hb=HEAD#l35

otavio:
So your commit log is misleading.

However, last time I checked, the u-boot-dtb.imx and u-boot.imx were the same (same checksum) and we could simply keep using the same name.

texierp:
Indeed, I just update the commit log, and I removed the part related to SPL & u-boot.img.

Now, I kept only the wic script with bootpart, good for you ?

PR #85, opened 2019-05-03
Build fixes

JoshuaWatt:


JoshuaWatt:
> Hi
> 
> @ imx-boot: Pass DTB names to make:
> 
> This code is also used for mx8 and mx8x which do not define UBOOT_DTB_NAME.
> Does it work for them too? I.e. does the following work?
> make SOC=iMX8QM dtbs= flash
> 
> Wouldn't it be better to define a new variable and then use that in the make call?
> UBOOT_DTBS_mx8m = "dtbs=${UBOOT_DTB_NAME}"
> UBOOT_DTBS = "*

I'm not sure if passing an empty dtbs= works or not; If it doesn't, it would probably be nice if it was more "automatic" though, something like:

UBOOT_DTB_NAME ??= ""
```
do_compile()
    ...
    DTBS=""
    if [ -n "${UBOOT_DTB_NAME}" ]; then
        DTBS="dtbs=${UBOOT_DTB_NAME}"
    fi
    make $DTBS ....
    ...
```
   



> 
> Max

MaxKrummenacher:
I tried it out.
Passing 'dtbs=<whitespace>' to the make call has no negative side-effects for mx8 and mx8x.
Sorry for the noise.
Max

JoshuaWatt:
Thanks for verifying.

PR #92, opened 2019-05-14
u-boot-imx upgrade to 2018.03

mihaix:


recipes-bsp/u-boot/u-boot-imx_2018.03.bb
```diff
@@ -1,6 +1,6 @@
 # Copyright (C) 2013-2016 Freescale Semiconductor
-# Copyright 2017-2018 NXP
 # Copyright 2018 (C) O.S. Systems Software LTDA.
+# Copyright (C) 2017-2019 NXP
```
angolini: Only because you are changing the Copyright. I would suggest you to use:

#Copyright 2017-2019 NXP

only. No need to add (C). If this is the only change needed in this PR I would accept it as is though.

MaxKrummenacher:
Hi

I propose to drop / postpone removing the dtc-145 class and recipe until all users in meta-freescale* have upgraded to a U-Boot which can be built with the dtc provided from oe-core.

Max

otavio:
I second @MaxKrummenacher request here; please rework the commit and send it for review.

recipes-bsp/u-boot/u-boot-imx_2018.03.bb
mihaix: Since I am updating the years in the copyright line, I wanted to also update the style and add (c), which is more MIT and GPL like and was previously used with Freescale. Now we have some files with 3 different styles.
It was just a thought, but ok I can drop this.

mihaix:
Ok, right, forgot about checking with 3rdparty, I was looked in meta-freescale only.

recipes-bsp/u-boot/u-boot-imx_2018.03.bb
angolini: It's not a big deal anyway

otavio:
If you are in the mood, please send a PR to remove the old dtc and we can open issues for the respective boards and merge it when complete.

mihaix:
> If you are in the mood, please send a PR to remove the old dtc and we can open issues for the respective boards and merge it when complete.

I've submitted a draft PR that we can use for this.

PR #94, opened 2019-05-16
utilities.inc: handle illicit KERNEL_DEVICETREE entries

jdallaglio:
Function `make_dtb_boot_files` didn't return if KERNEL_DEVICETREE contained illicit non .dtb or .dtbo files, resulting in Python `str` method `join` trying to concatenate `None` and string objects.
Instead we now raise a warning, remove the illicit device tree entries and return an empty string.

### DIFF
--- a/conf/machine/include/utilities.inc
+++ b/conf/machine/include/utilities.inc
@@ -13,4 +13,4 @@ def make_dtb_boot_files(d):
             #     destination: bcm2708-rpi-b.dtb
             return os.path.basename(dtb)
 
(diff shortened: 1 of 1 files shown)


angolini:
Is it supposed to be on warrior only? (and skip master?)

jdallaglio:
I tested it on warrior only. I'm testing it on master right now

jdallaglio:
It works on master as well, should I remove this pull request and do a new one on master?

angolini:
I would say, leave this PR and add a new one for master.

Thanks a lot

jdallaglio:
Cool, thanks

PR #95, opened 2019-05-16
utilities.inc: handle illicit KERNEL_DEVICETREE entries

jdallaglio:
In utilities.inc function make_dtb_boot_files now raises a warning and
removes the KERNEL_DEVICETREE entries if they don't have .dtb or .dtbo
extensions.


otavio:
This is the wrong approach and won't be accepted ... we cannot skip the files, we can raise and error.

jdallaglio:
Ok, I'll close this one and submit a corrected pull request. Thanks!

PR #101, opened 2019-05-20
linux-fslc-stable-4.19: initial add

schnitzeltony:
defconfigs were taken from

| commit 60df6bc2b69a7f405f4b14fee1d53e8b3b84c054
| Author: Otavio Salvador <otavio@ossystems.com.br>
| Date:   Thu Jan 31 10:24:42 2019 -0200
|
|     linux-fslc: Remove 4.9 leftover defconfig files
|
|    Signed-off-by: Otavio Salvador <otavio@ossystems.com.br>

With the next commit

| commit c41b6d8e6ec9e8394727ddd0e6f1a8a5591b5e58
| Author: Fabio Berton <fabio.berton@ossystems.com.br>
| Date:   Thu Feb 21 09:51:58 2019 -0300
|
|     linux-fslc: Upgrade to 4.20-based fork
|
|     This upgrades the kernel to the 4.20-based fork and sync defconfig
|     files with kernel source tree.
|
|    Signed-off-by: Fabio Berton <fabio.berton@ossystems.com.br>

linux-fslc moved to 4.20 and defconfigs were adjusted.

Signed-off-by: Andreas Müller <schnitzeltony@gmail.com>


otavio:
Any reason to keep 4.19? We could move to 5.1 no?

schnitzeltony:
I want a stable and upstream maintained kernel without surprises. Latest stable is 4.19. Cannot risk major version update for our products. Hence I suggest a new recipe. If somebody wants to upgrade linux-fslc that is a different story.
BTW: This is similar to what RPi foundation has done for many years and there kernel updates were almost boring - never had major issues (yeah there was some trouble recently - but now they are back to same model).

otavio:
@angolini @fbertux please let me know what you think about it.

@schnitzeltony if we intend to keep 4.19, I'd prefer we call the recipe `linux-fslc-lts` so we bump every new lts release.

schnitzeltony:
We can rename it to linux-fslc-lts - (need to adjust patches for -3rd party either then). Version number in name was intentional to have the chance of multiple stable/lts versions. E.g we could  continue 4.19 even when next stable/lts (5.?) is there.

otavio:
I don't intend to keep old versions for a long time; once a new lts is out we should start to push for migration as it avoids a lot of maintenance burden. We had, in the past, multiple versions and it ends is not maintained and often not buildable.

schnitzeltony:
It's a bit different here:

* It is not like all these vendor kernels (I consider them crap/maintenance burden and would love to see them go - but that is a different story). Linux-fslc is close to upstream - so maintenance is cheap (usually a simple merge) [1]  and for 4.19 I can take care.
* Even if we drop linux-fslc-lts-4.19 once next stable is out and replace it by linux-fslc-lts-5.x: we clearly state that this major version update and adjustments / issues are to expect.

[1] https://github.com/raspberrypi/linux/pull/2947#issuecomment-487566099

angolini:
Currently there are 5 LTS kernels. 2 of them EOL next year (2020). [link](https://www.kernel.org/category/releases.html)

I agree with having support for the LTS kernel in addition to the latest stable version (today, it would be 4.19 and 5.1).

My concern is regarding supporting the newer LTS or the longer supported LTS (does that make sense?) Today it would be choosing between 4.19 (2020) or 4.9 (2023).

And if we decide for the newer LTS, how long until we drop it support in favor of newer-newer LTS?

schnitzeltony:
* 5.1 LTS - did I miss something?
* 4.9: Would it work with recent mesa/etnaviv

For the record:
End of last year when rumors started that 4.19 will be LTS I sent some commits to linux-fsl, here and to 3rd-party to get 4.19 working fine with the intention to make it LTS.

OK - I have only VarSOM imx6 running with it - but on that I am REALLY happy - there are no issues biting me and graphic performance is pretty good with X11 AND wayland. This is a state I would like to preserve for as long as possible :)

otavio:
As I said, I am fine in getting this in ... in future, we will drop the 4.19 and move to next LTS.

It might be 5.1, 5.2 or next but eventually we should push for removing it. If you as maintainer wishes to keep it for the board, it is fine but it will be on the meta-freescale-3rdparty.

PR #107, opened 2019-05-24
FSLC this and that

schnitzeltony:
1.st: menuconfig: tested - yes both linux-fslc and linux-fslc-lts-4.19 have scrambled GUI without it
2.nd: USB_PRINTER and USB_SERIAL_PL2303 : should not explode build times
3.rd lts: update: run tested no issues


schnitzeltony:
Maybe the title was not chosen wisely...

Ping?

PR #113, opened 2019-06-19
qoriq-base: Use the same provider for u-boot and u-boot-fw-utils

AlbanBedel:
It make sense for the default config to use the same provider for
for both u-boot and u-boot-fw-utils. We now assume that there is a
matching -fw-utils package, if not the build will fail instead of
silently using a mismatching u-boot version.

Signed-off-by: Alban Bedel <alban.bedel@avionic-design.de>


otavio:
@AlbanBedel please make a PR for master as well.

PR #139, opened 2019-07-24
Fix problems with dirty DEPLOYDIR during do_deploy

dkl:
The management-complex recipe (and possibly others) assumed that DEPLOYDIR was empty before do_deploy. But this is not always true - with incremental builds, DEPLOYDIR can contain files from previous builds (possibly with different configuration even).

Here I saw the problem specifically with management-complex.bb, which created a recursive mc.itb symlink, because do_deploy didn't handle the case when mc.itb already exists. This can be fixed by writing do_deploy such that it can handle existing files (which I tried to do for management-complex). But the more generic approach seems to be doing `do_deploy[cleandirs] = "${DEPLOYDIR}"`, telling bitbake to empty the directory every time.

Both fixes are included in this pull request, to fix the specific issue with management-complex but also to prevent similar issues that may be possible now or in the future for the other recipes using do_deploy.

### DIFF
--- a/recipes-bsp/atf/atf_git.bb
+++ b/recipes-bsp/atf/atf_git.bb
@@ -207,6 +207,7 @@ do_deploy() {
         cp -r ${D}/boot/atf/fip_ddr_sec.bin ${DEPLOYDIR}/atf/fip_ddr_sec.bin
     fi
 }
+do_deploy[cleandirs] = "${DEPLOYDIR}"
 addtask deploy after do_install
 FILES_${PN} += "/boot"
 BBCLASSEXTEND = "native nativesdk"
--- a/recipes-bsp/ddr-phy/ddr-phy_git.bb
+++ b/recipes-bsp/ddr-phy/ddr-phy_git.bb
@@ -33,6 +33,7 @@ do_deploy () {
     install -d ${DEPLOYDIR}/ddr-phy
     install -m 755 ${S}/${REGLEX}/*.bin ${DEPLOYDIR}/ddr-phy
 }
+do_deploy[cleandirs] = "${DEPLOYDIR}"
 addtask deploy before do_populate_sysroot after do_install
 
 PACKAGES += "${PN}-image"
--- a/recipes-bsp/firmware-imx/firmware-imx-8_8.1.1.bb
+++ b/recipes-bsp/firmware-imx/firmware-imx-8_8.1.1.bb
@@ -16,6 +16,7 @@ do_deploy() {
     # SECO
     install -m 0644 ${S}/firmware/seco/mx8qm-ahab-container.img ${DEPLOYDIR}
 }
+do_deploy[cleandirs] = "${DEPLOYDIR}"
 addtask deploy after do_install before do_build
 
 PACKAGE_ARCH = "${MACHINE_SOCARCH}"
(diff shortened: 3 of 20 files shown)


otavio:
Is this still an issue?

dkl:
Yes, this problem can still happen, since as far as I know there were no other changes that fixed it. It's an issue only for incremental Yocto builds, not with fresh Yocto builds.

otavio:
@dkl I'd like to discuss this with @rpurdie as it seems it belongs to `classes/deploy.bbclass` and I wonder why it is not done there. @rpurdie any reason to not add this there?

rpurdie:
I have a feeling that some code will rely on being able to place things into this directory before the do_deploy task. Technically that would be a bug but I think if we were to do this we will see problems. It would of course make those problems clearer so perhaps we should...

otavio:
@dkl mind to cook a patch to OE-Core and add me on Cc?

dkl:
Hmm, sure, it would be nice to fix this properly. Making the initial patch for that is easy enough, but looking through poky, I noticed uboot-sign.bbclass which looks like it could be broken by that (due to usage of do_deploy[prefuncs] and cleandirs being applied once for every func/task) and I don't know how to fix it (much less how to test it). Maybe I can figure it out. At least the other users of deploy.bbclass (only a few in poky and meta-openembedded) look like they would be unaffected.

In that case the 2nd commit of this pull request would be obsolete, only the 1st one may still be interesting for meta-freescale.

otavio:
> Hmm, sure, it would be nice to fix this properly. Making the initial patch for that is easy enough, but looking through poky, I noticed uboot-sign.bbclass which looks like it could be broken by that (due to usage of do_deploy[prefuncs] and cleandirs being applied once for every func/task) and I don't know how to fix it (much less how to test it). Maybe I can figure it out. At least the other users of deploy.bbclass (only a few in poky and meta-openembedded) look like they would be unaffected.

Indeed; it seems it is a single function which is run so maybe it could be `prepend` to the `do_deploy` one. @rpurdie ?

dkl:
Doing `do_deploy_prepend` breaks `kernel_do_deploy` somehow, as mentioned in the comment in uboot-sign.bbclass, but `do_deploy_prepend_pn-${UBOOT_PN}` seems to work, so I'll make a patch with that.

dkl:
Update on this: I posted the patch for oe-core (https://patchwork.openembedded.org/series/19533/) some time ago, however it didn't get picked up. I still think it would be useful for meta-freescale to replace this pull request, but it doesn't look like it will happen.

I'll wait a few more days, then rebase this pull request. If the patch doesn't make it into oe-core, the issues should be fixed here at least.

otavio:
Did you resent the patch?

rpurdie:
I did merge this but it is causing problems for people and I'm getting complaints unfortunately. I do think the patch is correct though.

otavio:
Hello @rpurdie, what issues are people complaining about? We might try to support this cases.

rpurdie:
People are writing deploy data outside the deploy tasks in some recipes/layers. I think issues will come up on the mailing lists in due course, so far its been complaints on irc. I'm not sure they should be writing like this...

otavio:
Well, as you said, the fix is indeed correct and we might try to see if we can come up with a way to support old behavior. Or we just say it fixes an old error and they need to fix it in their side.

What do you say @rpurdie?

rpurdie:
We'll see what is reported. I just wanted to warn that it may be going to cause some complications.

dkl:
Since the patch made in into oe-core:master, we can drop this one here.

By the way, how is it working out upstream, any big issues, or are people adjusting their recipes to it?

PR #166, opened 2019-09-05
cst: Fix build for target recipe and limit dpdk to imx alone

kraj:
gen_keys when cross-built can not execute on build host
so better to depend on native version and call out native gen_keys
binary

Fixes
TOPDIR/build/tmp/work/mips32r2-yoe-linux/cst/git-r0/temp/run.do_install.29171: line 111: ./gen_keys: cannot execute binary file: Exec format error
WARNING: TOPDIR/build/tmp/work/mips32r2-yoe-linux/cst/git-r0/temp/run.do_install.29171:1 exit 126 from './gen_keys 1024'


kraj:
ping^1

PR #173, opened 2019-09-23
imx7ulpevk.conf: Fix MACHINEOVERRIDES order

thochstein:
Setting the SOC family and the SOC in MACHINEOVERRIDES after
tune-cortexa7.inc has updated MACHINEOVERRIDES causes the tuning
overrides to have higher priority than the SOC overrides, which
is wrong.


otavio:
What problem does this solve?

thochstein:
Not sure of the exact manifestation, fix was some time ago.

I guess I can at least give a general description, that any variable with a more-specific SOC override would incorrectly be overridden by a less-specific _imx override. You want me to add that to the commit message?

otavio:
Yes, please.

PR #175, opened 2019-09-23
imx7ulpevk.conf: Fix MACHINEOVERRIDES order

thochstein:
The exact manifestation that uncovered this problem was not recorded.
Generally speaking, a recipe misconfiguration was occurring due to an
incorrect variable value, caused by incorrectly choosing the _imx
override of the variable instead of the SOC override.

This is fixed by setting the SOC family and the SOC in MACHINEOVERRIDES
before including tune-cortexa7.inc.

### DIFF
--- a/conf/machine/imx7ulpevk.conf
+++ b/conf/machine/imx7ulpevk.conf
@@ -4,12 +4,12 @@
 #@DESCRIPTION: Machine configuration for NXP i.MX7ULP EVK
 #@MAINTAINER: Lauren Post <Lauren.Post@nxp.com>
 
+MACHINEOVERRIDES =. "mx7:mx7ulp:"
+
 require conf/machine/include/imx-base.inc
 require conf/machine/include/tune-cortexa7.inc
 
 MACHINE_FEATURES += " pci wifi bluetooth qca9377"
-
-MACHINEOVERRIDES =. "mx7:mx7ulp:"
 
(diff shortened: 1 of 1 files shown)


otavio:
@thochstein we need consistency. 

I am not aware of issues related to this so until we have a know problem, I am not accepting this.

thochstein:
Here's the incorrect value of `MACHINEOVERRIDES` with the current implementation:

`use-fsl-bsp:mx7:imxfbdev:imxpxp:imxgpu:imxgpu2d:imxgpu3d:mx7ulp:armv7ve:imx:imx7ulpevk`

`armv7ve` and `imx` should never have priority over any of `mx7:imxfbdev:imxpxp:imxgpu:imxgpu2d:imxgpu3d:mx7ulp`, correct? Here's the corrected value with my patch:

`use-fsl-bsp:armv7ve:imx:mx7:imxfbdev:imxpxp:imxgpu:imxgpu2d:imxgpu3d:mx7ulp:imx7ulpevk`

Are you saying that this obvious error in the variable order is not enough to fix it?

otavio:
No, I am fine in fixing this but not without a reason.

In any case, this must **first** be fixed on **master** and then do a backport. I am closing this as a new PR based on the master merged changes will be needed.

otavio:
Please do a cherry-pick and send a new PR.

otavio:
Please use `git cherry-pick -x`.

PR #178, opened 2019-10-02
imx-gpu-viv: fix build with libvulkan

MaxKrummenacher:
Please also compare with [this thread](https://lists.yoctoproject.org/pipermail/meta-freescale/2019-October/024078.html).
Note that this makes the build compile, I did not yet runtime test it.

### DIFF
--- a/recipes-graphics/imx-gpu-viv/imx-gpu-viv-6.inc
+++ b/recipes-graphics/imx-gpu-viv/imx-gpu-viv-6.inc
@@ -18,6 +18,7 @@ DEPENDS += " \
                                                        '', d)} \
 "
 DEPENDS_append_imxdrm = " libdrm wayland"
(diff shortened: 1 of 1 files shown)


MaxKrummenacher:
Will send an update which drops the additional vulkan directory as suggested by Ming Liu.

liuming50:
Maybe we could move ${D}${libdir}/vulkan/libvulkan_VSI.so to ${D}${libdir}/libvulkan_VSI.so, it seems does not have to be in vulkan sub-directory?

And we probably could drop "INSANE_SKIP_libvulkan-imx += "dev-deps dev-so"" as well.

MaxKrummenacher:
@liuming50 
Done. 
- lib/vulkan subdirectory is gone.
- The INSANE_SKIP is dropped. No Warnings appeared in my build.

PR #191, opened 2019-11-01
weston: fix the test condition for agl builds

zandrey:
Fix the variable test condition, wich is used during inclusion of
weston.ini file in the agl and non-agl builds.

Since for non-agl build there is no algprofilegraphics in the
collection, variable expands to an empty string and test condition
evaluates to true for both agl and non-agl builds.
This causes the weston.ini file to be installed regardless of whether
the agl is found or not.

Usage of a different condition check (lenght instead of empty) eliminates
this issue.

### DIFF
--- a/recipes-graphics/wayland/weston_5.0.0.imx.bb
+++ b/recipes-graphics/wayland/weston_5.0.0.imx.bb
@@ -120,7 +120,7 @@ do_install_append() {
 	fi
 
 	# install default weston.ini
-	if [ -z "${@bb.utils.filter('BBFILE_COLLECTIONS', 'aglprofilegraphical', d)}" ]; then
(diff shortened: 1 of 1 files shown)


zandrey:
As a side note to this PR: this fixes the master build, since now the weston.ini is provided by a separate recipe in core, and without this fix the rootfs build fails with the error that 2 packages provides this file.

@otavio @thochstein: Since I'm not actively working on the AGL and do not quite know the details of when and how ini files should be installed, can you please take a look at this PR to see whether in general it is a proper place to separate installation of weston.ini in this recipe, or i should be done in weston-init.bb? Thanks a lot!

-- andrey

otavio:
This needs rework but your fix is good.

otavio:
We should move the specific code to the dynamic-layers with a bbappend. @thochstein mind to look at it?

thochstein:
Will do.

PR #194, opened 2019-11-04
weston: Improve weston.ini selection logic

thochstein:
Add a variable WESTON_INI_INSTALL_FILE to make it easy to customize
the selection of the weston.ini file.

The main recipe now performs only a generic install of the selected file.

For AGL and IVI, we use the BBFILES_DYNAMIC mechanism to specify no
file and an IVI-specific file respectively.

### DIFF
--- a/conf/layer.conf
+++ b/conf/layer.conf
@@ -27,11 +27,17 @@ ${QORIQ_MIRROR} http://download.ossystems.com.br/bsp/freescale/source/ \n \
 "
 
 BBFILES_DYNAMIC += " \
+    aglprofilegraphical:${LAYERDIR}/dynamic-layers/aglprofilegraphical/*/*/*.bb \
+    aglprofilegraphical:${LAYERDIR}/dynamic-layers/aglprofilegraphical/*/*/*.bbappend \
+    \
     browser-layer:${LAYERDIR}/dynamic-layers/browser-layer/*/*/*.bb \
     browser-layer:${LAYERDIR}/dynamic-layers/browser-layer/*/*/*.bbappend \
     \
     filesystem-layer:${LAYERDIR}/dynamic-layers/filesystem-layer/*/*/*.bb \
     filesystem-layer:${LAYERDIR}/dynamic-layers/filesystem-layer/*/*/*.bbappend \
+    \
+    ivi:${LAYERDIR}/dynamic-layers/ivi/*/*/*.bb \
+    ivi:${LAYERDIR}/dynamic-layers/ivi/*/*/*.bbappend \
     \
     networking-layer:${LAYERDIR}/dynamic-layers/networking-layer/*/*/*.bb \
     networking-layer:${LAYERDIR}/dynamic-layers/networking-layer/*/*/*.bbappend \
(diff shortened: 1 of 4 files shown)


zandrey:
Hello all,

@thochstein, I'm having the same issues with this PR as I had previously before I've submitted PR #191 
Currently, when building the `fsl-image-multimedia-full` recipe, I receive following errors in `do_rootfs ()` task:
> Running transaction test
> Error: Transaction check error:
>   file /etc/xdg/weston/weston.ini conflicts between attempted installs of weston-init-1.0-r0.imx8mmevk and weston-5.0.0.imx-r0.aarch64_mx8mm

This is what I commented on in the PR #191 that weston-init recipe is now responsible to provide the weston.ini file, and since the WESTON_INI_INSTALL_FILE defines now "${B}/weston.ini" in the default recipe - this causes the above conflict, where 2 weston.ini files are provided from 2 different recipes.

I guess the weston.ini selection logic should be moved to a new `weston-init.bbappend` and not to be kept in the `weston_5.0.0.imx.bb` anymore.

thochstein:
Hi Andrey, sorry I overlooked your specific issue. Moving the logic to weston-init makes the most sense to me. For i.MX we've totally replaced the default file weston.ini in the weston fork, so no big loss there. I guess we can copy both the i.MX file and the IVI file into the appropriate Yocto layer for weston-init.

Otavio, what do you think?

otavio:
Yes; moving to `weston-init` tends tovarlds the upstream approuch so it seems as the best move.

thochstein:
Hmm, the problem with the weston-init implementation is it bysteps the weston.ini.in logic that properly sets the directory paths in the file. Is the idea not to tackle that problem at the moment?

zandrey:
Hello Tom! I think I didn't fully understood your last comment.. I just looked at the poky recipe, and it looks like `weston-init.bb` just provides the default weston.ini file committed locally next to the recipe.

My idea was: since there is already a `weston-ini.bbappend` present in the meta-freescale layer - then the logic of whether install the weston.ini or not should be moved from `weston_5.0.0.imx.bb` to that append, which I assumed was a trivial task...

If you see some additional obstacles here - can you please elaborate a bit further here?

Thanks a lot!
-- andrey

thochstein:
Right, moving what we already have to `weston-init` is fine. I just think that the upstream solution may not be the best solution. The `weston` build takes `weston.ini.in` and properly sets any paths in the options depending on the filesystem path configuration. Right now `weston-init` doesn't account for that.

zandrey:
Got it now, thanks a lot for explanation! In this case I think we are looking at 2 completely separate tasks: one being that upstream weston recipe does not account for dynamic creation of weston.ini file; and one being that current bbappend in meta-freescale does break the logic which is implemented in upstream by providing weston.ini file from the package it does not suppose to (weston vs weston-init).

I think the later could be quite trivially solved at meta-freescale layer, while former should actually be addressed at oe-core level first.

@thochstein, @otavio: What do you think about this?
Thanks a lot for continuing looking into this!
-- andrey

PR #202, opened 2019-11-15
added yocto thud version support

moorthy-bs:
Signed-off-by: Moorthy Baskar <moorthy.baskaravenkatraman-sambamoorthy@linaro.org>


otavio:
We don't intend to support those. We will support zeus only on master.

PR #203, opened 2019-11-17
Move weston.ini install to weston-init

thochstein:
Fixes a problem reported by @zandrey in PR #194, a build failure due to `weston.ini` installation by both `weston` and `weston-init` recipes. Also adds logic to reduce copying of `weston.ini` for customization.


thochstein:
Fixed an incorrect path in the `bbfatal` call of `uncomment`.
Added `#xwayland=true` to IVI version of `weston.ini`.

thochstein:
Fixed the timestamps so they don't appear out of order.

zandrey:
Hello Tom!

I've tested a non-AGL build, and with this PR is looks fine! I'll test the hw image tomorrow (imx8mm evk) and would update you with results of it.

Thanks a lot for fixing this conflict!

-- andrey

recipes-graphics/wayland/weston-init.bbappend
```diff
@@ -6,15 +6,39 @@ IMX_REQUIRED_DISTRO_FEATURES_REMOVE_imxgpu2d = "opengl"
...
+#     #xwayland=true
+# Then add the assignment to INI_UNCOMMENT_ASSIGNMENTS.
+INI_UNCOMMENT_ASSIGNMENTS = " \
+    ${@bb.utils.contains('DISTRO_FEATURES', 'x11 wayland', 'xwayland=true', '', d)} \
+"
+INI_UNCOMMENT_ASSIGNMENTS_append_mx7ulp = " \
+    use-g2d=1 \
+"
+INI_UNCOMMENT_ASSIGNMENTS_append_mx8mm = " \
+    use-g2d=1 \
```
zandrey: For imx8mm-evk, I would propose to include following lines as well:
```
\\[output\\] \
name=HDMI-A-1 \
mode=1920x1080@60 \
```
Without them, `fsl-image-multimedia-full` image built with `DISTRO="fslc-wayland"` does not come up with desktop, as the output is not enabled in the weston.ini file.

zandrey: LGTM, i.MX8M Mini EVK has a functional weston.

@thochstein: Once again, thanks a lot for solving this!
--andrey

thochstein: > For imx8mm-evk, I would propose to include following lines as well:
> 
> ```
> \\[output\\] \
> name=HDMI-A-1 \
> mode=1920x1080@60 \
> ```
> 
> Without them, `fsl-image-multimedia-full` image built with `DISTRO="fslc-wayland"` does not come up with desktop, as the output is not enabled in the weston.ini file.

Hi @zandrey, we don't want to hard code a specific display. Without an explicit display configuration, the default behavior should be to detect the active display and use it. Perhaps you're having a problem there.

PR #205, opened 2019-11-19
Updates to linux-fslc-imx_4.9 recipe

zandrey:
This PR brings an updated version of patchset required to resolve compilation issues with gcc9. Patchset has been cherry-picked from stable tree branch linux-4.9.y and effectively replaces current patchset committed by @thochstein as a part of #201. This change is done for the ease of tracking and should not create a conflict if kernel source is updated to revisions past 4.9.188 since parents are tracked.

Additional commit also solves the CVE scanner warning, which was unable to properly identify the kernel version this recipe provided.


otavio:
@zandrey I can apply this but do you mind also looking at our fslc fork and merging the newer releases so we have the kernel safe and sound?

zandrey:
> 
> @zandrey I can apply this but do you mind also looking at our fslc fork and merging the newer releases so we have the kernel safe and sound?

@otavio Sure, I can take a look! I would have one question to this one: which 4.9.y branch should be bumped up in this case? I've seen there are several branches for 4.9 kernel, namely:
- 4.9-2.3.x-imx
- 4.9-2.0.x-imx
- 4.9-1.0.x-imx
- 4.9.x+fslc

The one being actively used is `4.9-1.0.x-imx`, but this one seems to be pretty outdated... Should I try to bump that one, or should we also replace the branch in `linux-fslc.inc` recipe to another one (say `4.9-2.3.x-imx`) and bump that one instead?
I'm just not familiar with the linux-fslc branching strategy, therefore I ask it here...

-- andrey

otavio:
The `-imx` suffix are based on NXP release and map to the official GA which it is based on. The `+fslc` is the mainline one.

I'd focus on 4.9-2.3.x-imx and 4.9.x+fslc.

zandrey:
@otavio Understood, thanks a lot! I'd have a look at those 2 branches and how they could be updated.

-- andrey

zandrey:
@otavio Just one additional point to clarify here: what would be a more appropriate strategy for update here: merge or rebase?

I've completed a local rebase of both branches, but if merge would be more appropriate to accept as PR - I would try to do that instead.

Thanks!
-- andrey

otavio:
The specific fixes might be cherry-picked (using `git cherry-pick -x <hash>`). Newer stable releases, merged.

PR #212, opened 2019-11-29
fix 1909 bug 

guochunrong:


guochunrong:
please apply those patches to master

PR #213, opened 2019-12-02
Upstream kernel 4.19

panfir:


MaxKrummenacher:
master and zeus now built gcc 9.2. Since you dropped the two patches which backport the fixes for gcc 9.2 I guess that 4.19.35 will not build.
Did you actually build this patchset against master and did it build?

panfir:
> master and zeus now built gcc 9.2. Since you dropped the two patches which backport the fixes for gcc 9.2 I guess that 4.19.35 will not build.
> Did you actually build this patchset against master and did it build?

I built kernel for all release machines for 4.19 and did small images (core-image-minimal) without any issues.
I didn't check for modules, where actually were reproduced the issue about compiler attributes for __copy.
I'll recreate those two patches since are not applying anymore

zandrey:
@panfir, commit b5ad1cd breaks the build since LIC_FILES_CHKSUM does not match anymore. Looks like the COPYING file has been changed but md5 is not adjusted in this recipe.

New COPYING md5 appears to be: `bbea815ee2795b2f4230826c0c6b8814`

Can you please update it?

-- andrey

zandrey:
@panfir, I've closed the issue caused by b5ad1cd with PR #217 

-- andrey

PR #215, opened 2019-12-02
Upgrade/optee 3.7.0

clementfaure:
Add recipe for optee 3.7.0 :
 * optee-os
 * optee-client
 * optee-test

### DIFF
--- a/conf/machine/include/imx-base.inc
+++ b/conf/machine/include/imx-base.inc
@@ -267,9 +267,9 @@ PREFERRED_VERSION_libdrm_mx7 ?= "2.4.91.imx"
 PREFERRED_VERSION_libdrm_mx8 ?= "2.4.91.imx"
 
 # Use i.MX optee Version
-PREFERRED_VERSION_optee-os_mx8     ?= "3.2.0.imx"
-PREFERRED_VERSION_optee-client_mx8 ?= "3.2.0.imx"
-PREFERRED_VERSION_optee-test_mx8   ?= "3.2.0.imx"
+PREFERRED_VERSION_optee-os     ?= "3.2.0.imx"
+PREFERRED_VERSION_optee-client ?= "3.2.0.imx"
+PREFERRED_VERSION_optee-test   ?= "3.2.0.imx"
 
 # Handle default kernel
 IMX_DEFAULT_KERNEL = "linux-imx"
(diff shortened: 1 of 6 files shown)


conf/machine/include/imx-base.inc
```diff
@@ -267,9 +267,9 @@ PREFERRED_VERSION_libdrm_mx7 ?= "2.4.91.imx"
 PREFERRED_VERSION_libdrm_mx8 ?= "2.4.91.imx"
 
 # Use i.MX optee Version
-PREFERRED_VERSION_optee-os_mx8     ?= "3.2.0.imx"
-PREFERRED_VERSION_optee-client_mx8 ?= "3.2.0.imx"
-PREFERRED_VERSION_optee-test_mx8   ?= "3.2.0.imx"
+PREFERRED_VERSION_optee-os     ?= "3.2.0.imx"
+PREFERRED_VERSION_optee-client ?= "3.2.0.imx"
+PREFERRED_VERSION_optee-test   ?= "3.2.0.imx"
```
otavio: This does not seems well tested.

clementfaure: What do you mean?

clementfaure:
I moved pyelftool from internal imx layer to meta-freescale

clementfaure:
Right, ```pyelftool``` is present in OE in master but not in zeus branch ..

otavio:
> Right, the thing is that `pyelftool` is present in OE in master but not in zeus branch ..

This is not going to be included on Zeus.

clementfaure:
That's why I would like to add ```pyelftool``` in zeus branch of meta-freescale

clementfaure:
Remove ```pyelftool``` commit.
Merge it to ```zeus``` branch instead : https://github.com/Freescale/meta-freescale/pull/220

conf/machine/include/imx-base.inc
otavio: We should not carry the two versions. We should upgrade it.

So this development should be finished and send here when it is considered good for general use.

Why we cannot move master to this?

clementfaure: Optee 3.7.0 should be released very soon but not ready yet.
I'm okay with the fact you don't want two version of Optee.
I will come back when it's ready and do the full upgrade instead of having two version.
Thanks!

PR #216, opened 2019-12-03
firmware-imx: split sdma into their own packages

liuming50:
The sdma firmware of mx7d/mx6 are being installed to
${base_libdir}/firmware/imx/sdma, so they should be split into their
own packages by matching that directory.

Also change MACHINE_FIRMWARE to include these sdma firmware packages.

### DIFF
--- a/recipes-bsp/firmware-imx/firmware-imx_8.1.1.bb
+++ b/recipes-bsp/firmware-imx/firmware-imx_8.1.1.bb
@@ -60,6 +60,13 @@ python populate_packages_prepend() {
                       description='Freescale IMX Firmware %s',
                       extra_depends='',
                       prepend=True)
+
+    imxsdmadir = bb.data.expand('${base_libdir}/firmware/imx/sdma', d)
+    do_split_packages(d, imxsdmadir, '^sdma-([^-]*).*\.bin',
+                      output_pattern='firmware-imx-sdma-%s',
+                      description='Freescale IMX Firmware %s',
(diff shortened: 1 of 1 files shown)


otavio:
We are using `linux-firmware` for those. Why to move out from it?

liuming50:
@otavio I thought the firmware generated by firmware-imx are the preferred ones, they were previously not used just because they were not split correctly, but maybe I was wrong.

otavio:
> @otavio I thought the firmware generated by firmware-imx are the preferred ones, they were previously not used just because they were not split correctly, but maybe I was wrong.

We should make sure they are the same and I'd rather prefer to ask them (@thochstein) to send newer releases to linux-firmware than swap it.

That said, we should remove those files so they are not packaged. Do you mind to look at it?

liuming50:
> That said, we should remove those files so they are not packaged. Do you mind to look at it?

Or we could keep the mx7d/mx6 sdma split correctly, but just do not include them to MACHINE_FIRMWARE, wouldn't that be better?

Will make a V2, please comment on that.

liuming50:
Changes in patch set 2:

1 Drop firmware-imx-mx7d/mx6 from MACHINE_FIRMWARE.

otavio:
I'd prefer to remove the files so we ensure people do not get it deployed. Multiple providers is a support nightmare.

Also confirm if those are the same as the ones on linux-firmware @thochstein ?

PR #218, opened 2019-12-04
imx-test: update test package version to imx_4.19.35_1.1.0

zandrey:
Update the imx-test package branch to imx_4.19.35_1.1.0 which matches
the latest linux-imx kernel release (uses same branch naming).

Additionally, introduce the patch that solves trivial compilation issue
under gcc9.


zandrey:
@otavio Do you know by any chance who collects patches against `imx-test` repository? 
I see that this meta layer uses mirror from `codeaurora.org` but cannot see any origin of the repo. The patch file I just introduced here should be up-streamed, but I have no idea who I should send it to...

If you know any maintainer or a list - I would be glad to get this info! 

Thanks a lot!
-- andrey

otavio:
@thochstein please send the patch for gcc-9 internally.

thochstein:
Will do, thanks!

PR #220, opened 2019-12-06
python: add python-pyelftools

clementfaure:
```pyelftools``` is a pure-Python library for parsing and analyzing ELF files and DWARF debugging information. See the User's guide for more details.

Pyelftool has been added to OE core in ```master``` branch:
https://github.com/openembedded/meta-openembedded/commit/799925fd48cf0156caa85ef3b7f38da792fb5337

It won't be backport to OE zeus branch unfortunately.

This recipe is required by Optee 3.7.0


otavio:
Why do we need this on Zeus for now? We should work on master for next release now.

PR #221, opened 2019-12-06
Upstream qca

panfir:


recipes-bsp/firmware-qca/firmware-qca.inc
```diff
@@ -1,15 +1,15 @@
-# Copyright 2018 NXP
+# Copyright 2018-2019 NXP
 
 SUMMARY = "Qualcomm Wi-Fi and Bluetooth firmware"
 SECTION = "base"
 LICENSE = "Proprietary"
-LIC_FILES_CHKSUM = "file://COPYING;md5=6dfb32a488e5fd6bae52fbf6c7ebb086 \
+LIC_FILES_CHKSUM = "file://COPYING;md5=6c12031a11b81db21cdfe0be88cac4b3 \
```
zandrey: @panfir What is the reason MD5 is updated here? Are license terms stayed the same and it is just a License header update?

It is not obvious since it is not indicated in the commit message why this is needed.

zandrey:
@otavio @panfir Can you please address this PR? Without those modifications the build is broken. I've prepared couple of patches to address the build issue, but then I saw they are taken care of in this PR.

@panfir: There was only one question from my side regarding the License md5 update, can you please comment on this? It is quite important to understand whether License terms are changed.

Thanks a lot!

-- andrey

recipes-bsp/firmware-qca/firmware-qca9377_3.0.1.bb
```diff
@@ -12,11 +12,11 @@ inherit allarch
 do_install () {
     # Install firmware.conf for QCA modules
     install -d ${D}${sysconfdir}/bluetooth
-    install -m 644 ${S}/1PJ_QCA9377-3_LEA_2.0/etc/bluetooth/firmware.conf ${D}${sysconfdir}/bluetooth
+    install -m 644 ${S}/1PJ_QCA9377-3_LEA_3.0/etc/bluetooth/firmware.conf ${D}${sysconfdir}/bluetooth
```
zandrey: This file is also provided by `firmware-qca6174` which breaks the build when IPK is used as package manager. Even though the content of the file is identical - this needs to be addressed further, since in the machine config both `qca6174` and `qca9377` are selected via MACHINE_FEATURES.

During the boot process, only the _qca9377_ is loaded, therefore this needs further confirmation if it is safe to remove _qca6174_ from `imx8mmevk` machine.

@panfir can you please check this out and correct it before proceeding with this PR? For now I just remove the qca6174 from my machine description, and if it is the right thing to do - can you please do the same and push additional commit to this PR?

Thanks a lot!

-- andrey

zandrey:
Hello all,

I just did some additional digging and built this PR for `imx8mmevk` machine with IPK package format. It looks like that both qca* firmware recipes are providing identical bluetooth config file, which breaks the rootfs build when IPK is used with following error:

```
Downloading file:/development/yocto-master/build-output/work/imx8mmevk-fsl-linux/leica-image-multimedia-full/1.0-r0/oe-rootfs-repo/all/firmware-qca6174_3.0.1-r0_all.ipk.
Installing firmware-qca9377 (3.0.1) on root
Downloading file:/development/yocto-master/build-output/work/imx8mmevk-fsl-linux/leica-image-multimedia-full/1.0-r0/oe-rootfs-repo/all/firmware-qca9377_3.0.1-r0_all.ipk.
To remove package debris, try `opkg remove firmware-qca9377`.
To re-attempt the install, try `opkg install firmware-qca9377`.

<snap>

ConfiguringCollected errors:
 * check_data_file_clashes: Package firmware-qca9377 wants to install file /development/yocto-master/build-output/work/imx8mmevk-fsl-linux/leica-image-multimedia-full/1.0-r0/rootfs/etc/bluetooth/firmware.conf
	But that file is already provided by package  * firmware-qca6174
```
I've commented on this issue in the code of PR, hopefully this could be addressed further. For now I would locally remove the `qca6174` from my machine definition.

-- andrey

recipes-bsp/firmware-qca/firmware-qca.inc
panfir: > @panfir What is the reason MD5 is updated here? Are license terms stayed the same and it is just a License header update?
> 
> It is not obvious since it is not indicated in the commit message why this is needed.

The EULA file has been changed to August 2019 version.

recipes-bsp/firmware-qca/firmware-qca9377_3.0.1.bb
panfir: > This file is also provided by `firmware-qca6174` which breaks the build when IPK is used as package manager. Even though the content of the file is identical - this needs to be addressed further, since in the machine config both `qca6174` and `qca9377` are selected via MACHINE_FEATURES.
> 
> During the boot process, only the _qca9377_ is loaded, therefore this needs further confirmation if it is safe to remove _qca6174_ from `imx8mmevk` machine.
> 
> @panfir can you please check this out and correct it before proceeding with this PR? For now I just remove the qca6174 from my machine description, and if it is the right thing to do - can you please do the same and push additional commit to this PR?
> 
> Thanks a lot!
> 
> -- andrey

Thanks Andrey for your checking, indeed the qca6174 should be removed form MACHINE_FEATURES, we're not supporting anymore qca6174 for mm boards.
I forgot to push that commit, couple weeks ago.

panfir:
I've added also to this PR, the EULA update

zandrey:
Hello @panfir 

> 
> I've added also to this PR, the EULA update

This would require that following recipes would have MD5 updated, since the `do_populate_lic` task fails for them:
- meta-freescale/recipes-bsp/imx-vpu-hantro/imx-vpu-hantro_1.15.0.bb:do_populate_lic
- meta-freescale/recipes-graphics/imx-gpu-viv/imx-gpu-viv_6.4.0.p1.0-aarch64.bb:do_populate_lic
- meta-freescale/recipes-bsp/firmware-imx/firmware-imx_8.5.bb:do_populate_lic
- meta-freescale/recipes-bsp/firmware-imx/firmware-imx-8m_8.5.bb:do_populate_lic

Can you please also add those to this PR?

Thanks a lot!

-- andrey

recipes-bsp/firmware-qca/firmware-qca9377_3.0.1.bb
zandrey: @panfir: thanks for your confirmation here!

panfir:
> Hello @panfir
> 
> > I've added also to this PR, the EULA update
> 
> This would require that following recipes would have MD5 updated, since the `do_populate_lic` task fails for them:
> 
> * meta-freescale/recipes-bsp/imx-vpu-hantro/imx-vpu-hantro_1.15.0.bb:do_populate_lic
> * meta-freescale/recipes-graphics/imx-gpu-viv/imx-gpu-viv_6.4.0.p1.0-aarch64.bb:do_populate_lic
> * meta-freescale/recipes-bsp/firmware-imx/firmware-imx_8.5.bb:do_populate_lic
> * meta-freescale/recipes-bsp/firmware-imx/firmware-imx-8m_8.5.bb:do_populate_lic
> 
> Can you please also add those to this PR?
> 
> Thanks a lot!
> 
> -- andrey

In fact I need to update fsl-eula-unpack class

zandrey:
@panfir I've just took this PR as a patch and my local build ran without any issues. I cannot test the runtime now since I'm using a different machine, so I'd have to leave this up to you to verify.

Thanks for addressing this!

-- andrey

panfir:
It should be fine now.

panfir:
qca is building without any issues, also I checked with a clean full image (fsl-image-multimedia-full).
So in this PR, qca has been updated, one of the drivers has been dropped due to the fact that qca6174 is no longer supported since 4.19. Updated as well EULA and the bbclass.
@otavio, please let me know about any other concern.

PR #223, opened 2019-12-10
firmware-imx: Upgrade to version 8.5

panfir:
- SECO split from firmware-imx package
- Exclude "easrc" as this IP is only applicable for Nano Soc
- Remove imx-firmware.git from SRC_URI since Broadcom fw is no longer used

Signed-off-by: Cristinel Panfir <cristinel.panfir@nxp.com>


otavio:
Where the SECO firmware ended? It wasn't added anywhere.

Also, people still use the Broadcom and even though NXP has moved to another vendor there are customers which still make use of it and we cannot just drop it.

panfir:
> Where the SECO firmware ended? It wasn't added anywhere.
> 
> Also, people still use the Broadcom and even though NXP has moved to another vendor there are customers which still make use of it and we cannot just drop it.

SECO was moved to its own recipe and imx-boot will depend on it, I'll push it shortly.
Regarding Broadcom, who needs that will need to create a bbappend to add that SRC_URI along with installs for each BCM (binaries and configurations). This recipe is not installing anymore broadcom fw.

otavio:
> SECO was moved to its own recipe and imx-boot will depend on it, I'll push it shortly.

We will not apply this dropping this feature, so please add the new recipe to this PR.

> Regarding Broadcom, who needs that will need to create a bbappend to add that SRC_URI along with installs for each BCM (binaries and configurations). This recipe is not installing anymore broadcom fw.

We are installing as I did not apply this PR and will not apply it removing the firmwares. NXP can apply a `.bbappend` to drop them if they want.

mihaix:
> > Regarding Broadcom, who needs that will need to create a bbappend to add that SRC_URI along with installs for each BCM (binaries and configurations). This recipe is not installing anymore broadcom fw.
> 
> We are installing as I did not apply this PR and will not apply it removing the firmwares. NXP can apply a `.bbappend` to drop them if they want.

Why clone another git tree if it's no longer used? Maybe there's something I'm missing, but I don't see files copied or packaged from NXP/imx-firmware.git tree.

panfir:
> > SECO was moved to its own recipe and imx-boot will depend on it, I'll push it shortly.
> 
> We will not apply this dropping this feature, so please add the new recipe to this PR.
> 
> > Regarding Broadcom, who needs that will need to create a bbappend to add that SRC_URI along with installs for each BCM (binaries and configurations). This recipe is not installing anymore broadcom fw.
> 
> We are installing as I did not apply this PR and will not apply it removing the firmwares. NXP can apply a `.bbappend` to drop them if they want.

I've updated the recipe and kept the imx-firmware.git project.

PR #260, opened 2020-01-13
OP-TEE fixes

petegriffin:
Fix the build, as the new tag for imx optee forks already contains these patches so they no longer apply.


otavio:
Please rebase

PR #268, opened 2020-01-21
Jailhouse

panfir:
Enable Jailhouse for imx8mq and imx8mm.
Machine confs are already updated with the inmate .dtbs

### DIFF
--- a/conf/machine/include/imx-base.inc
+++ b/conf/machine/include/imx-base.inc
@@ -255,6 +255,11 @@ MACHINE_EXTRA_RRECOMMENDS_append = " ${@bb.utils.contains('MACHINE_FEATURES', 'q
 # Extra udev rules
 MACHINE_EXTRA_RRECOMMENDS += "udev-rules-imx"
 
+# Jailhouse
+MACHINE_EXTRA_RRECOMMENDS += " \
+    ${@bb.utils.contains('COMBINED_FEATURES', 'jailhouse', 'jailhouse', '', d)} \
+"
+
 # GStreamer 1.0 plugins
 MACHINE_GSTREAMER_1_0_PLUGIN ?= ""
 MACHINE_GSTREAMER_1_0_PLUGIN_mx6dl ?= "gstreamer1.0-plugins-imx-meta"
@@ -375,6 +380,9 @@ KERNEL_IMAGETYPE = "zImage"
 KERNEL_IMAGETYPE_aarch64 = "Image"
 
 MACHINE_FEATURES = "usbgadget usbhost vfat alsa touchscreen"
+
+SOC_MACHINE_FEATURES_BACKFILL_mx8m = "jailhouse"
+MACHINE_FEATURES_BACKFILL_append = " ${SOC_MACHINE_FEATURES_BACKFILL}"
 
 # Add the ability to specify _imx machines
 MACHINEOVERRIDES =. "imx:"
--- a/recipes-extended/jailhouse/jailhouse_0.11.bb
+++ b/recipes-extended/jailhouse/jailhouse_0.11.bb
@@ -1,0 +1,122 @@
+SUMMARY = "Jailhouse, i.MX fork"
+HOMEPAGE = "https://github.com/siemens/jailhouse"
+SECTION = "jailhouse"
+LICENSE = "GPL-2.0"
+
+LIC_FILES_CHKSUM = "file://COPYING;md5=9fa7f895f96bde2d47fd5b7d95b6ba4d \
+                 file://tools/root-cell-config.c.tmpl;beginline=6;endline=33;md5=2825581c1666c44a17955dc574cfbfb3 \
+                 file://include/jailhouse/hypercall.h;beginline=9;endline=36;md5=2825581c1666c44a17955dc574cfbfb3 \
+                 file://include/jailhouse/cell-config.h;beginline=9;endline=36;md5=2825581c1666c44a17955dc574cfbfb3 \
+                 file://include/arch/arm/asm/jailhouse_hypercall.h;beginline=9;endline=36;md5=2825581c1666c44a17955dc574cfbfb3 \
+                 file://include/arch/arm64/asm/jailhouse_hypercall.h;beginline=9;endline=36;md5=2825581c1666c44a17955dc574cfbfb3 \
+                 file://include/arch/x86/asm/jailhouse_hypercall.h;beginline=9;endline=36;md5=2825581c1666c44a17955dc574cfbfb3 \
+                 file://driver/jailhouse.h;beginline=9;endline=36;md5=2825581c1666c44a17955dc574cfbfb3 \
+"
+
+SRCBRANCH = "lf-5.4.y"
+SRCREV = "82d5698945eafaffa58bab3b4d6da320bf56289c"
+
+SRC_URI = "git://source.codeaurora.org/external/imx/imx-jailhouse.git;protocol=ssh;branch=${SRCBRANCH} \
+           file://0001-tools-scripts-update-shebang-to-python3.patch \
+"
+
+DEPENDS = " \
+    make-native \
+    python3-mako-native \
+    python3-mako \
+    dtc-native \
+"
+
+inherit module python3native bash-completion deploy setuptools3
+
+S = "${WORKDIR}/git"
+B = "${S}"
+
+JH_ARCH = "arm64"
+JH_DATADIR ?= "${datadir}/jailhouse"
+CELL_DIR ?= "${JH_DATADIR}/cells"
+CELLCONF_DIR ?= "${JH_DATADIR}/configs"
+INMATES_DIR ?= "${JH_DATADIR}/inmates"
+
+JH_CONFIG ?= "${S}/ci/jailhouse-config-x86.h"
+JH_CONFIG_x86 ?= "${S}/ci/jailhouse-config-x86.h"
+JH_CONFIG_x86-64 ?= "${S}/ci/jailhouse-config-x86.h"
+JH_CONFIG_arm ?= "${S}/ci/jailhouse-config-banana-pi.h"
+
+do_configure() {
+   if [ -d ${STAGING_DIR_HOST}/${CELLCONF_DIR} ];
+   then
+      cp "${STAGING_DIR_HOST}/${CELLCONF_DIR}/"*.c ${S}/configs/${ARCH}/
+   fi
+}
+
+USER_SPACE_CFLAGS = '${CFLAGS} -DLIBEXECDIR=\\\"${libexecdir}\\\" \
+		  -DJAILHOUSE_VERSION=\\\"$JAILHOUSE_VERSION\\\" \
+		  -Wall -Wextra -Wmissing-declarations -Wmissing-prototypes -Werror \
+		  -I../driver'
+
+TOOLS_SRC_DIR = "${S}/tools"
+TOOLS_OBJ_DIR = "${S}/tools"
+
+do_compile() {
+    unset LDFLAGS
+    oe_runmake V=1 CC="${CC}" \
+        ARCH=${JH_ARCH} CROSS_COMPILE=${TARGET_PREFIX} \
+        KDIR=${STAGING_KERNEL_BUILDDIR}
+
+    cd ${TOOLS_SRC_DIR}
+    export JAILHOUSE_VERSION=$(cat ../VERSION)
+    oe_runmake V=1 \
+        CFLAGS="${USER_SPACE_CFLAGS}" \
+        src=${TOOLS_SRC_DIR} obj=${TOOLS_OBJ_DIR} \
+        ${TOOLS_OBJ_DIR}/jailhouse-config-collect ${TOOLS_OBJ_DIR}/jailhouse
+}
+
+do_install() {
+    oe_runmake \
+        PYTHON=python3 \
+        V=1 \
+        LDFLAGS="" \
+        CC="${CC}" \
+        ARCH=${JH_ARCH} \
+        CROSS_COMPILE=${TARGET_PREFIX} \
+        KDIR=${STAGING_KERNEL_BUILDDIR} \
+        DESTDIR=${D} install
+
+    install -d ${D}${CELL_DIR}
+    install ${B}/configs/${JH_ARCH}/*.cell ${D}${CELL_DIR}/
+
+    install -d ${D}${INMATES_DIR}/tools/${JH_ARCH}
+    install ${B}/inmates/demos/${JH_ARCH}/*.bin ${D}${INMATES_DIR}
+
+    install -d ${D}${JH_DATADIR}/tools
+    install ${B}/tools/jailhouse-cell-linux ${D}${JH_DATADIR}/tools
+    install ${B}/tools/jailhouse-cell-stats ${D}${JH_DATADIR}/tools
+    install ${B}/tools/jailhouse-config-collect ${D}${JH_DATADIR}/tools
+    install ${B}/tools/jailhouse-config-create ${D}${JH_DATADIR}/tools
+    install ${B}/tools/jailhouse-gcov-extract ${D}${JH_DATADIR}/tools
+    install ${B}/tools/jailhouse-hardware-check ${D}${JH_DATADIR}/tools
+    install ${B}/inmates/tools/${JH_ARCH}/linux-loader.bin ${D}${INMATES_DIR}/tools/${JH_ARCH}
+}
+
+PACKAGE_BEFORE_PN = "kernel-module-jailhouse pyjailhouse"
+
+FILES_${PN} += "${base_libdir}/firmware ${libexecdir} ${sbindir} ${JH_DATADIR}"
+FILES_pyjailhouse = "${PYTHON_SITEPACKAGES_DIR}"
+
+RDEPENDS_${PN} += " \
+    python3-curses \
+    python3-datetime \
+    python3-mmap \
+"
+
+RDEPENDS_pyjailhouse = " \
+    python3-core \
+    python3-ctypes \
+    python3-fcntl \
+    python3-shell \
+"
+
+INSANE_SKIP_${PN} = "ldflags"
+
+COMPATIBLE_MACHINE = "(mx8m)"
--- a/recipes-extended/jailhouse/jailhouse/0001-tools-scripts-update-shebang-to-python3.patch
+++ b/recipes-extended/jailhouse/jailhouse/0001-tools-scripts-update-shebang-to-python3.patch
@@ -1,0 +1,69 @@
+From f07cd0e5f32a00a4091fa964d0dae0cd46c17554 Mon Sep 17 00:00:00 2001
+From: Cristinel Panfir <cristinel.panfir@nxp.com>
+Date: Sat, 23 May 2020 08:51:31 +0000
+Subject: [PATCH] tools/scripts: update shebang to python3
+
+Upstream-Status: Pending
+
+Signed-off-by: Cristinel Panfir <cristinel.panfir@nxp.com>
+---
+ scripts/arm64-parsedump.py     | 2 +-
+ tools/jailhouse-cell-linux     | 2 +-
+ tools/jailhouse-cell-stats     | 2 +-
+ tools/jailhouse-config-create  | 2 +-
+ tools/jailhouse-hardware-check | 2 +-
+ 5 files changed, 5 insertions(+), 5 deletions(-)
+
+diff --git a/scripts/arm64-parsedump.py b/scripts/arm64-parsedump.py
+index 9519eb5..c695706 100755
+--- a/scripts/arm64-parsedump.py
++++ b/scripts/arm64-parsedump.py
+@@ -1,4 +1,4 @@
+-#!/usr/bin/env python
++#!/usr/bin/env python3
+ 
+ # Jailhouse, a Linux-based partitioning hypervisor
+ #
+diff --git a/tools/jailhouse-cell-linux b/tools/jailhouse-cell-linux
+index 215932a..bcbb902 100755
+--- a/tools/jailhouse-cell-linux
++++ b/tools/jailhouse-cell-linux
+@@ -1,4 +1,4 @@
+-#!/usr/bin/env python
++#!/usr/bin/env python3
+ 
+ # Jailhouse, a Linux-based partitioning hypervisor
+ #
+diff --git a/tools/jailhouse-cell-stats b/tools/jailhouse-cell-stats
+index 2f7e966..4c5289f 100755
+--- a/tools/jailhouse-cell-stats
++++ b/tools/jailhouse-cell-stats
+@@ -1,4 +1,4 @@
+-#!/usr/bin/env python
++#!/usr/bin/env python3
+ 
+ # Jailhouse, a Linux-based partitioning hypervisor
+ #
+diff --git a/tools/jailhouse-config-create b/tools/jailhouse-config-create
+index 55601a6..9869b94 100755
+--- a/tools/jailhouse-config-create
++++ b/tools/jailhouse-config-create
+@@ -1,4 +1,4 @@
+-#!/usr/bin/env python
++#!/usr/bin/env python3
+ #
+ # Jailhouse, a Linux-based partitioning hypervisor
+ #
+diff --git a/tools/jailhouse-hardware-check b/tools/jailhouse-hardware-check
+index b86756a..0c8fb42 100755
+--- a/tools/jailhouse-hardware-check
++++ b/tools/jailhouse-hardware-check
+@@ -1,4 +1,4 @@
+-#!/usr/bin/env python
++#!/usr/bin/env python3
+ 
+ # Jailhouse, a Linux-based partitioning hypervisor
+ #
+-- 
+2.7.4
+


recipes-extended/jailhouse/jailhouse-defs.inc
```diff
@@ -0,0 +1,25 @@
...
+
+    if   re.match('(i.86|athlon|x86.64)$', a):  return 'x86'
+    elif re.match('armeb$', a):                 return 'arm'
+    elif re.match('aarch64$', a):               return 'arm64'
+    elif re.match('aarch64_be$', a):            return 'arm64'
+    elif a in valid_jh_archs:                   return a
+    else:
+        bb.error("cannot map '%s' to a jailhouse supported architecture" % a)
+
+export JH_ARCH = "${@map_jh_arch(d.getVar('TARGET_ARCH', True), d)}"
```
otavio: This does not belong to a .inc file, instead, you should add it as a class. I'd name it as `jailhouse-arch` or similar.

recipes-extended/jailhouse/jailhouse_0.10.bb
```diff
@@ -0,0 +1,11 @@
...
+
+SRCBRANCH = "imx_4.19.35_1.1.0"
+IMX_JAILHOUSE_SRC ?= "git://source.codeaurora.org/external/imx/imx-jailhouse.git;protocol=ssh"
+
+SRC_URI = "${IMX_JAILHOUSE_SRC};branch=${SRCBRANCH}"
+SRCREV = "e83ab629ccc6f44e4caa05bc51c66d2f86546ee6" 
+
+CELLS = ""
+
+COMPATIBLE_MACHINE = "${@bb.utils.contains('MACHINE_FEATURES', 'jailhouse', '${MACHINE}', '(^$)', d)}"
```
otavio: No reason for a .inc and a .bb for such small difference. Please squash the .inc and .bb so it is easier to follow and maintain.

conf/machine/include/imx-base.inc
```diff
@@ -353,6 +357,9 @@ KERNEL_IMAGETYPE_aarch64 = "Image"
 
 MACHINE_FEATURES = "usbgadget usbhost vfat alsa touchscreen"
 
+MACHINE_FEATURES_append_imx8mqevk = " jailhouse"
+MACHINE_FEATURES_append_imx8mmevk = " jailhouse"
```
zandrey: @panfir How about the support for `imx8mnevk`?

Commit aa8331679397a0ee699070e1de69252ccfbbf41b in the source repo suggests that the support for Nano is provided.

Can you please include it also here, since we already have Nano machine?

Thanks!

-- andrey

recipes-extended/jailhouse/jailhouse.inc
```diff
@@ -0,0 +1,104 @@
...
+
+require jailhouse-defs.inc
+inherit module pythonnative bash-completion setuptools
+
+JH_CONFIG ?= "${S}/ci/jailhouse-config-x86.h"
+JH_CONFIG_x86 ?= "${S}/ci/jailhouse-config-x86.h"
+JH_CONFIG_x86-64 ?= "${S}/ci/jailhouse-config-x86.h"
+JH_CONFIG_arm ?= "${S}/ci/jailhouse-config-banana-pi.h"
+JH_CONFIG_imx8mqevk = "${S}/ci/jailhouse-config-imx8mq-evk.h"
+JH_CONFIG_imx8mmevk = "${S}/ci/jailhouse-config-imx8mm-evk.h"
```
zandrey: Same applies here I guess - Nano support is missing...

recipes-extended/jailhouse/jailhouse-defs.inc
panfir: Yes, it makes more sense. In fact we inherited these .inc files in our fork.

recipes-extended/jailhouse/jailhouse_0.10.bb
panfir: Done

conf/machine/include/imx-base.inc
panfir: That is right, there was added support for the entire i.MX8M family, but I'm not sure how well it was tested for 8MN, I will come back here.
Do you have a certain use case for imx8mn? If so please let me know.

zandrey: > That is right, there was added support for the entire i.MX8M family, but I'm not sure how well it was tested for 8MN, I will come back here.

That would be really nice, thanks a lot!

> Do you have a certain use case for imx8mn? If so please let me know.

For the moment not, but when I saw that you've added support for both MQ and MM, leaving MN on the side I was wondering if there is something missing in the Nano part of the build... Since the layer contains Nano machine description and the source code for Nano is in jailhouse repo - it would be better to have the support here as well, so that people which are using Nano as a platform and interested in jailhouse hypervisor could just build it with those recipes. If you would be able to provide this support here with this PR - that would be really great!

-- andrey

recipes-extended/jailhouse/jailhouse_0.10.bb
```diff
@@ -0,0 +1,116 @@
...
+SRC_URI = "${IMX_JAILHOUSE_SRC};branch=${SRCBRANCH}"
+SRCREV = "e83ab629ccc6f44e4caa05bc51c66d2f86546ee6"
+
+DEPENDS = "virtual/kernel make-native \
+        python-mako-native python-mako \
+        dtc-native "
+
+S ="${WORKDIR}/git"
+
+inherit module pythonnative bash-completion setuptools jailhouse-arch
```
zandrey: @panfir This would not built with current `master` as python2 support is dropped.

Can you please change `pythonnative` and `setuptools` to python3 alternatives? Also please review DEPENDS above as this requires the python3 packages now.

-- andrey

recipes-extended/jailhouse/jailhouse_0.10.bb
```diff
@@ -18,13 +18,20 @@ IMX_JAILHOUSE_SRC ?= "git://source.codeaurora.org/external/imx/imx-jailhouse.git
 SRC_URI = "${IMX_JAILHOUSE_SRC};branch=${SRCBRANCH}"
 SRCREV = "e83ab629ccc6f44e4caa05bc51c66d2f86546ee6"
 
-DEPENDS = "virtual/kernel make-native \
-        python-mako-native python-mako \
-        dtc-native "
+DEPENDS = "virtual/kernel make-native\
+    python3-mako-native python3-mako\
+    dtc-native\
+"
```
otavio:
```suggestion
DEPENDS = " \
    virtual/kernel \
    make-native\
    python3-mako-native \
    python3-mako \
    dtc-native \
"
```

recipes-extended/jailhouse/jailhouse_0.10.bb
```diff
@@ -55,23 +62,28 @@ USER_SPACE_CFLAGS = '${CFLAGS} -DLIBEXECDIR=\\\"${libexecdir}\\\" \
 TOOLS_SRC_DIR = "${S}/tools"
 TOOLS_OBJ_DIR = "${S}/tools"
 
+B = "${S}"
+
```
otavio: Why? Couldn't this to be fixed?

recipes-extended/jailhouse/jailhouse_0.10.bb
```diff
@@ -55,23 +62,28 @@ USER_SPACE_CFLAGS = '${CFLAGS} -DLIBEXECDIR=\\\"${libexecdir}\\\" \
...
 do_compile() {
+
+   ln -sf ${PYTHON} ${STAGING_BINDIR_NATIVE}/python
```
otavio: Why? We should get the using the right python without linking this.

recipes-extended/jailhouse/jailhouse_0.10.bb
panfir: > Why? Couldn't this to be fixed?

I'll check again, at a first look wasn't

recipes-extended/jailhouse/jailhouse_0.10.bb
panfir: > Why? We should get the using the right python without linking this.

Why, is in commit message.
So you suggest to change into the scripts rather than use a link?

panfir:
Actually I will come back with this PR after 5.4 integration.
Anyways, it took too looong

recipes-extended/jailhouse/jailhouse_0.10.bb
otavio: Yes, because linking is easy to get wrong.

recipes-extended/jailhouse/jailhouse_0.11.bb
```diff
@@ -0,0 +1,128 @@
...
+
+SRCBRANCH = "lf-5.4.y"
+IMX_JAILHOUSE_SRC ?= "git://source.codeaurora.org/external/imx/imx-jailhouse.git;protocol=ssh"
+
+SRC_URI = "${IMX_JAILHOUSE_SRC};branch=${SRCBRANCH} \
+           file://0001-scripts-tools-switch-to-python3.patch \
+"
+SRCREV = "82d5698945eafaffa58bab3b4d6da320bf56289c"
+
+DEPENDS = "virtual/kernel \
```
otavio, line 24: This should not depend on kernel as the module does this correctly.

recipes-extended/jailhouse/jailhouse_0.11.bb
```diff
@@ -0,0 +1,123 @@
...
+    python3-mmap \
+"
+
+S = "${WORKDIR}/git"
+
+inherit module python3native bash-completion deploy setuptools3 jailhouse-arch
+
+COMPATIBLE_MACHINE = "${@bb.utils.contains('MACHINE_FEATURES', 'jailhouse', '${MACHINE}', '(^$)', d)}"
+
+CELLS = ""
```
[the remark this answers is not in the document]
panfir: yes, we can get rid off that option, we do not use other cell recipes

recipes-extended/jailhouse/jailhouse_0.11.bb
panfir: I think you are right.

recipes-extended/jailhouse/jailhouse_0.10.bb
panfir: I'd rather keep this, since build is looking for setup.py which is present in {S}

conf/machine/imx8mnevk.conf
```diff
@@ -1,6 +1,6 @@
 #@TYPE: Machine
 #@NAME: NXP i.MX8M Nano DDR4 Evaluation Kit
-#@SOC: i.MX8MM
+#@SOC: i.MX8MN
```
zandrey, line 3: Nice to be corrected, but since this is the only change in this file and it is not related to the Jailhouse - maybe it makes sense to drop this from the PR.

I'll deal with this typo later, thanks for pointing it out! 👍

recipes-extended/jailhouse/jailhouse/0001-scripts-tools-switch-to-python3.patch
```diff
@@ -0,0 +1,81 @@
+From c9fc6bbd535d0d92578cf3258c6db052ec6094dd Mon Sep 17 00:00:00 2001
+From: Cristinel Panfir <cristinel.panfir@nxp.com>
+Date: Tue, 28 Apr 2020 13:59:21 +0300
+Subject: [PATCH] scripts/tools: switch to python3
+
+Signed-off-by: Cristinel Panfir <cristinel.panfir@nxp.com>
+---
```
zandrey, line 7: Upstream-Status is missing in the patch file. Is it a backport or this patch committed upstream>

recipes-extended/jailhouse/jailhouse_0.11.bb
```diff
@@ -0,0 +1,115 @@
...
+                 file://include/jailhouse/hypercall.h;beginline=9;endline=36;md5=2825581c1666c44a17955dc574cfbfb3 \
+                 file://include/jailhouse/cell-config.h;beginline=9;endline=36;md5=2825581c1666c44a17955dc574cfbfb3 \
+                 file://include/arch/arm/asm/jailhouse_hypercall.h;beginline=9;endline=36;md5=2825581c1666c44a17955dc574cfbfb3 \
+                 file://include/arch/arm64/asm/jailhouse_hypercall.h;beginline=9;endline=36;md5=2825581c1666c44a17955dc574cfbfb3 \
+                 file://include/arch/x86/asm/jailhouse_hypercall.h;beginline=9;endline=36;md5=2825581c1666c44a17955dc574cfbfb3 \
+                 file://driver/jailhouse.h;beginline=9;endline=36;md5=2825581c1666c44a17955dc574cfbfb3 \
+"
+
+SRCBRANCH = "lf-5.4.y"
+IMX_JAILHOUSE_SRC ?= "git://source.codeaurora.org/external/imx/imx-jailhouse.git;protocol=ssh"
```
zandrey, line 17: Just for clarity: why did you extracted `SRCBRANCH` and `IMX_JAILHOUSE_SRC` and didn't put them directly in `SRC_URI`?

Do you intend to override them later?
In that case, the `SRCBRANCH` should be weakly-defined (like `IMX_JAILHOUSE_SRC`).

conf/machine/include/imx-base.inc
```diff
@@ -332,6 +332,10 @@ IMAGE_BOOT_FILES ?= " \
     ${@bb.utils.contains('MACHINE_FEATURES', 'optee', '${OPTEE_BOOT_IMAGE}', '', d)} \
 "
 
+IMAGE_INSTALL_append = " \
+    ${@bb.utils.contains('COMBINED_FEATURES', 'jailhouse', 'jailhouse', '', d)} \
+"
```
zandrey, line 337: `IMAGE_INSTALL` in machine files is usually not such a great idea to have (even if it is conditional). Can this be moved out?

conf/machine/imx8mnevk.conf
panfir: Yes, that's right. He stayed after squash. I had to check machine config first since 8mn is supporting jailhouse, so I caught that typo.

recipes-extended/jailhouse/jailhouse/0001-scripts-tools-switch-to-python3.patch
panfir: It's not a backport or upstream submitted, the origin has not been updated to support newer user space apps, which this patch is referring to.
OK, it can be set to Pending.

recipes-extended/jailhouse/jailhouse_0.11.bb
panfir: Indeed, for override. It is easier to maintain recipes when we develop software internally, then publish it externally to different mirrors.
So it's not weak for branch, will keep it as is.
Anyway, it doesn't look that bad for me, it's pretty clear.

conf/machine/include/imx-base.inc
panfir: I'm not sure where is the best place ..but as you already noticed, it is conditional.

zandrey: > I'm not sure where is the best place ..but as you already noticed, it is conditional.

Conditional is exactly the problematic part... What if I want to have only one image recipe with `jailhouse` included? If I set a feature - all my image recipes would have `jailhouse` which would not be my intention.

I guess this line could be completely removed, and let developer to decide which image needs `jailhouse`.

otavio: It can go to `MACHINE_EXTRA_RRECOMMENDS` so this is included on the image but can be filtered if needed.

recipes-extended/jailhouse/jailhouse_0.11.bb
zandrey: > It is easier to maintain recipes when we develop software internally, then publish it externally to different mirrors.

Got it. So you override this in your development environment to internal GIT and use a weak definition in the recipe.

> So it's not weak for branch, will keep it as is.

But I guess in this case it does not hurt to convert it to a weak definition as well, just in case if you would like to switch the branch internally as well. This would also bring consistency here, IMHO.

> Anyway, it doesn't look that bad for me, it's pretty clear.

I didn't say it's bad, was just a bit unclear why URI is weak and BRANCH is not.

zandrey:
General question: @panfir Siemens has a newer version of `jailhouse` published on [their GitHub](https://github.com/siemens/jailhouse/releases/tag/v0.12). Does it make sense to upgrade the fork to that version? `v0.11` is from July 2019 which is almost one year old...

recipes-extended/jailhouse/jailhouse_0.11.bb
panfir: OK, not a problem

panfir: actually, we can go without weak definitions

conf/machine/include/imx-base.inc
panfir: Simply, like `MACHINE_EXTRA_RRECOMMENDS = " jailhouse"` ?

panfir: or `MACHINE_EXTRA_RRECOMMENDS_append_mx8 = " jailhouse"`

panfir:
> General question: @panfir Siemens has a newer version of `jailhouse` published on [their GitHub](https://github.com/siemens/jailhouse/releases/tag/v0.12). Does it make sense to upgrade the fork to that version? `v0.11` is from July 2019 which is almost one year old...

Jailhouse has been updated internally to `v0.12` some time ago, I guess will be pushed next release, it wasn't ready at 5.4.3-1.0.0 release time

conf/machine/include/imx-base.inc
otavio: no, it needs to be conditional to the combined features.

panfir:
Guys, do you think something needs to be changed here?

panfir:
I tested with core-image-base to see if it drop the package from image in case not fulfill the conditions.

otavio:
Is this ready for review? if so, please squash the commit and check the commit log.

panfir:
> Is this ready for review? if so, please squash the commit and check the commit log.

Done

recipes-extended/jailhouse/jailhouse_0.11.bb
```diff
@@ -0,0 +1,115 @@
...
+
+PACKAGE_BEFORE_PN = "kernel-module-jailhouse pyjailhouse"
+FILES_${PN} += "${base_libdir}/firmware ${libexecdir} ${sbindir} ${JH_DATADIR}"
+FILES_pyjailhouse = "${PYTHON_SITEPACKAGES_DIR}"
+
+RDEPENDS_pyjailhouse = "python3-core python3-ctypes python3-fcntl python3-shell"
+
+INSANE_SKIP_${PN} = "ldflags"
+
+COMPATIBLE_MACHINE = "${@bb.utils.contains('MACHINE_FEATURES', 'jailhouse', '${MACHINE}', '(^$)', d)}"
```
otavio, line 115: Use explicit:

```
COMPATIBLE_MACHINE = "(mx8m)"
```

So we avoid unnecessary complexity.

panfir: > Use explicit:
> 
> ```
> COMPATIBLE_MACHINE = "(mx8m)"
> ```
> 
> So we avoid unnecessary complexity.

That's right.
Done

conf/machine/include/imx-base.inc
```diff
@@ -376,6 +381,8 @@ KERNEL_IMAGETYPE_aarch64 = "Image"
 
 MACHINE_FEATURES = "usbgadget usbhost vfat alsa touchscreen"
 
+MACHINE_FEATURES_append_mx8m = " jailhouse"
```
otavio, line 384: I was thinking about this and I believe we ought to use another mechanism for this. Please consider:

```
SOC_MACHINE_FEATURES_BACKFILL_mx8m = "jailhouse"
MACHINE_FEATURES_BACKFILL += "${SOC_MACHINE_FEATURES_BACKFILL}"
```

panfir: Okay, I'm gonna test this.
So `MACHINE_FEATURES` will become an append of `MACHINE_FEATURES_BACKFILL`.

recipes-extended/jailhouse/jailhouse_0.11.bb
```diff
@@ -0,0 +1,115 @@
...
+    install ${B}/tools/jailhouse-gcov-extract ${D}${JH_DATADIR}/tools
+    install ${B}/tools/jailhouse-hardware-check ${D}${JH_DATADIR}/tools
+    install ${B}/inmates/tools/${JH_ARCH}/linux-loader.bin ${D}${INMATES_DIR}/tools/${JH_ARCH}
+}
+
+PACKAGE_BEFORE_PN = "kernel-module-jailhouse pyjailhouse"
+FILES_${PN} += "${base_libdir}/firmware ${libexecdir} ${sbindir} ${JH_DATADIR}"
+FILES_pyjailhouse = "${PYTHON_SITEPACKAGES_DIR}"
+
+RDEPENDS_pyjailhouse = "python3-core python3-ctypes python3-fcntl python3-shell"
```
otavio, line 111: Please move the other `RDEPENDS` here as well as this is related to the runtime settings.

recipes-extended/jailhouse/jailhouse_0.11.bb
```diff
@@ -0,0 +1,115 @@
...
+    install ${B}/tools/jailhouse-cell-linux ${D}${JH_DATADIR}/tools
+    install ${B}/tools/jailhouse-cell-stats ${D}${JH_DATADIR}/tools
+    install ${B}/tools/jailhouse-config-collect ${D}${JH_DATADIR}/tools
+    install ${B}/tools/jailhouse-config-create ${D}${JH_DATADIR}/tools
+    install ${B}/tools/jailhouse-gcov-extract ${D}${JH_DATADIR}/tools
+    install ${B}/tools/jailhouse-hardware-check ${D}${JH_DATADIR}/tools
+    install ${B}/inmates/tools/${JH_ARCH}/linux-loader.bin ${D}${INMATES_DIR}/tools/${JH_ARCH}
+}
+
+PACKAGE_BEFORE_PN = "kernel-module-jailhouse pyjailhouse"
```
otavio, line 107: Newline after this definition.

otavio:
@zandrey @rehsack @fbertux please take a look.

recipes-extended/jailhouse/jailhouse/0001-scripts-tools-switch-to-python3.patch
```diff
@@ -0,0 +1,83 @@
...
+diff --git a/scripts/include.mk b/scripts/include.mk
+index 028f0ab..a8c94b6 100644
+--- a/scripts/include.mk
++++ b/scripts/include.mk
+@@ -42,7 +42,7 @@ INSTALL_PROGRAM ?= $(INSTALL)
+ INSTALL_DATA    ?= $(INSTALL) -m 644
+ INSTALL_DIR     ?= $(INSTALL) -d -m 755
+ 
+-PYTHON ?= python
++PYTHON = /usr/bin/env python3
```
zandrey, line 37: Is it OK to re-define conditional assignment here? If yes - then why was that in upstream, and would upstream guys accept the conditional assignment drop?

My concern is: there might be some further discussions with upstream here about this patch, since it still has `Pending` state.

recipes-extended/jailhouse/jailhouse_0.11.bb
```diff
@@ -0,0 +1,121 @@
+SUMMARY = "Jailhouse, i.MX fork"
```
zandrey, line 1: Do you maybe need/want to put `Copyright` and `SPDX-License-Identifier` in the recipe?

recipes-extended/jailhouse/jailhouse_0.11.bb
```diff
@@ -0,0 +1,121 @@
...
+                 file://include/arch/x86/asm/jailhouse_hypercall.h;beginline=9;endline=36;md5=2825581c1666c44a17955dc574cfbfb3 \
+                 file://driver/jailhouse.h;beginline=9;endline=36;md5=2825581c1666c44a17955dc574cfbfb3 \
+"
+
+SRCBRANCH = "lf-5.4.y"
+IMX_JAILHOUSE_SRC = "git://source.codeaurora.org/external/imx/imx-jailhouse.git;protocol=ssh"
+
+SRC_URI = "${IMX_JAILHOUSE_SRC};branch=${SRCBRANCH} \
+           file://0001-scripts-tools-switch-to-python3.patch \
+"
```
zandrey, line 21: Defining `IMX_JAILHOUSE_SRC` unconditional and using it in adjacent `SRC_URI` is a bit pointless IMHO, as it cannot be overridden.

It is maybe easier to do one of the following:
1. Define `IMX_JAILHOUSE_SRC` conditional and default to `codeaurora.org`
2. Include `codeaurora.org` directly into `SRC_URI` and drop `IMX_JAILHOUSE_SRC` completely

classes/jailhouse-arch.bbclass
```diff
@@ -0,0 +1,20 @@
+# Set jailhouse architecture JH_ARCH variable
+#
+# return value must match one of architectures supported by jailhouse
+#
+valid_jh_archs = "x86 arm"
```
zandrey, line 5: Does `x86` really required in this layer? I guess it would be built for `arm` only since `COMPATIBLE_MACHINE` is set to *mx8m* only...

panfir: I'm not sure about other boards but most likely all IMX8 will get JH, so we can map on arm64 and get rid of this class, we can safely set JH_ARCH directly in the recipe

recipes-extended/jailhouse/jailhouse_0.11.bb
panfir: Okay, np, we can use `codeaurora.org` directly into `SRC_URI`

recipes-extended/jailhouse/jailhouse_0.11.bb
panfir: I could set SPDX to MIT but it's not necessary.
I will not add copyright.

panfir: Copyrights in recipes indicate who wrote the recipe, if originator didn't add a copyright on it, we should leave it without one.

recipes-extended/jailhouse/jailhouse/0001-scripts-tools-switch-to-python3.patch
panfir: I would have liked to get rid of this patch using instead that symlink in the recipe, so as not to mix the python's path:
`ln -sf ${PYTHON} ${STAGING_BINDIR_NATIVE}/python`
I had to put conditional python3 in the patch to make it work.

recipes-extended/jailhouse/jailhouse_0.11.bb
panfir: Also, it seem LICENSE should be change as well, it's not about a dual license, it's GPL-2.0

classes/jailhouse-arch.bbclass
zandrey: Cool, I guess the class should be dropped.

recipes-extended/jailhouse/jailhouse/0001-scripts-tools-switch-to-python3.patch
zandrey: IMHO, modifying source in recipe is not all that great... So symlinks would not be a solution here, I guess.

What I meant is: could you leave the conditional assignment in the patch as it was done in the upstream, something like this:

```
-PYTHON ?= python
+PYTHON ?= /usr/bin/env python3
```

Just replace the interpreter, and leave the assignment as-is. That would get less questions from the upstream guys once the patch is reported.

recipes-extended/jailhouse/jailhouse_0.11.bb
zandrey: OK, I see. I guess since you've originated this work from [this repository](https://bitbucket.org/retotech/meta-jailhouse) - maybe you can put this link in the recipe comments header?

In that case - the original author and copyright would be retained IMHO.

recipes-extended/jailhouse/jailhouse/0001-scripts-tools-switch-to-python3.patch
panfir: I know what you meant (I wrote it wrong) ..but it doesn't work, it's messing up the python path. 
Maybe I should add an assignment into the recipe

recipes-extended/jailhouse/jailhouse_0.11.bb
panfir: Indeed, it seems debatable..
I think that reference should stay in the commit message, to have it in history rather than in the header, to always have it there.
Anyway, to be copyrightable, something must contain an original expression. We just added / adapted some things to the original to work with i.MX.

classes/jailhouse-arch.bbclass
panfir: That's what I'm saying

rehsack:
> @zandrey @rehsack @fbertux please take a look.

Looks to me as if you and @zandrey already did entire work for rewiewing :)

From a strategic point of view, I am uncertain. On the one hand it is a possibility to use the i.MX8 cleverly - but there is a lack of e.g. QorIQ, which surely have the technical requirements. On the other hand, the basic support should probably be in [meta-virtualization](https://git.yoctoproject.org/cgit/cgit.cgi/meta-virtualization/) and only the i.MX8 etc. specialization here.

otavio:
> > @zandrey @rehsack @fbertux please take a look.
> 
> Looks to me as if you and @zandrey already did entire work for rewiewing :)
> 
> From a strategic point of view, I am uncertain. On the one hand it is a possibility to use the i.MX8 cleverly - but there is a lack of e.g. QorIQ, which surely have the technical requirements. On the other hand, the basic support should probably be in [meta-virtualization](https://git.yoctoproject.org/cgit/cgit.cgi/meta-virtualization/) and only the i.MX8 etc. specialization here.

Agreed and we can move common parts there. I just don't want to hold it for much longer.

I also support the idea of making this dependent of meta-virtualization in future.

PR #269, opened 2020-01-22
Python3 update for meta-freescale layer

zandrey:
Python2 support has been recently dropped from OE-Core due to it's effective EOL, therefore some recipes and corresponding source code packages should be updated to use Python3 BitBake classes.

@otavio @thochstein:
`optee-imx` and `imx-gpu-apitrace` has source code patches included in order to be able to use Python3 and should be applied at the source. Can you please forward them so they could be removed from the layer afterwards?

Thanks a lot!

-- andrey

### DIFF
--- a/recipes-extended/ovs-dpdk/ovs-dpdk_2.11.bb
+++ b/recipes-extended/ovs-dpdk/ovs-dpdk_2.11.bb
@@ -5,7 +5,7 @@ LIC_FILES_CHKSUM = "file://LICENSE;md5=1ce5d23a6429dff345518758f13aaeab"
 DEPENDS = "dpdk python-six-native coreutils-native"
 RDEPENDS_${PN} = "bash libcrypto libssl python"
 
-inherit pythonnative
+inherit python3native
 
 SRC_URI = "git://source.codeaurora.org/external/qoriq/qoriq-components/ovs-dpdk;nobranch=1 \
 "
--- a/recipes-extended/secure-obj/secure-obj.inc
+++ b/recipes-extended/secure-obj/secure-obj.inc
@@ -6,7 +6,7 @@ RDEPENDS_${PN} = "bash libcrypto libssl"
 
 DEPENDS += "python-pycrypto-native"
 
-inherit pythonnative
+inherit python3native
 
 LDFLAGS += "${TOOLCHAIN_OPTIONS}"
 
--- a/recipes-extended/vpp-core/vpp-core.bb
+++ b/recipes-extended/vpp-core/vpp-core.bb
@@ -22,7 +22,7 @@ DEPENDS_class-native = "openssl-native  python-ply-native util-linux-native"
 
 inherit cmake
 inherit pkgconfig
-inherit python-dir pythonnative
+inherit python3-dir python3native
 
 export ARCH="arm64"
 export OPENSSL_PATH = "${RECIPE_SYSROOT}"
--- a/recipes-graphics/imx-gpu-apitrace/imx-gpu-apitrace/0001-switch-from-python2-to-python3.patch
+++ b/recipes-graphics/imx-gpu-apitrace/imx-gpu-apitrace/0001-switch-from-python2-to-python3.patch
@@ -1,0 +1,10004 @@
+From bfa1b7c7e675f198f4573325f1252e91f80f0b4d Mon Sep 17 00:00:00 2001
+From: =?UTF-8?q?Piotr=20Podsiad=C5=82y?= <ppodsiadly@mykolab.com>
+Date: Thu, 3 Jan 2019 20:39:55 +0100
+Subject: [PATCH 1/4] switch from python2 to python3
+
+Upstream-Status: Backport [https://github.com/apitrace/apitrace/commit/0b8b019952f2668a5e8786cc2d331062b958f02e]
+
+(cherry picked from commit 0b8b019952f2668a5e8786cc2d331062b958f02e)
+Signed-off-by: Andrey Zhizhikin <andrey.zhizhikin@leica-geosystems.com>
+---
+ CMakeLists.txt                |    7 +-
+ cli/CMakeLists.txt            |    2 +-
+ dispatch/CMakeLists.txt       |    2 +-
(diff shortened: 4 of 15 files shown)


recipes-graphics/imx-gpu-apitrace/imx-gpu-apitrace/0002-cmake-Use-PythonInterp-instead-of-Python3.patch
```diff
@@ -0,0 +1,225 @@
+From 762587079dea1c24de95c38c271ed1692c8f2a18 Mon Sep 17 00:00:00 2001
+From: Jose Fonseca <jfonseca@vmware.com>
+Date: Sun, 14 Apr 2019 22:22:24 +0900
+Subject: [PATCH 2/7] cmake: Use PythonInterp instead of Python3.
+
+As the latter is not in widely available CMake versions.
+
+Upstream-Status: Pending
```
otavio: It seems to be a backport, isn't it?

recipes-graphics/imx-gpu-apitrace/imx-gpu-apitrace/0004-cli-Pickle-blobs-as-Python-3-bytes-objects.patch
```diff
@@ -0,0 +1,82 @@
+From 63b651332cfeb02cc723675ad2d150641d037e21 Mon Sep 17 00:00:00 2001
+From: Jose Fonseca <jfonseca@vmware.com>
+Date: Sun, 28 Apr 2019 13:47:29 +0100
+Subject: [PATCH 4/7] cli: Pickle blobs as Python 3 bytes objects.
+
+Upstream-Status: Pending
```
otavio: Here too

recipes-graphics/imx-gpu-apitrace/imx-gpu-apitrace/0002-cmake-Use-PythonInterp-instead-of-Python3.patch
zandrey: Correct! This one is ported, would drop it now.

recipes-graphics/imx-gpu-apitrace/imx-gpu-apitrace/0004-cli-Pickle-blobs-as-Python-3-bytes-objects.patch
zandrey: I've checked the latest state of `imx_7.1` branch and looks like this patch applies. It modifies generated blobs from using _bytearrays_ to python3 _byte objects_.

Original support for _bytearrays_ has been introduced with eba2deca7 from upstream, and is not changed at the HEAD of `imx_7.1`.

thochstein:
> `optee-imx` and `imx-gpu-apitrace` has source code patches included in order to be able to use Python3 and should be applied at the source. Can you please forward them so they could be removed from the layer afterwards?

Will do, thanks a lot Andrey!

zandrey:
@thochstein Thanks for taking those in!

Can you please wait with forwarding those patches, @otavio has correctly pointed out that some of patches in `imx-gpu-apitrace` were already backported and are not required.

I'm working on this now to sort them out, and would update the PR with only those that are necessary and not backported yet.

Thanks a lot!

-- andrey

recipes-graphics/imx-gpu-apitrace/imx-gpu-apitrace/0004-cli-Pickle-blobs-as-Python-3-bytes-objects.patch
zandrey: I thought about this a bit, and actually this patch is not really crucial here to have python3 support. Originally, it has been introduced on the `python3` branch in upstream together with all other modifications required to move to python3, that's what made me think this one is also necessary.

I've tried to exclude this patch from recipe and build succeeded.

@otavio @thochstein Can you please advise whether you'd prefer this patch should be included in the series or should I drop it?

Thanks a lot!

-- andrey

zandrey: I've dropped this patch as well and updated PR with new version.

zandrey:
@otavio Got the `Upstream-Status` set correctly with links to commits in the upstream. I guess now it should be OK.

-- andrey

zandrey:
@otavio Thanks for taking this in! Now the layer can be built with latest oe-core (at least the python3-related recipes).

@thochstein Can you please examine patches to `optee-os` and `imx-gpu-apitrace` and send them internally? I've committed only those that are required to resolve python3 build issues, they are ready to be forwarded.

Thank you all!

-- andrey

thochstein:
> @thochstein Can you please examine patches to `optee-os` and `imx-gpu-apitrace` and send them internally? I've committed only those that are required to resolve python3 build issues, they are ready to be forwarded.

Thanks, @zandrey. We'll work on this for our 5.4 release end of March, so we'll contribute to meta-freescale in April. We'll bump apitrace to 9.0 and add your optee-imx patch.

thochstein:
> Thanks, @zandrey. We'll work on this for our 5.4 release end of March, so we'll contribute to meta-freescale in April. We'll bump apitrace to 9.0 and add your optee-imx patch.

We are working on optee version 3.7, which already has python 3 support. We'd rather not modify our current 3.2 source.

zandrey:
> 
> We are working on optee version 3.7, which already has python 3 support. I checked your patch on our 3.7 and found it to be an exact match (except for 2 minor whitespace changes).
> 
> We'd rather not modify our current 3.2 source.

@thochstein That's OK, as long as the patch is in the layer - `master` could be built for 4.19.y kernel with optee 3.2 and @otavio is fine with keeping it here.

It was actually my task to report the patch upstream (OE rules of the game), but if upstream decides it is better not to take it - we would have it applied at the meta layer level.

Thanks a lot for the feedback!

-- andrey

PR #273, opened 2020-01-24
U boot qoriq fix and update

rehsack:
Fix u-boot-qoriq (python2 -> python3 dependency) and once having it in my finger, I take the opportunity to update it immediately.


rehsack:
@otavio - bump
There're some other QorIQ related recipes using python2 I'd like to patch sooner or later - would like to do it independently from LSDK-related patches :)

PR #282, opened 2020-02-12
u-boot-fslc: add lzop-native dependency for imx6qdlsabreauto

twoerner:
The imx6qdlsabreauto MACHINE needs lzop-native for its u-boot to build.


otavio:
I think we might add it to all machines. It is a small dependency and more and more machines will start to use it.

What do you think?

twoerner:
Okay, I'll add it to all u-boot builds.

PR #283, opened 2020-02-12
Fix i.MX 8 partition overlap

thochstein:
Can we cherry-pick to zeus as well?


otavio:
@thochstein please prepare the PR for the backport.

PR #287, opened 2020-02-13
Allow to enable Etnaviv and Gallium for i.MX8

fschrempf:
**Change 1 (cosmetic):**
We don't need to manipulate the GALLIUMDRIVERS variable. Setting the
correct PACKAGECONFIG options is enough as the main recipe handles
the rest.

**Change 2 (support Etnaviv for i.MX8)**
Currently the settings for i.MX8 are fixed to only allow to use
OSMesa for offscreen rendering. Actually etnaviv is usable on
i.MX8MM an it should therefore be possible to let Gallium be
enabled instead of unconditionally disabling it.

Therefore we introduce the USE_OSMESA_ONLY variable, that can be
set to "no" for i.MX8 if etnaviv should be used just like for i.MX6.

### DIFF
--- a/recipes-graphics/mesa/mesa_%.bbappend
+++ b/recipes-graphics/mesa/mesa_%.bbappend
@@ -16,8 +16,8 @@ python () {
 
 # Enable Etnaviv and Freedreno support
 PACKAGECONFIG_append_use-mainline-bsp = " gallium"
-GALLIUMDRIVERS_append_use-mainline-bsp_armv7a = ",etnaviv,freedreno,kmsro,vc4"
-GALLIUMDRIVERS_append_use-mainline-bsp_armv7ve = ",etnaviv,freedreno,kmsro,vc4"
+PACKAGECONFIG_append_use-mainline-bsp_armv7a = " etnaviv freedreno kmsro vc4"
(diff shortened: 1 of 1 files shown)


otavio:
This looks great! very nice and clean solution.

@zandrey and @thochstein, can you guys give this a look?

zandrey:
@otavio @fschrempf I've had a look at changes and built the multimedia image - it all looks OK.

Actually, this fixes the temporary workaround that I've committed awhile ago to solve the intermittent build issues, and it looks much cleaner than what I've proposed back then.

Can only vote up for this!

-- andrey

fschrempf:
Thanks @otavio and @zandrey for reviewing and merging!

Is there any chance to get this backported to zeus? Without this it's not possible to enable the Gallium driver because of the `PACKAGECONFIG_remove_mx8 = "gallium"`. So this is actually a fix that I require in the current stable branch.

otavio:
@fschrempf yes, please prepare a backport PR (using `git cherry-pick -x`) so we can properly backport it to Zeus branch.

fschrempf:
> @fschrempf yes, please prepare a backport PR (using `git cherry-pick -x`) so we can properly backport it to Zeus branch.

Ok, I created a backport PR against the zeus branch: https://github.com/Freescale/meta-freescale/pull/289

PR #288, opened 2020-02-14
optee-os-qoriq: fix build with python3

guochunrong:
*Use /usr/bin/env instead of the fixed path of python3 incase the interpreter couldn't be found.

Signed-off-by: Chunrong Guo <chunrong.guo@nxp.com>


otavio:
Ping.

otavio:
@guochunrong ?

otavio:
Fixed in #351 .

PR #294, opened 2020-03-03
fsl-eula-unpack.bbclass: Facilitate downstream EULA update

thochstein:

### DIFF
--- a/classes/fsl-eula-unpack.bbclass
+++ b/classes/fsl-eula-unpack.bbclass
@@ -6,7 +6,18 @@
 #
 #  SRC_URI = "${FSL_MIRROR}/firmware-imx-${PV};fsl-eula=true"
 
-LIC_FILES_CHKSUM_append = " file://${FSL_EULA_FILE};md5=6c12031a11b81db21cdfe0be88cac4b3"
+FSL_EULA_FILE_MD5SUM_LA_OPT_BASE_LICENSE_V24        = "ab61cab9599935bfe9f700405ef00f28"
+FSL_EULA_FILE_MD5SUM_LA_OPT_NXP_SOFTWARE_LICENSE_V9 = "6c12031a11b81db21cdfe0be88cac4b3"
+FSL_EULA_FILE_MD5SUMS = \
+    "${FSL_EULA_FILE_MD5SUM_LA_OPT_BASE_LICENSE_V24} \
+     ${FSL_EULA_FILE_MD5SUM_LA_OPT_NXP_SOFTWARE_LICENSE_V9}"
+
+# The checksum for the EULA in the layer
+FSL_EULA_FILE_MD5SUM ?= \
+    "${FSL_EULA_FILE_MD5SUM_LA_OPT_NXP_SOFTWARE_LICENSE_V9}"
+
+LIC_FILES_CHKSUM_LAYER ?= "file://${FSL_EULA_FILE};md5=${FSL_EULA_FILE_MD5SUM}"
+LIC_FILES_CHKSUM_append = " ${LIC_FILES_CHKSUM_LAYER}"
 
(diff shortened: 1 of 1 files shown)


classes/fsl-eula-unpack.bbclass
```diff
@@ -6,7 +6,8 @@
 #
 #  SRC_URI = "${FSL_MIRROR}/firmware-imx-${PV};fsl-eula=true"
 
-LIC_FILES_CHKSUM_append = " file://${FSL_EULA_FILE};md5=6c12031a11b81db21cdfe0be88cac4b3"
+FSL_EULA_FILE_MD5SUM ?= "6c12031a11b81db21cdfe0be88cac4b3"
+LIC_FILES_CHKSUM_append = " file://${FSL_EULA_FILE};md5=${FSL_EULA_FILE_MD5SUM}"
```
otavio: I kind of understand why you wish to do that however it'll end adding uncertainty as multiple files might be used using different terms. How a legal team is supposed to handle with it?

thochstein: Whenever we release our NPI layer, we replace your EULA with the newer one. We've been doing this by deleting your EULA and your fsl-eula-unpack.bbclass and using our updated one. (This pull request is about trying to make the existing bbclass flexible so we don't have to delete yours and copy it.)

Is there a problem with this model of replacing the EULA?

otavio: Legally, might be. We are supposed to have all licenses available for the customer legal team to review and ensure it is fine.

What happens is when we remove the license, we will never know which terms cover the old or not updated binaries.

thochstein: This EULA is mainly for the click-through. Perhaps we shouldn't be adding it to LIC_FILES_CHKSUM.

otavio: The EULA enforces legal compromises. So legal team must be aware of what they are.

thochstein: Okay, that makes sense. Can be complicated. I still don't see anything wrong with the current implementation, but perhaps there's a legal issue I don't understand. Our legal team will review our current implementation to see if there are any issues.

Correct me if I'm wrong, but this is how I see our current implementation. What happens now is each package will have two license files, one from the package itself, and one from the click-through. These can get out of sync for community builds whenever you update `meta-freescale/EULA`. They can also get out of sync for NXP builds when we replace your click-through.

otavio: As far as the patch came from NXP, we can imply it is fine. 

The ideal solution would be we could have the EULA acceptance but we had a list of the EULA files which the EULA covers, with the respective hashes, so one thing could confirm the others.

thochstein: Sorry, I don't understand your solution, and I don't know what problem you are trying to solve. Can you please start with my description of the current implementation and tell me what is wrong with it?

> What happens now is each package will have two license files, one from the package itself, and one from the click-through. These can get out of sync for community builds whenever you update meta-freescale/EULA. They can also get out of sync for NXP builds when we replace your click-through.

Ann Thornton didn't understand the purpose of your solution either:

> I don’t think there is any need to do that.  There is already the latest EULA in meta-fsl-bsp-release and the hash for it in there somewhere.  I thought it had to be accepted somewhere in the install process which covers all earlier versions, so no need to accept all of those or even have a list of them.

otavio: Considering we have a single EULA file and it does cover all earlier versions, the class should point to the very file and include this hash. This provides a single EULA.

However, my understanding, is that the list of earlier EULA files it covers needs to be somewhere and in this case the inner file, inside the package, should be checked to ensure it is on this list.

thochstein: Okay, thanks for the clarification. Working on it now.

PR #300, opened 2020-03-10
glmark2: Remove imxgpu3d PACKAGECONFIG override

JoshuaWatt:
The default PACKAGECONFIG for glmark2 builds just fine on imxgpu3d
platforms, and doing so enables support for the drm based tests.

Signed-off-by: Joshua Watt <Joshua.Watt@garmin.com>

### DIFF
--- a/dynamic-layers/openembedded-layer/recipes-benchmark/glmark2/glmark2_%.bbappend
+++ b/dynamic-layers/openembedded-layer/recipes-benchmark/glmark2/glmark2_%.bbappend
@@ -1,4 +1,0 @@
-PACKAGECONFIG_imxgpu3d = "${@bb.utils.contains('DISTRO_FEATURES', 'wayland opengl', 'wayland-gles2', \
-                                bb.utils.contains('DISTRO_FEATURES', 'x11 opengl',     'x11-gl x11-gles2', '', d), d)}"
-PACKAGECONFIG_imxgpu2d = "${@bb.utils.contains('DISTRO_FEATURES', 'wayland opengl', '', \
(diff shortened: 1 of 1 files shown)


otavio:
Did you check it against which distros?

JoshuaWatt:
It's our in house distro (based on poky). With it, the recipe has:

 PACKAGECONFIG="wayland-gl wayland-gles2 drm-gl drm-gles2"

The notable additions over what the bbappend was doing are `wayland-gl drm-gl drm-gles2`, all of which build.

We don't use X, but the bbappend and the base recipe were doing the same thing be adding `x11-gl x11-gles2` if `x11` and `opengl` were in `DISTRO_FEATURES`, which means there is no change there.

otavio:
@thochstein could you take a look?

thochstein:
I would like to take a look. It does appear on the surface that we can simplify this. However, this change did not build properly for me for a 3d i.MX 8 part. I didn't have time to debug it since we've got a release going on. Hopefully there's not a great urgency on this?

JoshuaWatt:
Not a huge urgency. We were trying to do some performance profiling on an IMX8, and found this helpful. We like to upstream our changes before backporting them locally, so I think we'd like to figure out how to make this work eventually.

JoshuaWatt:
Looks like neither of these overrides are necessary, at least AFAICT, so I removed the whole bbappend.

MaxKrummenacher:
@JoshuaWatt 
Did you test this also for i.MX 6 for which the Vivante drivers do not provide virtual/libgbm

ERROR: Nothing PROVIDES 'virtual/libgbm' (but .../glmark2_git.bb DEPENDS on or otherwise requires it)

JoshuaWatt:
No, I don't have an imx6 test platform. Thanks for fixing it.

PR #302, opened 2020-03-16
correct freescale address

roumy:
change freescale git repo address on rocko branch to make https://wiki.somlabs.com/index.php/Yocto_for_VisionSOM-6ULL/6UL consistent


otavio:
You should not bump the revision.

roumy:
hi , could you explain me more please

i get that 
> SRCREV = "f6d0859f9435796f03ae93b70b5f92f4406bc56d
can be removed and latest revision will be automatically get
but i am not the original SRCREV definition , i just correct the stuff to make it work.

otavio:
Please keep `951c1363abe95dd75ab3e9447f640d7807240236` as revision, so we don't change it.

roumy:
the thing is that original source URI "git.freescale.com/imx/imx-firmware.git"
do no more exist. 
It has be migrated to "github.com/NXP/imx-firmware.git" .
may be linked to 
https://github.com/NXP/imx-firmware/issues/1

otavio:
The `SRC_URI` is fine, the `SRCREV` should not change.

roumy:
The SRCREV  is not present in github repo. It cannot work .

otavio:
I see.

matteomartelli:
It looks like there is the same issue in the [pyro branch](https://github.com/Freescale/meta-freescale/tree/pyro/). Could you merge it there too?

otavio:
Someone need to open specific PRs for those.

PR #306, opened 2020-03-24
imx-base: Enable wic.bmap image generation for mxs

nicola-lunghi:
See similar commit fb753d427add595d9c425fab9d4720a041db1b6d:
  imx-base: Enable wic.bmap image generation as default

  Bmaptool is a generic tool for creating the block map (bmap) for a
  file and copying files using the block map. Is faster than use dd
  to flash images to SD Cards. More info about bmap here [1].

Signed-off-by: Nicola Lunghi <nick83ola@gmail.com>

### DIFF
--- a/conf/machine/include/imx-base.inc
+++ b/conf/machine/include/imx-base.inc
@@ -298,7 +298,7 @@ IMX_DEFAULT_KERNEL_use-mainline-bsp = "linux-fslc"
 PREFERRED_PROVIDER_virtual/kernel ??= "${IMX_DEFAULT_KERNEL}"
 
 SOC_DEFAULT_IMAGE_FSTYPES = "wic.bmap wic.gz"
-SOC_DEFAULT_IMAGE_FSTYPES_mxs = "uboot-mxsboot-sdcard wic.gz"
+SOC_DEFAULT_IMAGE_FSTYPES_mxs = "uboot-mxsboot-sdcard wic.bmap wic.gz"
 
 # Do not update fstab file when using wic images
 WIC_CREATE_EXTRA_ARGS ?= "--no-fstab-update"


conf/machine/include/imx-base.inc
```diff
@@ -298,7 +298,7 @@ IMX_DEFAULT_KERNEL_use-mainline-bsp = "linux-fslc"
 PREFERRED_PROVIDER_virtual/kernel ??= "${IMX_DEFAULT_KERNEL}"
 
 SOC_DEFAULT_IMAGE_FSTYPES = "wic.bmap wic.gz"
-SOC_DEFAULT_IMAGE_FSTYPES_mxs = "uboot-mxsboot-sdcard wic.gz"
+SOC_DEFAULT_IMAGE_FSTYPES_mxs = "uboot-mxsboot-sdcard wic.bmap wic.gz"
```
otavio: Please do the same change to the other default. So other SoC does the same.

nicola-lunghi: the other default is already creating a wic.bmap ...

otavio: possibly for specific machines but not defined here. See the line above of your change.

nicola-lunghi: Sorry I am not understanding what you're saying
before my changes the two lines read as
```
SOC_DEFAULT_IMAGE_FSTYPES = "wic.bmap wic.gz"
SOC_DEFAULT_IMAGE_FSTYPES_mxs = "uboot-mxsboot-sdcard wic.gz"
```
so the problem here is for mxs the wic.bmap is not generated.
for the rest of the architectures is already there

After my changes:
```
SOC_DEFAULT_IMAGE_FSTYPES = "wic.bmap wic.gz"
SOC_DEFAULT_IMAGE_FSTYPES_mxs = "uboot-mxsboot-sdcard wic.bmap wic.gz"
```
I simply added the wic.bmap to SOC_DEFAULT_IMAGE_FSTYPES_mxs

otavio: I am very sorry, I need to take a coffee ... you're right.

nicola-lunghi: Don't worry in the current situation were all on the same boat :-)

PR #311, opened 2020-03-28
fsl-eula-unpack.bbclass: Enhance EULA model

thochstein:
Each new version of the EULA is meant to be a click
through that includes any existing packages with an
older EULA. The latest EULA is also copied in the
layer.

Capture this model in the class by listing the
known EULAs and checking that any package
has a known EULA. The new model also allows
extensibility, so new EULAs can be introduced
separately.

### DIFF
--- a/classes/fsl-eula-unpack.bbclass
+++ b/classes/fsl-eula-unpack.bbclass
@@ -6,7 +6,53 @@
 #
 #  SRC_URI = "${FSL_MIRROR}/firmware-imx-${PV};fsl-eula=true"
 
-LIC_FILES_CHKSUM_append = " file://${FSL_EULA_FILE};md5=6c12031a11b81db21cdfe0be88cac4b3"
+FSL_EULA_FILE_MD5SUM_LA_OPT_BASE_LICENSE_V8          = "01ba08a2b77c3a0a9d0ab5d4d83fed64"
+FSL_EULA_FILE_MD5SUM_LA_OPT_BASE_LICENSE_V10         = "3880bb9c943b135a30fad5e8aabd3ee9"
+FSL_EULA_FILE_MD5SUM_LA_OPT_BASE_LICENSE_V11         = "a31dce3023402a821cfcdc1bb8dbdbc0"
+FSL_EULA_FILE_MD5SUM_LA_OPT_BASE_LICENSE_V12         = "8cf95184c220e247b9917e7244124c5a"
+FSL_EULA_FILE_MD5SUM_LA_OPT_BASE_LICENSE_V13         = "102094e9e695888c6a9f9fda9004165d"
+FSL_EULA_FILE_MD5SUM_LA_OPT_BASE_LICENSE_V14         = "d4f548f93b5fe0ee2bc86758c344412d"
+FSL_EULA_FILE_MD5SUM_LA_OPT_BASE_LICENSE_V16         = "be67a88e9e6c841043b005ad7bcf8309"
+FSL_EULA_FILE_MD5SUM_LA_OPT_BASE_LICENSE_V17         = "88595f92466edd79be7f4f9547ffa0ef"
+FSL_EULA_FILE_MD5SUM_LA_OPT_BASE_LICENSE_V18         = "6b552f505eedab4a11ab538cf3db743a"
+FSL_EULA_FILE_MD5SUM_LA_OPT_BASE_LICENSE_V20         = "08fd295cce89b0a9c74b9b83ed74f671"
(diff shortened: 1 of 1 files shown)


classes/fsl-eula-unpack.bbclass
```diff
@@ -6,7 +6,29 @@
 #
 #  SRC_URI = "${FSL_MIRROR}/firmware-imx-${PV};fsl-eula=true"
 
-LIC_FILES_CHKSUM_append = " file://${FSL_EULA_FILE};md5=6c12031a11b81db21cdfe0be88cac4b3"
+FSL_EULA_FILE_MD5SUM_LA_OPT_BASE_LICENSE_V24         = "ab61cab9599935bfe9f700405ef00f28"
+FSL_EULA_FILE_MD5SUM_LA_OPT_BASE_LICENSE_V26         = "5ab1a30d0cd181e3408077727ea5a2db"
+FSL_EULA_FILE_MD5SUM_LA_OPT_NXP_SOFTWARE_LICENSE_V3  = "6dfb32a488e5fd6bae52fbf6c7ebb086"
+FSL_EULA_FILE_MD5SUM_LA_OPT_NXP_SOFTWARE_LICENSE_V5  = "80c0478f4339af024519b3723023fe28"
+FSL_EULA_FILE_MD5SUM_LA_OPT_NXP_SOFTWARE_LICENSE_V7  = "fb0303e4ee8b0e71c094171e2272bd44"
```
thochstein: I added some more license versions this time. The problem reported on zeus was specifically due to missing v7.

classes/fsl-eula-unpack.bbclass
```diff
@@ -34,9 +57,39 @@ python fsl_bin_do_unpack() {
...
+    if found > 1:
+        bb.warn("The package contains multiple Freescale EULA-licensed archives. The consistency logic may not be able to detect a EULA problem.")
+    layer_license = d.getVar('LIC_FILES_CHKSUM_LAYER')
+    licenses = d.getVar('LIC_FILES_CHKSUM') or ""
+    md5sums = d.getVar('FSL_EULA_FILE_MD5SUMS') or ""
+    found_layer_license = False
+    found_package_license = False
+    for license in licenses.split():
+        try:
+            (method, host, path, user, pswd, parm) = bb.fetch.decodeurl(license)
```
thochstein: I changed the code order so now the layer license is also subjected to this call to decodeurl.

classes/fsl-eula-unpack.bbclass
thochstein: I will be adding some more older licenses today.

thochstein:
Ping

thochstein:
Otavio mentioned [here](https://github.com/Freescale/fsl-community-bsp-base/pull/14#issuecomment-611563662):
> I am very worried about it as we caused regressions.

As I mentioned above in the comments, the regression was due to a simple reason: I hadn't included enough EULA versions. It was a false-positive. It was fixed simply by adding the missing license. I've now included as many license versions as I could find.

As a reminder, this PR was originally a simple request to make the existing code extensible. Would you prefer to go back to that and skip the additional license checking feature? We could also just treat problems as a warning for now, to be turned into an error later when we have confidence.

otavio:
No. The failure is due a more serious issue, the testing. You did test it only using NXP BSP but not pristine FSLC one.

So please conduct a testing using it before we consider merging it.

thochstein:
Yes, sorry for my testing failure.

Building `fsl-image-machine-test` now.

```
MACHINE              = "imx6qdlsabresd"
DISTRO               = "fslc-xwayland"
```

otavio:
You need to try that **without** all NXP BSP one the workspace. It must be tested on a FSLC clean environment.

thochstein:
Yes, I am always doing that for upstreaming work.

thochstein:
Is fsl-image-machine-test currently building on master? Seems perf is broken.

thochstein:
So perf compiles fine with the patch but then fails on the install.

thochstein:
Switched to `fsl-xwayland` and `imx8mqevk`, but gstreamer is broken there:
```
| ../../git/libs/overlaycompositionmeta/imxoverlaycompositionmeta.h:24:10: fatal error: gstimxcommon.h: No such file or directory
|    24 | #include "gstimxcommon.h"
|       |          ^~~~~~~~~~~~~~~~
```

thochstein:
Finally got a combination with no existing problems, `fslc-xwayland` and `imx8mqevk`. Can we get this PR merged?

PR #313, opened 2020-03-30
u-boot: Split out u-boot tools

JoshuaWatt:
Splits out the IMX u-boot tools build to it's own recipe. This is
required because these are native tools and thus need to be in a -native
recipe to be properly relocated (instead of being built as part of the
target u-boot-imx recipe). If the tools aren't built in a -native
recipe, then the mkimage tool will contain a hard coded path to the
uninative program loader from the sandbox where it was first built and
this will not get corrected when restored from sstate. The result is
that the mkimage_uboot executable can't run because it's looking for a
hard coded program loader that does not exist, resulting in strange
build errors like:

 ./mkimage_uboot: Command not found

even though the executable is obviously in the correct location.

Signed-off-by: Joshua Watt <Joshua.Watt@garmin.com>

### DIFF
--- a/conf/machine/include/imx-base.inc
+++ b/conf/machine/include/imx-base.inc
@@ -10,6 +10,10 @@ IMX_DEFAULT_BOOTLOADER = "u-boot-fslc"
 IMX_DEFAULT_BOOTLOADER_mx8 = "u-boot-imx"
 
 PREFERRED_PROVIDER_u-boot ??= "${IMX_DEFAULT_BOOTLOADER}"
+PREFERRED_PROVIDER_u-boot-tools-native ??= "${IMX_DEFAULT_BOOTLOADER}-tools-native"
+PREDERRED_PROVIDED_nativesdk-u-boot-tools ??= "${IMX_DEFAULT_BOOTLOADER}-tools-native"
+PREFERRED_PROVIDER_u-boot-mkimage-native ??= "${IMX_DEFAULT_BOOTLOADER}-tools-native"
+PREFERRED_PROVIDER_nativesdk-u-boot-mkimage ??= "${IMX_DEFAULT_BOOTLOADER}-tools-native"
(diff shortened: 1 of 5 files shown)


otavio:
@JoshuaWatt thanks for the patch. It does clean up it a lot and easy the proper use for it.

conf/machine/include/imx-base.inc
```diff
@@ -10,6 +10,10 @@ IMX_DEFAULT_BOOTLOADER = "u-boot-fslc"
 IMX_DEFAULT_BOOTLOADER_mx8 = "u-boot-imx"
 
 PREFERRED_PROVIDER_u-boot ??= "${IMX_DEFAULT_BOOTLOADER}"
+PREFERRED_PROVIDER_u-boot-tools-native ??= "${IMX_DEFAULT_BOOTLOADER}-tools-native"
+PREDERRED_PROVIDED_nativesdk-u-boot-tools ??= "${IMX_DEFAULT_BOOTLOADER}-tools-native"
+PREFERRED_PROVIDER_u-boot-mkimage-native ??= "${IMX_DEFAULT_BOOTLOADER}-tools-native"
+PREFERRED_PROVIDER_nativesdk-u-boot-mkimage ??= "${IMX_DEFAULT_BOOTLOADER}-tools-native"
```
kraj: this line should be `PREFERRED_PROVIDER_nativesdk-u-boot-mkimage ??= "nativesdk-${IMX_DEFAULT_BOOTLOADER}-tools"`

otavio: @kraj mind to open a PR for it?

JoshuaWatt: Oops. Sorry about that!

JoshuaWatt: https://github.com/Freescale/meta-freescale/pull/314

PR #322, opened 2020-04-03
u-boot-imx: resolve build issue with missing target

zandrey:
Since the replacement of u-boot-fw-utils with libubootenv in OE-Core,
U-Boot recipe requires an additional target to be built, which contains
the default environment.

This target has been introduced in U-Boot with version 2019.07 and is
missing in NXP fork of U-Boot source tree, therefore one patch from the
U-Boot source tree is backported in order to introduce a missing build
target.


otavio:
@thochstein this patch must be applied on internal tree for next release.

thochstein:
Done. Thank you @zandrey and @otavio .

PR #325, opened 2020-04-07
LSDK 20.04 update II

rehsack:
Some more LSDK-20.04 updates 
- VPP seems in progress and not prepared for LSDK-20.04
- dpdk/dpdkvpp should probably be merged (later)


otavio:
Do you see a reason to keep multiple linux recipes? I'd consider keeping 5.4 only.

rehsack:
@otavio - Yes, I think there're people who want to stay on last LTS for a while ... (I know at least one company with a conservative upgrade strategy ^^)

otavio:
But 4.14 can die, in this case.

rehsack:
I updated it to 19.09-update-311219, because NXP did so.
I can add a note to recipe that it won't survive until next LSDK-update and people who use it should upgrade soon. But I would kill it after dunfell release, not before. So people have a little chance :)

otavio:
The biggest problem is that dunfell is lts so we'd need to keep 4.19 forever.

rehsack:
Good point. Then I do a remove PR later this week (after I rebased our BSP to 5.4^^), if it's ok.
People can checkout the last state then if they want.

otavio:
Deal. Merging it now.

PR #342, opened 2020-04-15
mx8qxp: Use cortexa35-crypto tuning for i.MX 8 QXP default

thochstein:
The i.MX 8QXP has Cortex-A35 cores plus Crypto
extensions, so enable them by default.

### DIFF
--- a/conf/machine/imx8qxpmek.conf
+++ b/conf/machine/imx8qxpmek.conf
@@ -6,7 +6,7 @@
 MACHINEOVERRIDES =. "mx8:mx8x:mx8qxp:"
 
 require conf/machine/include/imx-base.inc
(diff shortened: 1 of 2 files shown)


MaxKrummenacher:
Hi Tom @thochstein 

Do you know of any benchmarks done which compare the use of a generic aarch64 tuning with this core specific settings.

Moving now all CPU's to their specific tuning has considerable build time penality, I wonder what we get for this penalty.

Max

P.S. I have nothing against these changes, I just wonder if such numbers exist.

thochstein:
Hi @MaxKrummenacher, no, we don't have these numbers.

MaxKrummenacher:
Ok, thanks.
I hope I get to it and get our build servers busy.

Max

PR #345, opened 2020-04-16
[zeus] backport tunings for 8QM & 8QXP

thochstein:


thochstein:
With different default tunings, is it still worth it to have the same `MACHINE_SOCARCH_SUFFIX` for mx8qm and mx8qxp?

otavio:
For now yes.

PR #363, opened 2020-04-22
optee-test_3.2.0.imx: fix TA_DEV_KIT_DIR is not correctly defined error

petegriffin:
DEFAULTTUNE is no longer aarch64 which means optee-test
build fails. Update to use the same mechanism as optee-os
recipe for setting arch.

Build now passes on imx8mq-evk and optee xtest suite
passes

+-----------------------------------------------------
16078 subtests of which 3 failed
74 test cases of which 3 failed
0 test case was skipped
TEE test application done!

Signed-off-by: Peter Griffin <peter.griffin@linaro.org>

### DIFF
--- a/recipes-security/optee-imx/optee-test_3.2.0.imx.bb
+++ b/recipes-security/optee-imx/optee-test_3.2.0.imx.bb
@@ -26,14 +26,14 @@ S = "${WORKDIR}/git"
 
 SRCREV = "b7114b828b82f2c0eec124ed424eff1230cc5319"
 
+OPTEE_ARCH ?= "arm32"
+OPTEE_ARCH_armv7a = "arm32"
+OPTEE_ARCH_aarch64 = "arm64"
+
 do_compile () {
-    if [ ${DEFAULTTUNE} = "aarch64" ];then
-        export TA_DEV_KIT_DIR=${STAGING_INCDIR}/optee/export-user_ta_arm64/
-        export ARCH=arm64
-    else
-        export TA_DEV_KIT_DIR=${STAGING_INCDIR}/optee/export-user_ta_arm32/
-        export ARCH=arm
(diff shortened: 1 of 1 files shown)


recipes-security/optee-imx/optee-test_3.2.0.imx.bb
```diff
@@ -26,14 +26,14 @@ S = "${WORKDIR}/git"
...
 do_compile () {
-    if [ ${DEFAULTTUNE} = "aarch64" ];then
-        export TA_DEV_KIT_DIR=${STAGING_INCDIR}/optee/export-user_ta_arm64/
-        export ARCH=arm64
-    else
-        export TA_DEV_KIT_DIR=${STAGING_INCDIR}/optee/export-user_ta_arm32/
-        export ARCH=arm
-    fi
+    export TA_DEV_KIT_DIR=${STAGING_INCDIR}/optee/export-user_ta_${OPTEE_ARCH}/
+    export ARCH=${OPTEE_ARCH}
```
moorthy-bs, line 35: ARCH is previously set with 'arm' but here it sets as 'arm32'. Does it correct?

petegriffin: @moorthy-bs  Good point, I should of mentioned that in the commit message. I believe that was a bug as optee-os recipe installs TA_DEV_KIT_DIR into OPTEE_ARCH which is set to arm32 (see https://github.com/Freescale/meta-freescale/blob/master/recipes-security/optee-imx/optee-os_3.2.0.imx.bb#L83). Although I've never tested the arm32 side of things

@otavio I think as you requested in previous pull, this is now tested with 5.4 kernel and all the latest meta-freescale master changes to integrate latest NXP BSP components

PR #374, opened 2020-04-29
Weston cleanup and fixes

thochstein:


recipes-graphics/wayland/weston_8.0.0.imx.bb
```diff
@@ -1,9 +1,6 @@
+require recipes-graphics/wayland/weston_8.0.0.bb
+
```
otavio, line 2: We cannot link the files; you need to copy the patches and other files that are included on `SRC_URI`. Linking enforces a specific layout for the platform and we should never enforce that.

One pattern you could use is add modified or new patches/files under an `imx-specific` subdir and use this as part of `SRC_URI`.

recipes-graphics/wayland/weston_8.0.0.imx.bb
```diff
@@ -29,109 +22,21 @@ PACKAGECONFIG_OPENGL              = "opengl"
...
-
-EXTRA_OEMESON += "-Dbackend-default=auto -Dbackend-rdp=false -Dpipewire=false"
-
-PACKAGECONFIG ??= "${@bb.utils.contains('DISTRO_FEATURES', 'wayland', 'kms fbdev wayland egl clients', '', d)} \
-                   ${@bb.utils.contains('DISTRO_FEATURES', 'x11 wayland', 'xwayland', '', d)} \
-                   ${@bb.utils.filter('DISTRO_FEATURES', 'pam systemd x11', d)} \
-                   ${@bb.utils.filter('DISTRO_FEATURES', '${PACKAGECONFIG_OPENGL}', d)} \
-                   ${@bb.utils.contains_any('DISTRO_FEATURES', 'wayland x11', '', 'headless', d)} \
-                   launch"
+PACKAGECONFIG_append = " ${@bb.utils.filter('DISTRO_FEATURES', '${PACKAGECONFIG_OPENGL}', d)}"
```
otavio, line 25: This seems as not causing any effect as the OPENGL specific ones are already included on the default.

recipes-graphics/wayland/weston_8.0.0.imx.bb
```diff
@@ -29,109 +22,21 @@ PACKAGECONFIG_OPENGL              = "opengl"
...
-                chmod u+s ${D}${bindir}/weston-launch
-        fi
-}
-
-PACKAGES += "${@bb.utils.contains('PACKAGECONFIG', 'xwayland', '${PN}-xwayland', '', d)} \
-             libweston-${WESTON_MAJOR_VERSION} ${PN}-examples"
-
-FILES_${PN}-dev += "${libdir}/${BPN}/libexec_weston.so"
-FILES_${PN} = "${bindir}/weston ${bindir}/weston-terminal ${bindir}/weston-info ${bindir}/weston-launch ${bindir}/wcap-decode ${libexecdir} ${libdir}/${BPN}/*.so* ${datadir}"
 FILES_${PN}_append = " ${sysconfdir}/xdg/weston"
```
otavio: This is wrong; the idea behind `weston-init` is to allow it to be customized without recompiling the `weston`. Please add the settings and specifics there.

recipes-graphics/wayland/weston_8.0.0.imx.bb
thochstein: ? This is not a change. The previous file already had this.

recipes-graphics/wayland/weston_8.0.0.imx.bb
thochstein: This is also not a change. The last commit in the set does address the issue you note.

recipes-graphics/wayland/weston_8.0.0.imx.bb
otavio: Sure, what I am saying is it is pointless as the OPENGL is already on the default `PACKAGECONFIG` value.

recipes-graphics/wayland/weston_8.0.0.imx.bb
otavio: Good, when I reviewed I missed this last commit.

recipes-graphics/wayland/weston_8.0.0.imx.bb
thochstein: It is not on by default:
```
PACKAGECONFIG_OPENGL              = "opengl"
PACKAGECONFIG_OPENGL_imxgpu2d     = ""
PACKAGECONFIG_OPENGL_imxgpu3d     = "opengl"
```

otavio: Sorry, you're right.

zandrey:
@thochstein I don't think this is a proper way to construct a derivative recipe. If there is a need to amend to oe-core recipe - `.bbappend` should be used.

thochstein:
We used `.bbappend` in the past, but @otavio wanted the fork recipe to be kept separate.

recipes-graphics/wayland/weston_8.0.0.imx.bb
thochstein: Is it common to move layers to custom locations? At least with a linked patch file, the use of a custom location would generate an immediate build error with a simple fix. With a copied patch file, there is no simple way to prevent our patch files from going stale.

zandrey:
> 
> 
> We used `.bbappend` in the past, but @otavio wanted the fork recipe to be kept separate.

In this case I would rather go with the complete copy of the recipe, since it would:
a) help us to see the delta between oe-core and fork;
b) do not depend on the recipe from upstream, because once upgraded - it would break the build in this layer

I tend not to include complete recipes into others, as it breaks the architectural approach of Yocto itself, and creates an additional maintenance effort where the included `.bb` file should be tracked not to disappear due to upgrade.

Bottom line is - the final call is on @otavio since he maintains this layer, but I would strongly advise against such an inclusion.

-- andrey

recipes-graphics/wayland/weston_8.0.0.imx.bb
zandrey: > 
> 
> Is it common to move layers to custom locations?

No, but recipe get upgraded via rename. When `require` is used - this would effectively break parsing stage once upstream gets upgraded (via `git mv ...`)

> At least with a linked patch file, the use of a custom location would generate an immediate build error with a simple fix. 

Immediate build failure - yes, but I'm not sure about an easy fix. 

> With a copied patch file, there is no simple way to prevent our patch files from going stale.

That I guess that is the whole intention of the fork - we get to keep it to ourselves, with all the goodies and baddies. Patch would never go stale on the fork, as it does not advance together with upstream. But even when it does - then it is the responsibility of the person upgrading the fork to upgrade patches.

Besides, who prevents from taking those patches submitted in oe-core layer and directly applying them to the fork?

thochstein:
> a) help us to see the delta between oe-core and fork;

I much prefer to see exactly what we changed. And I find it more difficult looking at the delta when you have to wonder whether a change is due to something we did or something that happened upstream. You can use this same argument to say that you shouldn't use `.bbappend`, no?

> b) do not depend on the recipe from upstream, because once upgraded - it would break the build in this layer

This is handled simply by copying the last good `.bb` into the layer. No big deal, and no different than a `.bbappend` that no longer has a `.bb`.

> 
> I tend not to include complete recipes into others, as it breaks the architectural approach of Yocto itself, and creates an additional maintenance effort where the included `.bb` file should be tracked not to disappear due to upgrade.

I know this pattern is unique, but I see no problem with it, and in my opinion helps us to get the benefits of re-use without the drawbacks of re-use through copying. If you can't have `.bbappend`, then this is the next best thing.

recipes-graphics/wayland/weston_8.0.0.imx.bb
thochstein: > > Is it common to move layers to custom locations?
> 
> No, but recipe get upgraded via rename. When `require` is used - this would effectively break parsing stage once upstream gets upgraded (via `git mv ...`)

Actually, the `require` is the one part that would just work because the require is a relative path.

> 
> > At least with a linked patch file, the use of a custom location would generate an immediate build error with a simple fix.
> 
> Immediate build failure - yes, but I'm not sure about an easy fix.

Okay, maybe not exactly easy. I was thinking about the user at least knowing about his custom layout, but finding the broken link might take some luck/insight on the user's part. But if the custom layout is uncommon, than to me it's still the lesser evil compared to not picking up some fix or feature from a modified upstream patch.

> 
> > With a copied patch file, there is no simple way to prevent our patch files from going stale.
> 
> That I guess that is the whole intention of the fork - we get to keep it to ourselves, with all the goodies and baddies. Patch would never go stale on the fork, as it does not advance together with upstream. But even when it does - then it is the responsibility of the person upgrading the fork to upgrade patches.
> 
> Besides, who prevents from taking those patches submitted in oe-core layer and directly applying them to the fork?

Sure, there's nothing preventing someone from monitoring the upstream changes and updating our patches. The problem is what do you do to make sure that it gets done in a timely fashion? With my proposal there is zero effort required.

zandrey:
> I much prefer to see exactly what we changed.

This is what the `.bbappend` is exactly for (well, one of its implications).

> And I find it more difficult looking at the delta when you have to wonder whether a change is due to something we did or something that happened upstream.

This is truly painful, I agree. But this was always the case with forks: if you tend to keep your changes, then you have to merge them yourself.

> You can use this same argument to say that you shouldn't use .bbappend, no?

Correct. 😉 

> This is handled simply by copying the last good .bb into the layer. No big deal, and no different than a .bbappend that no longer has a .bb.

That's the catch - I haven't seen that there is a copy of the `.bb`, it is rather a (relatively strong) `require` statement which refers to the recipe upstream.

May I suggest what to my opinion can be considered as a compromise:

Let's have a copy of the upstream recipe in this layer, and identify block(s) that are modified for NXP fork with some markers, something of this kind:
```
########### imx-specific section ###############
[snip]
########### imx-specific section ###############
```

By this we would keep the original recipe and imx changes in one file, which we can always dircetly  compare against upstream version. Since there are not a whole lots of changes in *imx* part of the recipe, it should be pretty feasible to do. There is also one big advantage here - we get to read the same file to understand for example `PACKAGECONFIG` keys, and would not have to jump between *oe-core* and *imx* `.bb` files.

How does that sounds?

thochstein:
> Let's have a copy of the upstream recipe in this layer, and identify block(s) that are modified for NXP fork with some markers, something of this kind:

I appreciate your compromise, but I feel that that's only marginally better than what we currently have, and misses out on the main benefits of including the recipe. I guess you are thinking of this fork recipe as somehow different than a `.bbappend`. I feel it is the same, just that we don't have the benefit of using the same mechanism because @otavio wants our recipe separated by version so we don't pollute the upstream recipe.

I'm personally fine with using a `.bbappend`, and think we can prevent `pollution` with overrides. @otavio, is there another reason I'm missing for not using the `.bbappend`? Perhaps you would prefer that we lock down this recipe more like what @zandrey seems to be thinking?

zandrey:
> I guess you are thinking of this fork recipe as somehow different than a .bbappend.

I actually do not - this PR represents the *alternative* solution to `.bbappend`.

> @otavio wants our recipe separated by version so we don't pollute the upstream recipe.

That is the key point, that made me torn by a logical problem, namely:
Implementation in this PR suggests that it should be taken care by `.bbappend` , but since it is a fork which provides a package with different name (`weston-imx`), there has to be a separate recipe that covers it. But if the recipe is introduced - it would look like the one from upstream.

With using the `.bbappend` we would end up with the package named `weston` (like in oe-core), but the content would be provided from the fork which would break reproducibility.

On the other hand, *copying* (not *including*) the upstream recipe, renaming it to `weston-imx`, and keeping in this layer increases the maintenance effort (upstream against fork modifications).

All-in-all, for me seeing the `.bb` included in another `.bb` is not really a good solution, IMHO.

thochstein:
> That is the key point, that made me torn by a logical problem, namely:
> Implementation in this PR suggests that it should be taken care by `.bbappend` , but since it is a fork which provides a package with different name (`weston-imx`), there has to be a separate recipe that covers it. But if the recipe is introduced - it would look like the one from upstream.
> 
> With using the `.bbappend` we would end up with the package named `weston` (like in oe-core), but the content would be provided from the fork which would break reproducibility.

Not sure I understand. In this case, if we were to use `.bbappend`, then we would want to limit the fork by override `SRC_URI_imx`.

otavio:
The use of `.bbappend` for forked repositories is a no go. The changes here hidden and seems as it were the original source. Using inline recipe or inclusion of the recipe we could discuss.

Including the original recipe (using `require`) is similar to adding verbatim inside of the recipe. The positive gains are:

* we gain bugfixes on the recipe of patches are merged on OE-Core
* we have smaller recipes which clearly show what is added

On the other hand, once there is a version mismatch this might become cumbersome. In this case, adding the recipe verbatim allows for easier maintenance. Doing so, @zandrey suggestion of adding a section on the recipe with the changes is clearly a win.

If we opt to this route, I'd like to extend @zandrey proposal of the header with:

```
########### imx-specific section ###############
# Upstream hash: <git-hash>

<code changes>

########### imx-specific section ###############
```

Using the hash allows us to compare with OE-Core when doing updates and bugfixes.

What do you guys think?

thochstein:
I'm fine with that. An alternative to copying the recipe in the file would be to copy the recipe to an include file, say `weston-8.0.0-xxxxxxx.inc`, thus keeping the basic file structure consistent (and even making the fact of the copy visible at the file level).

otavio:
Personally, I am not a big fan of `.inc` files where they are included once. I'd prefer it verbatim on the recipe as the use of the header, with the changes, highlight the differences from the original recipe and avoids one level of indirection.

zandrey:
> 
> What do you guys think?

This is the strategy I meant. We can compare the original recipe to the one in the fork when we would have those *section anchors* installed. And in this case we would provide a package different from the one in oe-core (not the SRC_URI only 😉).

`.inc` is conceptually the same as having a copy of recipe - as long as *imx* sections are defined, it is still possible to compare upstream with the fork.

There is one thing from my side to note regarding this:
> we gain bugfixes on the recipe of patches are merged on OE-Core

What if the upstream patch would not apply clean to the fork? If we hold the copy - this could easily be identified since it should be merged by hand, while with inclusion we might have some interesting build/runtime issues. 

-- andrey

thochstein:
> This is the strategy I meant.

To be clear, I understood Otavio's proposal to mean that we would use `require` until the upstream recipe is updated. Did I understand correctly?

otavio:
> To be clear, I understood Otavio's proposal to mean that we would use `require` unless there is a mismatch with the upstream recipe. Did I understand correctly?

No; looking at @zandrey I think his arguments are good and we should add the recipe verbatim /but/ move the changes down the recipe with the header, hash and code block.

thochstein:
Okay, so bottom line is you want the recipe logic locked. Fixes and features from upstream (stable branch) will not be available immediately, but will instead wait until someone notices them and manually moves them.

thochstein:
One drawback of this approach is someone opening the recipe will see a standard weston recipe and not know that the bottom part switches it to a fork. I guess we need some more structure:

```
# This recipe is for the i.MX fork of weston. For ease of
# maintenance, the top section is a verbatim copy of an OE-core
# recipe. The second section customizes the recipe for i.MX.

########### OE-core copy ##################
# Upstream hash: 9b1d30810eeecb46b977c8eed68be69aef891312

<copy>

########### End of OE-core copy ###########

########### i.MX overrides ################

<overrides>

########### End of i.MX overrides #########
```

zandrey:
> Look okay?

Yes, LGTM

Actually, when I did upgrade of the `gstreamer` recipes - I would've been dreaming to have something of this kind, as it was incredibly hard to match update from upstream with changes from the fork.

> Okay, so bottom line is you want the recipe logic locked. 

Correct.

> Fixes and features from upstream (stable branch) will not be available immediately, but will instead wait until someone notices them and manually moves them.

Unfortunately, yes.
Fortunately, this would create a clear separation between fork and upstream.
😉

Issue #378, opened 2020-04-30
bug(dunfell): MACHINEOVERRIDES change breaks qtbase (and probably more)


cpboyd:
I think this might be related to https://github.com/Freescale/meta-freescale/commit/341ba18eccc7703eb5ac23dc900b3bf087d3be20

Basically, I've noticed that in my configuration, it adds `-opengl desktop` rather than `-opengl egl` which throws a configuration error.

cpboyd:
After further inspection, the real issue lies with changes to `MACHINEOVERRIDES` https://github.com/Freescale/meta-freescale/blob/dunfell/conf/machine/include/imx-base.inc#L23

Possibly because of the old line related to the `use-mainline-bsp`: https://github.com/Freescale/meta-freescale/blob/dunfell/conf/machine/include/imx-base.inc#L109

cpboyd:
Probably related to #86 

May make more sense to modify `MACHINEOVERRIDES_EXTENDER_FILTER_OUT` than to revert the base `MACHINEOVERRIDES` addition of `use-mainline-bsp`, but I'm not familiar with how the `machine-overrides-extender` works

otavio:
It does not seem to be the case. I did:

```
% MACHINE=nitrogen6x ye x qtbase PACKAGECONFIG
Loading cache: 100% |#########################################################################################################| Time: 0:00:00
Loaded 3344 entries from dependency cache.
=== Final value
PACKAGECONFIG =      release     accessibility dbus udev evdev widgets tools libs freetype tests pcre                openssl     gles2                         jpeg libpng zlib            eglfs

=== Expansion
PACKAGECONFIG ==>      ${PACKAGECONFIG_RELEASE}     ${PACKAGECONFIG_DEFAULT}     ${PACKAGECONFIG_OPENSSL}     ${PACKAGECONFIG_GL}     ${PACKAGECONFIG_FB}     ${PACKAGECONFIG_X11}     ${PACKAGECONFIG_KDE}     ${PACKAGECONFIG_FONTS}     ${PACKAGECONFIG_SYSTEM}     ${PACKAGECONFIG_DISTRO}  ${PACKAGECONFIG_PLATFORM}
    PACKAGECONFIG_RELEASE ==> release
[...]

PR #379, opened 2020-05-01
Revert "qtbase: Do not pin gles2 when using open source graphic stack"

otavio:
This reverts commit 1cf9efd20fd94afdd2f7484b0570c570a21b8b65.

The qtbase configure fail to find any suitable GL library, so it fails
with:

,----
| WARNING: No QPA platform plugin enabled! This will produce a Qt that
| cannot run GUI applications.  See "Platform backends" in the output of
| --help.
|
| ERROR: Feature 'opengl-desktop' was enabled, but the pre-condition
| '(config.win32 && !config.winrt && !features.opengles2
|     && (config.msvc || libs.opengl))
|  || (!config.watchos && !config.win32 && !config.wasm && libs.opengl)' failed.
|
| ERROR: Feature 'eglfs' was enabled, but the pre-condition
|   '!config.android && !config.darwin && !config.win32 && !config.wasm
|   && features.egl' failed.
`----

Fixed: #378

### DIFF
--- a/dynamic-layers/qt5-layer/recipes-qt/qt5/qtbase_%.bbappend
+++ b/dynamic-layers/qt5-layer/recipes-qt/qt5/qtbase_%.bbappend
@@ -19,7 +19,7 @@ SRC_URI_append_imxgpu3d = " \
 PACKAGECONFIG_GL_imxpxp   = "gles2"
 PACKAGECONFIG_GL_imxgpu2d = "${@bb.utils.contains('DISTRO_FEATURES', 'x11', ' gl', '', d)}"
 PACKAGECONFIG_GL_imxgpu3d = "gles2"
-PACKAGECONFIG_GL_use-mainline-bsp = "gles2 gbm kms"
+PACKAGECONFIG_GL_use-mainline-bsp ?= "gles2 gbm kms"
 
 PACKAGECONFIG_PLATFORM          = ""
 PACKAGECONFIG_PLATFORM_imxgpu2d = "no-opengl linuxfb"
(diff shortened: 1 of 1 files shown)


otavio:
@schnitzeltony please take a look. @cpboyd please test.

schnitzeltony:
OK - we need a default. Can we do something like

PACKAGECONFIG_GL_use-mainline-bsp ?= "gles2 gbm kms"

to make it easier to override? I ask this because there are few recipes in meta-qt5-extra that do build with full gl only (and full gl works fine and similar perfomance as gles)

cpboyd:
@otavio Everything seems to work.

@schnitzeltony The underlying issue is that by default, `gl` corresponds to the flag `-opengl desktop` which doesn't work with Framebuffer. Maybe using some sort of `${@bb.utils.contains('DISTRO_FEATURES', 'x11', ' gl', 'gles2', d)}`

otavio:
> OK - we need a default. Can we do something like
> 
> PACKAGECONFIG_GL_use-mainline-bsp ?= "gles2 gbm kms"
> 
> to make it easier to override? I ask this because there are few recipes in meta-qt5-extra that do build with full gl only (and full gl works fine and similar perfomance as gles)

I'm wondering why it failed to build. Did you get it building with `opengl` and `eglfs`?

schnitzeltony:
> I'm wondering why it failed to build. Did you get it building with opengl and eglfs?

You lost me a bit.

1. Never used eglfs - why are you asking for it?
2. There are packages made by people knowing nothing but PC and assuming qtbase build for OpenGL. These fail to build with qtbase build for gles.

Therefore I ask to not hard pin gles2 by using '?=' instead of '='. to give (distro-)layers the chance to override. Nothing more - no magic failing on corners.
And by the way: It is not upon a BSP-Layer to decide what gl to use particularly not when using open source mesa which frees us from all the nasty either-or decisions BLOBs put upon us.

otavio:
> > I'm wondering why it failed to build. Did you get it building with opengl and eglfs?
> 
> You lost me a bit.
> 
>     1. Never used eglfs - why are you asking for it?

Good but did you built it without Wayland and X11? So, framebuffer only.

>     2. There are packages made by people knowing nothing but PC and assuming qtbase build for OpenGL. These fail to build with qtbase build for gles.
> 
> 
> Therefore I ask to not hard pin gles2 by using '?=' instead of '='. to give (distro-)layers the chance to override. Nothing more - no magic failing on corners.
> And by the way: It is not upon a BSP-Layer to decide what gl to use particularly not when using open source mesa which frees us from all the nasty either-or decisions BLOBs put upon us.

Sure but my question was if you have any idea why the `opengl` is not building at all.

PR #380, opened 2020-05-01
mainline BSP small improvements

otavio:
@agners @texierp @thochstein @zandrey @fbertux 

Please take a look.

### DIFF
--- a/recipes-graphics/mesa/mesa_%.bbappend
+++ b/recipes-graphics/mesa/mesa_%.bbappend
@@ -15,7 +15,7 @@ python () {
 }
 
 # Enable Etnaviv and Freedreno support
-PACKAGECONFIG_append_use-mainline-bsp = " gallium etnaviv kmsro vc4 freedreno"
+PACKAGECONFIG_append_use-mainline-bsp = " gallium etnaviv kmsro freedreno"
 
 USE_OSMESA_ONLY ?= "no"
 
--- a/recipes-graphics/mesa/mesa_%.bbappend
+++ b/recipes-graphics/mesa/mesa_%.bbappend
@@ -15,10 +15,7 @@ python () {
 }
 
 # Enable Etnaviv and Freedreno support
-PACKAGECONFIG_append_use-mainline-bsp = " gallium"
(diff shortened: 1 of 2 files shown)


recipes-graphics/mesa/mesa_%.bbappend
```diff
@@ -15,10 +15,7 @@ python () {
 }
 
 # Enable Etnaviv and Freedreno support
-PACKAGECONFIG_append_use-mainline-bsp = " gallium"
-PACKAGECONFIG_append_use-mainline-bsp_armv7a = " etnaviv freedreno kmsro vc4"
-PACKAGECONFIG_append_use-mainline-bsp_armv7ve = " etnaviv freedreno kmsro vc4"
-PACKAGECONFIG_append_use-mainline-bsp_mx8 = " etnaviv kmsro vc4"
+PACKAGECONFIG_append_use-mainline-bsp = " gallium etnaviv kmsro vc4 freedreno"
```
petegriffin, line 18: shouldn't it just be " gallium etnaviv kmsro"?

Why do you want to enable vc4 and freedreno in meta-freescale layer?

zandrey: @otavio This would also enable `freedreno` for entire imx[6/7/8], which IMHO should use `etnaviv` only. The only derivative I know with Adreno GPU is i.MX5-series.

@petegriffin Agree, according to this [patch](https://patchwork.openembedded.org/patch/159857/) `vc4` can be dropped, as it was a requirement for `kmsro` build from *pre-meson* era. 😄

otavio: > @otavio This would also enable `freedreno` for entire imx[6/7/8], which IMHO should use `etnaviv` only. The only derivative I know with Adreno GPU is i.MX5-series.

When using mainline we can share the binaries among them, do yes, it needs to be enabled so all those SoC work.

> @petegriffin Agree, according to this [patch](https://patchwork.openembedded.org/patch/159857/) `vc4` can be dropped, as it was a requirement for `kmsro` build from *pre-meson* era. 😄 

`vc4` can indeed to be dropped. Will prepare a fix dropping it.

PR #382, opened 2020-05-02
imx-boot: fix mkimage_uboot linking issue

liuming50:
In imx-boot's Makefile, it downloads a native tool mkimage_uboot and
executes it, but this native tool requires libssl.so.1.1, which is in
${STAGING_LIBDIR_NATIVE}, we need set LD_LIBRARY_PATH to include it, or
it might run into following errors:
| ./mkimage_uboot -E -p 0x3000 -f u-boot.its u-boot.itb
| ./mkimage_uboot: error while loading shared libraries: libssl.so.1.1: cannot open shared object file: No such file or directory

Signed-off-by: Ming Liu <ming.liu@toradex.com>

### DIFF
--- a/recipes-bsp/imx-mkimage/imx-boot_0.2.bb
+++ b/recipes-bsp/imx-mkimage/imx-boot_0.2.bb
@@ -105,6 +105,8 @@ compile_mx8x() {
     cp ${DEPLOY_DIR_IMAGE}/${UBOOT_NAME}                     ${BOOT_STAGING}/u-boot.bin
 }
 do_compile() {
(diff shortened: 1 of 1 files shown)


liuming50:
Hmm, just saw this: https://github.com/Freescale/meta-freescale/pull/381, same problem, closing my pull request, sorry for disturbing.

otavio:
Your approach seems better.

petegriffin:
@liuming50 I didn't think of setting LD_LIBRARY_PATH :)
@otavio are you going to merge this one? Can you also apply on Dunfell branch?

otavio:
Yes, just waiting few days and also to gather more changes to go. Likely during this morning we merge this.

PR #385, opened 2020-05-05
Update kernel perf patch for linux-imx and linux-qoriq packages

zandrey:
As it has been noted in #376, initial commit of the `perf` patch for `linux-imx` has `Upstream-Status:` tag missing, which made it hard to trace. Tag is introduced by this PR, and `linux-imx` patch is refreshed.

The same patch is also required for `linux-qoriq` kernel package, therefore a copy of the same patch is added to `linux-qoriq` folder and recipe.

This PR supersedes the PR #376 

@otavio @rehsack Let's keep the discussion started in #376 in this PR as it reflects the proposal closer to the point.

-- andrey


rehsack:
Hey, we have a master-next? How cool is that?

otavio:
@rehsack I've been keeping it on sync until we finish those changes.

rehsack:
@otavio: maybe it's a good thing we should keep (mind the CI proposal). Not today, but hopefully soonish ... like this year :)

PR #390, opened 2020-05-06
linux-imx: Remove duplicate CRYPTO_SHA512 from defconfig

kevswims:
The mx8 defconfig already contains CRYPTO_SHA512 as a module which is
what is honored so this removes the extra line that also sets
CRYPTO_SHA512 to be built in

PR #391, opened 2020-05-07
Update Kernel recipes and resolve config issues reported by sanity checks

zandrey:
`linux-fslc` and `linux-fslc-imx` recipes are upgraded to *v5.4.39* tag from *korg*.

`linux-fslc-imx` received an additional set of patches from `lf-5.4.y` tracking branc in NXP tree

Config files for `linux-imx` and `linux-fslc-imx` have conflicting items reported by Yocto config verification mechanisms resolved (with a help from @kevswims).

This PR contains the original patch from #390 and supersedes it.

-- andrey


thochstein:
Oops, the kernel team pointed out a bigger issue here. We switched from the default `defconfig` to a custom `imx_v8_defconfig`.

https://source.codeaurora.org/external/imx/linux-imx/tree/arch/arm64/configs?h=lf-5.4.y

Seems this was done to align i.MX and QorIQ kernels, and includes a custom config fragment mechanism:

https://source.codeaurora.org/external/imx/meta-imx/tree/meta-bsp/recipes-kernel/linux/linux-imx_5.4.bb?h=zeus-5.4.3-1.0.0#n54

zandrey:
> Oops, the kernel team pointed out a bigger issue here. We switched from the default `defconfig` to a custom `imx_v8_defconfig`.

@thochstein Thanks for pointing this out!

This is a bit confusing, since until now `defconfig` was always used to provide a kernel configuration file... 😕 

Nevertheless, I guess this should be trivial work: replace the `defconfig` in `linux-imx` and `linux-fslc-imx` recipes to the content from `imx_v8_defconfig`.

> Seems this was done to align i.MX and QorIQ kernels, and includes a custom config fragment mechanism:
> 
> https://source.codeaurora.org/external/imx/meta-imx/tree/meta-bsp/recipes-kernel/linux/linux-imx_5.4.bb?h=zeus-5.4.3-1.0.0#n54

I had a look at the recipe and want to clear few things up here, just for my understanding.

QorIQ has 3 config files: one base and 2 fragments which are added to `DELTA_KERNEL_DEFCONFIG` like this:
```python
DELTA_KERNEL_DEFCONFIG ?= ""
DELTA_KERNEL_DEFCONFIG_prepend_qoriq-arm64 = "lsdk.config "
DELTA_KERNEL_DEFCONFIG_prepend_fsl-lsch2-32b = "multi_v7_lpae.config multi_v8.config lsdk.config "
DELTA_KERNEL_DEFCONFIG_prepend_ls102xa = "multi_v7_lpae.config lsdk.config "
```
With the new kernel recipe upgrade, QorIQ can take `lsdk,conf` as `defconfig` in the layer, and have `multi_v7_lpae.config` and `multi_v8.config` added as kernel fragments into the build based on the machine via `scc` files. This would be interesting for @rehsack to use further I believe.

*i.MX* had only one `imx_v8_defconfig` - it is safe to update the *i.MX* `defconfig` to the content from `imx_v8_defconfig`.

If you can take the new recipe layout into the NXP BSP distribution - then custom config merging mechanisms could be dropped, and standard Yocto mechanisms could be used instead, which would make NXP recipes much simpler.

What do you think about it?

-- andrey

thochstein:
> This is a bit confusing, since until now `defconfig` was always used to provide a kernel configuration file... 😕
> 
> Nevertheless, I guess this should be trivial work: replace the `defconfig` in `linux-imx` and `linux-fslc-imx` recipes to the content from `imx_v8_defconfig`.

Sounds good. I asked the kernel team to investigate whether we need to keep `defconfig` around.

> If you can take the new recipe layout into the NXP BSP distribution - then custom config merging mechanisms could be dropped, and standard Yocto mechanisms could be used instead, which would make NXP recipes much simpler.

👍 Already entered a ticket.

PR #396, opened 2020-05-13
upgrade: firmware-imx and imx-sc-firmware

panfir:


recipes-bsp/firmware-imx/firmware-imx_8.6.bb
```diff
@@ -0,0 +1,118 @@
...
+ALLOW_EMPTY_${PN} = "1"
+
+PACKAGES_DYNAMIC = "${PN}-vpu-* ${PN}-sdma-*"
+
+PACKAGES =+ "${PN}-epdc ${PN}-scfw ${PN}-sdma ${PN}-easrc ${PN}-regulatory ${PN}-hdmi"
+
+RDEPENDS_${PN}-epdc = "bash"
+RDEPENDS_${PN}-sdma = "bash"
+RDEPENDS_${PN}-regulatory = "bash"
+RDEPENDS_${PN}-hdmi = "bash"
```
zandrey, line 109: How about those who would like to build GPLv3-less images?

`bash` cannot be used as a shell on those, therefore what this recipe would do it to lockdown licensing terms for images, since those FW blobs are quite essential.

Could those additional scripts be re-worked to be shell-neutral and *bashism-free*?

recipes-bsp/firmware-imx/firmware-imx_8.6.bb
```diff
@@ -0,0 +1,118 @@
...
+
+FILES_${PN}-epdc = "${base_libdir}/firmware/imx/epdc/ ${sysconfdir}/epdc ${systemd_system_unitdir}/epdc-firmware.service"
+FILES_${PN}-scfw = "${base_libdir}/firmware/scfw/"
+FILES_${PN}-sdma = "${base_libdir}/firmware/imx/sdma ${sysconfdir}/sdma ${systemd_system_unitdir}/sdma-firmware.service"
+FILES_${PN}-easrc = "${base_libdir}/firmware/imx/easrc/"
+FILES_${PN}-regulatory = "${sysconfdir}/regulatory ${systemd_system_unitdir}/regulatory-firmware.service"
+FILES_${PN}-hdmi = "${base_libdir}/firmware/imx/hdmi/ ${sysconfdir}/hdmi ${systemd_system_unitdir}/hdmi-firmware.service"
+
+COMPATIBLE_MACHINE = "(imx)"
```
zandrey, line 118: Question from my side: is `imx` a valid MACHINEOVERRIDE? I haven't seen it so far, but maybe I just missed it...

recipes-bsp/firmware-imx/files/hdmi
```diff
@@ -0,0 +1,28 @@
...
+cat /lib/firmware/imx/hdmi/hdmirxfw.bin > data
+echo 0 > loading
+fi
+
+if [ 0 -lt $(ls $MX8_HDMIDP 2>/dev/null | wc -w) ]; then
+cd ${MX8_HDMIDP}
+echo 1 > loading
+cat /lib/firmware/imx/hdmi/dpfw.bin > data
+echo 0 > loading
+fi
```
zandrey, line 28: Can this whole part be simplified with using a function?

recipes-bsp/firmware-imx/firmware-imx_8.6.bb
```diff
@@ -0,0 +1,118 @@
...
+SYSTEMD_PACKAGES = "${PN}-sdma ${PN}-epdc ${PN}-regulatory ${PN}-hdmi"
+SYSTEMD_SERVICE_${PN}-sdma = "sdma-firmware.service"
+SYSTEMD_SERVICE_${PN}-epdc = "epdc-firmware.service"
+SYSTEMD_SERVICE_${PN}-regulatory = "regulatory-firmware.service"
+SYSTEMD_SERVICE_${PN}-hdmi = "hdmi-firmware.service"
+
+do_install() {
+    install -d ${D}${base_libdir}/firmware/imx
+
+    if ${@bb.utils.contains('DISTRO_FEATURES','systemd','true','false',d)}; then
```
zandrey, line 33: This mean there would be no scripts provided unless `systemd` is present?

recipes-bsp/firmware-imx/firmware-imx_8.6.bb
```diff
@@ -0,0 +1,118 @@
...
+            bbnote Excluding folder $d
+            ;;
+        *)
+            cp -rfv $d ${D}${base_libdir}/firmware
+            ;;
+        esac
+    done
+    cd -
+
+    # Install SDMA Firmware: sdma-imx6q.bin & sdma-imx7d.bin into lib/firmware/imx/sdma
```
zandrey, line 60: Not sure, but are those FW packages needed for *i.MX8M* family?

recipes-bsp/firmware-imx/firmware-imx_8.6.bb
```diff
@@ -0,0 +1,118 @@
...
+    mv ${D}${base_libdir}/firmware/easrc/ ${D}${base_libdir}/firmware/imx/easrc/
+
+    # Install HDMI Firmware: hdmitxfw.bin, hdmirxfw.bin & dpfw.bin into lib/firmware/imx/hdmi
+    install -d ${D}${base_libdir}/firmware/imx/hdmi
+    mv ${D}${base_libdir}/firmware/hdmi/cadence/hdmitxfw.bin ${D}${base_libdir}/firmware/imx/hdmi
+    mv ${D}${base_libdir}/firmware/hdmi/cadence/hdmirxfw.bin ${D}${base_libdir}/firmware/imx/hdmi
+    mv ${D}${base_libdir}/firmware/hdmi/cadence/dpfw.bin ${D}${base_libdir}/firmware/imx/hdmi
+
+    find ${D}${base_libdir}/firmware -type f -exec chmod 644 '{}' ';'
+    find ${D}${base_libdir}/firmware -type f -exec chown root:root '{}' ';'
```
zandrey, line 77: Why is this required?

And if you use `find` - please use `+` instead of `;` for performance reasons... 😉

recipes-bsp/firmware-imx/firmware-imx_8.6.bb
```diff
@@ -0,0 +1,118 @@
...
+        ddr|seco)
+            # These folders are for i.MX 8 and are included in the boot image via imx-boot
+            bbnote Excluding folder $d
+            ;;
+        *)
+            cp -rfv $d ${D}${base_libdir}/firmware
+            ;;
+        esac
+    done
+    cd -
```
zandrey, line 58: Can this whole construct be simplified? For example, there is a mechanism in NXP `imx-gpu-sdk` [recipe](https://source.codeaurora.org/external/imx/meta-imx/tree/meta-sdk/recipes-graphics/imx-gpu-sdk/imx-gpu-sdk_5.4.0.bb?h=zeus-5.4.3-1.0.0#n66) which has a filtering and removal mechanism that is more elegant.

recipes-bsp/firmware-imx/firmware-imx_8.6.bb
panfir: Not sure, it's inherited. I wasn't focus on this

recipes-bsp/firmware-imx/firmware-imx_8.6.bb
panfir: Hmm, good point on this one, I'll look into. Thanks!

> @panfir I dropped few comments and questions regarding this PR.
> 
> In general, can you use `.sh` extensions for scripts? This would be easier to identify on the target then.
> 
> -- andrey

Well, it's good to keep it that way, the `shebang` is enough.
It is not meant to be explicitly identified on the target as a regular script, it is run as `systemd.service` at boot.
I'm not sure any rules about it but I haven't seen the use of extensions to the scripts used by `systemd` services.
I general, I think this rule is used:
.. /`script_name`
.. /`script_name.service`

recipes-bsp/firmware-imx/firmware-imx_8.6.bb
panfir: Yes

recipes-bsp/firmware-imx/firmware-imx_8.6.bb
panfir: Yes. 
Those scripts are running via systemd service (e.g. sdma-firmware.service)

recipes-bsp/firmware-imx/firmware-imx_8.6.bb
panfir: Yes, one of those bins
- sdma-imx7d.bin for i.MX7D/i.MX8M use
- sdma-imx6q.bin for all i.MX6

recipes-bsp/imx-sc-firmware/imx-sc-firmware_1.2.10.1.bb
```diff
@@ -1,5 +1,5 @@
 # Copyright (C) 2016 Freescale Semiconductor
-# Copyright (C) 2017-2019 NXP
+# Copyright 2017-2020 NXP
```
otavio, line 2: Useless change, please drop.

recipes-bsp/imx-sc-firmware/imx-sc-firmware_1.2.10.1.bb
```diff
@@ -35,4 +35,4 @@ INHIBIT_PACKAGE_STRIP = "1"
 INHIBIT_PACKAGE_DEBUG_SPLIT = "1"
 PACKAGE_ARCH = "${MACHINE_ARCH}"
 
-COMPATIBLE_MACHINE = "(mx8qm|mx8qxp)"
+COMPATIBLE_MACHINE = "(mx8qm|mx8x)"
```
otavio, line 38: Why this was changed? it seems wrong. If so, please explain on commit log.

recipes-bsp/firmware-imx/files/epdc
```diff
@@ -0,0 +1,12 @@
...
+# firmware loader script
+
+MX7_EPDC="/sys/bus/platform/devices/*/firmware/imx!epdc!epdc_ED060XH2C1.fw"
+
+if [ 0 -lt $(ls $MX7_EPDC 2>/dev/null | wc -w) ]; then
+cd ${MX7_EPDC}
+echo 1 > loading
+cat /lib/firmware/imx/epdc/epdc_ED060XH2C1.fw > data
+echo 0 > loading
+fi
```
otavio, line 12: This should be done using udev, and not a `.service` and script. Also, this should be a separated commit as it is not related to the upgrade.

recipes-bsp/firmware-imx/files/hdmi
otavio: Again, let's move this to udev.

recipes-bsp/imx-sc-firmware/imx-sc-firmware_1.2.10.1.bb
panfir: I don't think is wrong. I used that one internally to match another machine, not upstream yet. I'll change it back

recipes-bsp/firmware-imx/files/hdmi
panfir: > Again, let's move this to udev.

..Yeah, it seems to be the right place

recipes-bsp/firmware-imx/firmware-imx_8.6.bb
panfir: > How about those who would like to build GPLv3-less images?
> 
> `bash` cannot be used as a shell on those, therefore what this recipe would do it to lockdown licensing terms for images, since those FW blobs are quite essential.

Hi Andrey, I just checked with our trusted advisors ..
As far as Licensing goes, it depends on what these are doing with bash.  If bash is being compiled into anything, that is a problem.  If we are modifying bash and re-distributing it, that could be a problem.  If it is just being required because we are using bash to run in and we are just pulling it from the external source, it is not a problem that bash is GPL-v3.

> 
> Could those additional scripts be re-worked to be shell-neutral and _bashism-free_?

recipes-bsp/imx-sc-firmware/imx-sc-firmware_1.2.10.1.bb
otavio: Not upstreamed yet is wrong. Also it dropped a valid one.

panfir: Not upstreamed because is not ready yet for public.
I just checked 
`./conf/machine/imx8qxpmek.conf:6:MACHINEOVERRIDES =. "mx8:mx8x:mx8qxp:"`
But anyway, I'll change it back to `mx8qxp`

recipes-bsp/firmware-imx/firmware-imx_8.6.bb
zandrey: > > How about those who would like to build GPLv3-less images?
> > `bash` cannot be used as a shell on those, therefore what this recipe would do it to lockdown licensing terms for images, since those FW blobs are quite essential.
> 
> Hi Andrey, I just checked with our trusted advisors ..
> As far as Licensing goes, it depends on what these are doing with bash. If bash is being compiled into anything, that is a problem. If we are modifying bash and re-distributing it, that could be a problem. If it is just being required because we are using bash to run in and we are just pulling it from the external source, it is not a problem that bash is GPL-v3.
> 

I have a different point of view regarding usage of GPLv3 components is final target images, namely the so-called `Tivoization clause`. You might want to get back to the legal department with exactly this question, but my understanding of that clause is:

Once GPLv3-licensed component is included in any firmware image which is signed by vendor and provided as a binary to any party, then this party can request from the firmware vendor mechanisms and means to re-create and re-flash that firmware image, which automatically means that firmware vendor should make signing keys and facilities public.

Usually, this is is not a concern as most of devices built are purposed in the way they could be updated without having those signatures in place.

But some HW (and firmware) vendors are placing restrictions on their design (security, vendor lock-ins, etc) and therefore opting-out from GPLv3 components in their images. This could be done in Yocto by blacklisting GPLv3 and excluding all components from the image.

What this commit would effectively do - it introduces the core component required for HW to operate and at the same time enforcing the license for the dependent component (`bash` in this case) to be set to GPLv3, automatically making final FW file non-GPLv3-free, thus making all those HW and FW vendors above to turn away from this platform, or seek alternatives like taking an older version of this layer.

I'm not really sure if this was an intention here with introducing `bash` dependency.

As I said: I'm not a layer, therefore please treat above as my personal opinion and a guideline, rather than a legal advise or any statement of some sort. You might would want to check the above with the Legal department.

*"If in doubt - call the Lawyer..."* 😉 

-- andrey

recipes-bsp/firmware-imx/firmware-imx_8.6.bb
zandrey: Got it, thanks

recipes-bsp/firmware-imx/firmware-imx_8.6.bb
zandrey: OK, then why would we install FW files below if there are no scripts to load them on the device? 😉

recipes-bsp/firmware-imx/firmware-imx_8.6.bb
zandrey: > Yes, one of those bins
> 
>     * sdma-imx7d.bin for i.MX7D/i.MX8M use
> 
>     * sdma-imx6q.bin for all i.MX6

Can they be made machine-dependent here? Having i.MX6 FW file on i.MX8 could be confusing, visa-versa might be even dangerous. 😄

kevswims: Can the name of the sdma-imx7d.bin be changed to also indicate that it is applicable to imx8m? I have run into multiple spots over the last few weeks where trying to figure out if something is applicable to my particular machine is impossible because of naming being specific to an older part but the new parts also use those resources.

otavio: In worse case we could add an RPROVIDES to it, and use the imx8m package name.

recipes-bsp/firmware-imx/firmware-imx_8.6.bb
panfir: It will not be the case, we'll get rid of bash from the f/w loading scripts. Interestingly, the so-called `Tivoization clause` and Yes, GPLv3 is a problem, we want to avoid it as much as possible. Thanks!

recipes-bsp/firmware-imx/firmware-imx_8.6.bb
panfir: The condition was more for `.services`
But Yes, this will be corrected once we switch to `udev`.
I just started working on it. It will take a while to test in our labgrid, on different imx6/7/8 boards..or maybe thro' some CI

recipes-bsp/firmware-imx/firmware-imx_8.6.bb
panfir: > Can the name of the sdma-imx7d.bin be changed to also indicate that it is applicable to imx8m? I have run into multiple spots over the last few weeks where trying to figure out if something is applicable to my particular machine is impossible because of naming being specific to an older part but the new parts also use those resources.

I know it's confusing, the firmware guys don't change names too often.
In a way it makes sense considering that the imx8M family is derived from imx7d, and the sdma IP is exactly the same as imx7d.

kevswims: > I know it's confusing, the firmware guys don't change names too often.
> In a way it makes sense considering that the imx8M family is derived from imx7d, and the sdma IP is exactly the same as imx7d.

For some background here, I have only ever used the imx8m parts and finding out which bits of the 8m IP are the same as previous generations and which are new or changed is difficult. As a newcomer to the i.MX series this has been really confusing to me and I have not found any sort of documentation that compares the different generations as to what changed and what did not.

panfir:
> Please organize the commits; they can be squashed as there are no need of multiple commits changing same recipe doing and reverting those changes.

done

recipes-bsp/firmware-imx/firmware-imx_8.6.bb
otavio: With the udev rules, no dependency to systemd is need.

recipes-bsp/firmware-imx/firmware-imx_8.6.bb
otavio: We likely will need to cover the `use-mainline-bsp` as well.

zandrey:
@thochstein @panfir 

Looks like there are few missing pieces in the *8.8* firmware delivery that I just stumbled upon and hope that you can help me out to resolve it:

1. I cannot seem to find the VPU FW file for `imx8qxp`, instead archive contains `vpu_fw_imx8_dec.bin` and `vpu_fw_imx8_enc.bin` files. Do you which one of those (or neither) should be packaged for `imx8qxp`?
2. On the contrary: `imx27` has 2 VPU FW  files - `vpu_fw_imx27_TO1.bin` and `vpu_fw_imx27_TO2.bin`. Do you know which one should be actually used?

It would be really great if you can clarify this! Thanks a lot!

-- andrey

thochstein:
> 1. I cannot seem to find the VPU FW file for `imx8qxp`, instead archive contains `vpu_fw_imx8_dec.bin` and `vpu_fw_imx8_enc.bin` files. Do you which one of those (or neither) should be packaged for `imx8qxp`?

Seems both:

```
r60874@tx30imx-01:~/zeus/build/tmp/work/all-poky-linux/firmware-imx/1_8.10-r0/packages-split$ ls -R firmware-imx-vpu-imx8
firmware-imx-vpu-imx8:
lib

firmware-imx-vpu-imx8/lib:
firmware

firmware-imx-vpu-imx8/lib/firmware:
vpu

firmware-imx-vpu-imx8/lib/firmware/vpu:
vpu_fw_imx8_dec.bin  vpu_fw_imx8_enc.bin
```

thochstein:
> 2. On the contrary: `imx27` has 2 VPU FW  files - `vpu_fw_imx27_TO1.bin` and `vpu_fw_imx27_TO2.bin`. Do you know which one should be actually used?

Again seems both:

```
r60874@tx30imx-01:~/zeus/build/tmp/work/all-poky-linux/firmware-imx/1_8.10-r0/packages-split$ ls -R firmware-imx-vpu-imx27
firmware-imx-vpu-imx27:
lib

firmware-imx-vpu-imx27/lib:
firmware

firmware-imx-vpu-imx27/lib/firmware:
vpu

firmware-imx-vpu-imx27/lib/firmware/vpu:
vpu_fw_imx27_TO1.bin  vpu_fw_imx27_TO2.bin
```

PR #400, opened 2020-05-14
qoriq: remove usdpaa dtb and relocate sdk dtb

ting-liu:
To build with pure mainline kernel, update dtb in KERNEL_DEVICETREE
* usdpaa is an unsupported feature now, remove dtb for it.
* add QORIQ_DEFAULT_BSP definition to easily choose which BSP to use
* dtb with -sdk suffix only exist in kernel from NXP bsp, append them on NXP bsp only.

### DIFF
--- a/conf/machine/ls1043ardb.conf
+++ b/conf/machine/ls1043ardb.conf
@@ -22,7 +22,7 @@ UBOOT_CONFIG ??= "tfa-secure-boot tfa"
 UBOOT_CONFIG[tfa] = "ls1043ardb_tfa_defconfig,,u-boot-dtb.bin"
 UBOOT_CONFIG[tfa-secure-boot] = "ls1043ardb_tfa_SECURE_BOOT_defconfig,,u-boot-dtb.bin"
 
-KERNEL_DEVICETREE ?= "freescale/fsl-ls1043a-rdb-sdk.dtb freescale/fsl-ls1043a-rdb-usdpaa.dtb freescale/fsl-ls1043a-qds.dtb freescale/fsl-ls1043a-qds-sdk.dtb"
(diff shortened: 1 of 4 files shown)


conf/machine/ls1043ardb.conf
```diff
@@ -22,7 +22,10 @@ UBOOT_CONFIG ??= "tfa-secure-boot tfa"
 UBOOT_CONFIG[tfa] = "ls1043ardb_tfa_defconfig,,u-boot-dtb.bin"
 UBOOT_CONFIG[tfa-secure-boot] = "ls1043ardb_tfa_SECURE_BOOT_defconfig,,u-boot-dtb.bin"
 
-KERNEL_DEVICETREE ?= "freescale/fsl-ls1043a-rdb-sdk.dtb freescale/fsl-ls1043a-qds.dtb freescale/fsl-ls1043a-qds-sdk.dtb"
```
otavio: Please don't. For NXP specific one, please use:

```
KERNEL_DEVICETREE_append_use-nxp-bsp = "  ... "
```

ting-liu: so qoriq needs to set similar var as IMX_DEFAULT_BSP. can we rename IMX_DEFAULT_BSP to be more generic and move its definition from imx-base.inc? or define a new QORIQ_DEFAULT_BSP?

otavio: Define a new one.

ting-liu: define QORIQ_DEFAULT_BSP and use KERNEL_DEVICETREE_append_use-nxp-bsp. PR is updated.

Issue #389, opened 2020-05-06
Weird behavior using menuconfig with configuration fragments


kevswims:
I am trying to use the new support for configuration fragments that was introduced in https://github.com/Freescale/meta-freescale/pull/386. I am following the instructions from the kernel-dev manual here: https://www.yoctoproject.org/docs/latest/kernel-dev/kernel-dev.html#creating-config-fragments to create configuration fragments. I have run the `kernel_configme` step, run the `menuconfig` step and then generated a fragment using `diffconfig`. 

After adding that fragment to my BSP I wanted to make additional changes so I reran the `kernel_configme` step and checked that the `.config` file generated contained the correct changes from my first fragment and it did. However, when I go ahead and run the `menuconfig` step all those changes go away and are not present in the `.config` anymore or in the menuconfig gui. If I load the previous .config into menuconfig then I can see all the old changes I am expecting but then running `diffconfig` creates a fragment with all the new and old changes which is not what I want since I am trying to keep the fragments separate. 

Am I missing a step here or is there a bug in the `menuconfig` step that is causing it to not pull the fragments into the config before it runs the tool?

Overall these kernel improvements are a big improvement though!

Thanks!

otavio:
You need to make multiple configuration fragments, for individual change/feature.

kevswims:
That's what I was trying to do. The issue I am having is that the changes from the first one are not reflected in menuconfig when I am working on the second one.

Somewhat unrelated to my initial issue. It's really difficult to make a slimmed down kernel based on the default defconfig because the default one contains so many selections that are not applicable to the hardware. It seems a little odd to me to have a bunch of configuration for SOCs that are not imx8's in the mx8 default defconfig. Would there be interest in working on a slimmed down one that only supports the hardware on the eval kit for example?

zandrey:
> That's what I was trying to do. The issue I am having is that the changes from the first one are not reflected in menuconfig when I am working on the second one.

Have you tried to generate first fragment, include it via `SRC_URI` and do a clean build? `menuconfig` together with `kernel_configme` is (from what I understand) brings the kernel to a known state which is the *defconfig* one.

This actually ensures that consecutive fragments would not include modifications introduced by already existing ones, hence the point from @otavio above is totally valid:

> You need to make multiple configuration fragments, for individual change/feature.

After every feature/change, include the new fragment and compile the kernel. This would ensure the iterative state of the modifications you're doing.

[...]

PR #401, opened 2020-05-14
Kernel recipes consolidated update

zandrey:
This PR addresses following points:

1. Update both `linux-fslc` and `linux-fslc-imx` kernels to *v5.4.41* tag from korg
2. Resolve inconsistency in initial `defconfig` for both kernels by replacing the content with the one from `imx_v8_defconfig` kernel config file found in NXP tree (see discussion in PR #390)
3. Discard NXP kernel repository in `linux-imx-headers` and inherit kernelsrc class, which allows to reuse the shared workdir provided by kernel recipes.

@kevswims Your commit is squashed now, since the content of `imx_v8_defconfig` didn't have SHA512 mismatch on the first place.

@otavio @thochstein @rehsack 
Can you please have a look at the item 3 from the list above (this is commit 8cf7d62) and tell me if this is OK to be solved like this?

I do recall we had conversations about not to update headers with every release, and this change might have this effect. On the other hand: this offers consistency between various kernels, especially when switching between `linux-imx` and `linux-fslc-imx`, since in this case headers would be pulled from the respective kernel tree and not the fixed one. Besides, there is a big save on the fetch since kernel repo wouldn't have to be fetched twice (which is the case now).

If this modification is OK - my next target to re-work in the similar way would be `kernel-module-imx-gpu-viv_6.4.0.p2.0.bb` since it should use the same kernel shared workdir.

-- andrey


recipes-kernel/linux/linux-imx-headers_5.4.bb
```diff
@@ -7,15 +7,18 @@ New headers are installed in ${includedir}/imx."
...
 
-SRCBRANCH = "lf-5.4.y"
-LOCALVERSION = "-1.0.0"
-SRC_URI = "git://source.codeaurora.org/external/imx/linux-imx.git;protocol=https;branch=${SRCBRANCH}"
-SRCREV = "f8118585ee3c7025265b28985fdfe0af96a84466"
+# Inherit class that provides the kernel source.
+# This allows to switch kernels from NXP to combined NXP+fslc and not keep
+# fixed NXP-only repository to fetch headers, thus making it consistent
+# accross kernel trees chosen for the build.
+inherit kernelsrc
```
otavio, line 14: I understand the motivation about this however, we cannot change this across machines.

So this is the reason we had this "locked" under a revision. This provides a ABI and changing it means all packages using this (and its dependencies) become `MACHINE_ARCH`.

rehsack: Would there be any reason to install linux-imx-headers but let's say - linux-qoriq?
Is a sane "way out" maybe installing linux-xxx-headers with linux-xxx and just add such a package?

otavio: @rehsack we do not wish to replace common headers and we wish to reduce the headers installed here.

Doing so, we reduce the specific ABI we maintain so we don't make all packages machine specific. We also try to restrict the packages adding this so we have less specific packages.

Do you see a different approach?

zandrey: > Doing so, we reduce the specific ABI we maintain so we don't make all packages machine specific. We also try to restrict the packages adding this so we have less specific packages.

Agree, this is a very valid point.

I guess handling of this via `use-imx-headers.bbclass` ensures that recipes installing imx headers are contracting themselves to use NXP BSP only.

The next step would be to actually prohibit all recipes to add `linux-imx-headers` directly into their `DEPENDS` and rather inherit the class, which should IMHO complete the isolation.

> Do you see a different approach?

Actually, the cleanest way for me would be that NXP provides those headers in the archive (similar to how `linux-libc-headers` are provided), and make sure they do match to the kernel tree they release. But until this happens - fetching the whole NXP tree seems to be inevitable...

zandrey:
@otavio @rehsack Thanks for the review guys!

I had those kind of thought at the back of my mind when I brought this change, and that is the reason I explicitly asked in the PR for RFC regarding this.

I would drop the 8cf7d62754d67e3f785aa212de8541b41be2ebcd and update the PR.

One question from my side which is still open: should I do this change for `kernel-module-imx-gpu-viv_6.4.0.p2.0.bb`?

IMHO the driver should be built from the same kernel tree as Kernel itself, as it is not about propagating headers - it is about making sure source code consistency.

@rehsack:
> 
> As far as I understood, the unified kernel strategy targets mainly the RDBs. I'm quite unsure whether fslc could and should follow that strategy (in such a case, the appropriate explanation should be added to kernel recipe).

Not at all, I also do not see the `-fslc` kernels are only targeting RDBs. On the other hand - I really would like to see the NXP kernels are used for something else than RDB. 😉 

> One source is a heroic step either - and quite exhausting anyway.

But chances of further success are not nil. 😛 

> I think - but I'm not so far the request it to be changed - different platforms can and should use different defconfig's - e.g. QorIQ: There is no reason to enable Felix DSA everywhere just to support it on ls102x* ...

Here I totally agree - families should have their respective `defconfig`s and derivatives within faimly could be configured with fragments.
I do not think it makes a lot of sense to configure *i.MX* and *QorIQ* with one `defconfig` and keeping all those differences in separate fragments, rather use fragments within one family to enable/disable specific derivative features.

-- andrey

recipes-kernel/linux/linux-imx-headers_5.4.bb
rehsack: > @rehsack we do not wish to replace common headers and we wish to reduce the headers installed here.

I'm probably to far away from this specific problem ...

I understand it that-wise: linux-imx-headers are required to build against linux-imx-kernel some userspace executables - likely on target (otherwise the recipe itself can use `kernelsrc.bbclass`).
Consequently - on such a target with imx kernels, the linux-imx-headers are part of the linux-headers (or better: should be). Am I wrong with this conclusion?

rehsack:
> @otavio @rehsack Thanks for the review guys!
> 
> Not at all, I also do not see the `-fslc` kernels are only targeting RDBs. On the other hand - I really would like to see the NXP kernels are used for something else than RDB. 😉

We use the QorIQ kernel plus some enhancements on non-RDB HW ;)
So: they are :D
And maybe at a later point, when there is a fslc+qoriq, we officially base on this one.

> > One source is a heroic step either - and quite exhausting anyway.
> 
> But chances of further success are not nil. 😛

First things first ;)

> > I think - but I'm not so far the request it to be changed - different platforms can and should use different defconfig's - e.g. QorIQ: There is no reason to enable Felix DSA everywhere just to support it on ls102x* ...
> 
> Here I totally agree - families should have their respective `defconfig`s and derivatives within faimly could be configured with fragments.
> I do not think it makes a lot of sense to configure _i.MX_ and _QorIQ_ with one `defconfig` and keeping all those differences in separate fragments, rather use fragments within one family to enable/disable specific derivative features.

I think we should here follow your idea using feature-sets. Maybe we can convince NXP (@ting-liu) long-term-wise to join. this approach.

otavio:
@zandrey 
> One question from my side which is still open: should I do this change for `kernel-module-imx-gpu-viv_6.4.0.p2.0.bb`?

Yes, please. This does make sense. However, the default should be our forked module code.

recipes-kernel/linux/linux-imx-headers_5.4.bb
otavio: > I understand it that-wise: linux-imx-headers are required to build against linux-imx-kernel some userspace executables - likely on target (otherwise the recipe itself can use `kernelsrc.bbclass`).
> Consequently - on such a target with imx kernels, the linux-imx-headers are part of the linux-headers (or better: should be). Am I wrong with this conclusion?

On an ideal world, the imx-headers would not exist. The current approach was the better balance I could come with as we reduce the number of rebuilds.

Any idea how to improve this even more?

kevswims:
@rehsack @zandrey 
> > Here I totally agree - families should have their respective `defconfig`s and derivatives within faimly could be configured with fragments.
> > I do not think it makes a lot of sense to configure _i.MX_ and _QorIQ_ with one `defconfig` and keeping all those differences in separate fragments, rather use fragments within one family to enable/disable specific derivative features.
> 
> I think we should here follow your idea using feature-sets. Maybe we can convince NXP (@ting-liu) long-term-wise to join. this approach.

This would be ideal for our use case which I brought up here: https://github.com/Freescale/meta-freescale/issues/389#issuecomment-624990672. I have spent a lot of time figuring out which features can be disabled and which cannot to try and slim down the kernel. Our application has no graphical environment but we still utilize the GPU for processing so it's a lot of effort right now to strip the current defconfig's down to something usable. For us the ideal solution would be something where the base defconfig is just enough to get a console to boot on the RDB and then all other features are added through fragments.

zandrey:
> 
> 
> @zandrey
> 
> > One question from my side which is still open: should I do this change for `kernel-module-imx-gpu-viv_6.4.0.p2.0.bb`?
> 
> Yes, please. This does make sense. However, the default should be our forked module code.

Sorry, I think I was misleading with my last question: I meant that I wanted to introduce the `kernelsrc` inheritance in `kernel-module-imx-gpu-viv_6.4.0.p2.0.bb`. In this case we cannot default to any tree - it would be provided by a respective kernel recipe.

@otavio with your reply you maybe meant that I should replace the `SRC_URI` to point to `linux-fslc` repo per default?

zandrey:
> 
> I think we should here follow your idea using feature-sets. Maybe we can convince NXP (@ting-liu) long-term-wise to join. this approach.

I was thinking to make an example here on how to include `CONFIG_TRACEPOINTS` in case if *tools-profile* is selected in features, thus making the build clean out of warnings and functional in runtime. 😉 This is ambitious, but I wanted to try it out anyway.

Feature-sets are no *Voodoo Magic*, and is actually really beneficial in terms of scalability and tracking. I really hope that NXP can consider this mechanism to move on with kernel recipe and tree maintenance.

But to get start with it - we would need a bare minimal `defconfig` (as @kevswims indicated) which should actually come from NXP guys since they know the SoC *inside-out* and can slim down the configuration to the one providing only basic SoC support. The rest should go into [kernel fragments]/[yocto feature sets], IMHO.

zandrey:
OK, guys! I've dropped `linux-imx-header` from the PR and decided to postpone the work on `kernel-module-imx-gpu-viv_6.4.0.p2.0.bb` since it is a bit more complicated than I thought. I'd come back to it later in another PR, also in order to not stall the kernel upgrade.

Please have a look further here.

-- andrey

PR #411, opened 2020-05-25
QorIQ tidying

rehsack:
While hacking on https://github.com/rehsack/meta-freescale/commit/6d9db907aabac0a3f80cc85a547bfd6ccd4783b1 (@rabeeh, maybe you want to take a look, either), I stumbled over minor issues worth to be cleaned up.


rehsack:
Updated it with another "finding" :)

PR #422, opened 2020-06-02
[zeus] Cherry-pick qtbase i.MX 8 configuration

thochstein:


otavio:
Please cherry-pick the fix for vulkan we just merged in #426 as we're affected here as well.

PR #413, opened 2020-05-26
optee-os: fix gcc10 build for imx fork

zandrey:
OP-TEE OS component fails to build on gcc10 due to missing __getauxval
function. This can be resolved via disabling of outline-atomics feature
in gcc10.

Backport upstream patch [1] with adaptions to optee-imx version.

Original patch uses $(call cc-option,-mno-outline-atomics,) to disable
outline-atomics which itself is broken and required 4 additional patches
to be compatible with gcc and clang.

This is resolved in the upstream PR [2], which is applicable to the
latest upstream version.

NXP fork contains old version of OP-TEE, and therefore would receive the
update for Makefiles with new version. Since it is not required to make
current NXP version compatible with old compiler versions on master
branch - outline-atomics are disabled directly.

[1]: https://github.com/OP-TEE/optee_os/pull/3891/commits/e07c2b062846df4385542f4e4fe08b40cf7d8191
[2]: https://github.com/OP-TEE/optee_os/pull/3891

Fixes #407

PR #437, opened 2020-06-17
OP-TEE components fixes

zandrey:
This PR is a follow-up of #413 and addresses issues with `optee-os` and `optee-client` components.

`optee-os` received a backport of upstream PR which solves compilation issues when GCC10 is used, together with providing a backward compatibility with GCC9

`optee-client` received a patch refresh, which is a backport of originally submitted one. Original patch is re-worked to accommodate additional cases, and is not compatible with current NXP fork, therefore it was backported after proposed patch has been accepted upstream.

@petegriffin Can you please have a look at this PR, as it should address the GCC9 build issue for you.

-- andrey


petegriffin:
@zandrey Thanks for spending the time to fix this :) I've kicked off a build on linaro ci system which incorporates this PR (see here https://ci.linaro.org/job/rpb-openembedded-master/1385/). I'll update once it has actually been processed, as it is still pending atm.

petegriffin:
@zandrey previous link was incorrect the build with this PR included is https://ci.linaro.org/job/rpb-openembedded-master/1386/DISTRO=rpb-wayland,MACHINE=imx8mqevk,label=docker-stretch-amd64/console

It completed all optee-os parts fine, but failed like we spoke about before on cogl.

zandrey:
@petegriffin Thanks again for testing this PR!

As for `cogl`: this is exactly the same error I've observed as well, and unfortunately I still have no clean solution to this issue... I've put a brief description of what the issue is in #413, and would duplicate it in a separate issue here. It look like there is a configuration missing in `mesa` that causes *autoconf* to identify that `EGL/eglmesaext.h` is present, but it is not found during the build...

-- andrey

PR #454, opened 2020-08-13
[zeus] fsl-eula-unpack: exclude FSL_EULA_FILE from sstate signature

sapiippo:
Use vardepsexclude for FSL_EULA_FILE variable to to make sstate cache
usable from different build directories.

(cherry picked from commit 4689674e6c147f6891b541c273abf2889aa6f2da)


otavio:
To backport this we need to do PRs for all branches up to Zeus, so we are with it fixes on all newer branches. Please prepare those so we merge them all together.

PR #468, opened 2020-08-28
imx-atf: fix build with -fno-common

ricardosalveti:
Backport fixes required to fix build with -fno-common. Build failure
found when building for a imx8qx based target.


otavio:
@ricardosalveti does master branch version also accept this patches? Could you prepare a PR against it as well? I'd like to keep master in a working state while we work on the new BSP.

@thochstein those need to be applied internally.

ricardosalveti:
> @ricardosalveti does master branch version also accept this patches? Could you prepare a PR against it as well? I'd like to keep master in a working state while we work on the new BSP.

Sure, opened #470.

thochstein:
> @thochstein those need to be applied internally.

Done.

PR #485, opened 2020-09-09
NXP 5.4.24-2.1.0 updates for EULA, imx-parser, alsa-lib

thochstein:


conf/machine/include/imx-base.inc
```diff
@@ -158,7 +158,6 @@ MACHINE_SOCARCH_FILTER_append_imx = " \
 "
 MACHINE_SOCARCH_FILTER_append_imxvpu = " \
     imx-codec \
-    imx-parser \
```
otavio, line 161: @thochstein how so? could you explain why the change?

otavio (changes requested):

conf/machine/include/imx-base.inc
thochstein: There's nothing machine or SOC-specific in the recipe?

Issue #475, opened 2020-09-03
Disabled v4l feature in OpenCV


krzk:
Disabling V4L2 in dynamic-layers/openembedded-layer/recipes-support/opencv/opencv_4.%.bbappend is really a pain:
1. It's important feature of OpenCV.
2. Explanation by Otavio Salvador <otavio@ossystems.com.br> when adding this line is not even correct and you keep pushing it to all customers:
_"The camera support does not work using v4l. Disable it."_
This is simply not true, as V4L works fine and it is a proper way to get pictures from Linux mainline sensors.

Seriously... remove this patch.

otavio:
Hello @krzk, I checked the `bbappend` file and it disabled it only when using the NXP BSP as is quoted below:
```
FILESEXTRAPATHS_prepend := "${THISDIR}/${PN}:"

PACKAGECONFIG_remove_imxgpu2d = "v4l"
```

I am copying @thochstein here so he can ask internally if we can drop this override.

So I went further and checked where the comment you are referring to comes from and it seems you are referring to the following commit:
```
commit 1a344295d13fc5a483743037b99a996a47f38baa
Author: Otavio Salvador <otavio@ossystems.com.br>
Date:   Fri Jul 17 09:05:50 2015 -0300

    opencv: Remove v4l support for all i.MX6
    
    The camera support does not work using v4l. Disable it.
    
    Signed-off-by: Otavio Salvador <otavio@ossystems.com.br>
```

and as it can be seen on the commit log, this is from **2015** and at that time, we really were not using V4L. So it seems, you are referring to a quite old branch and I'd like to ask you to upgrade your layer revision.

krzk:
All NXP BSPs have imxgpu2d  even if they do not actually support it yet (e.g. i.MX 8).
Zeus branch is not old. You kept this change all the time, copying, updating, rebasing till OpenCV v4.3 and Yocto v3.0.

zandrey:
> All NXP BSPs have imxgpu2d even if they do not actually support it yet (e.g. i.MX 8).

That is not quite true, and according to [NXP Comparison table](https://www.nxp.com/docs/en/supporting-information/IMX8LAYERCMPR.pdf) *i.MX8M Nano* derivative lacks 2D HW accelerator or blit engine of some sort. I'm not sure what you mean by support in this case - packages or HW support?

Packages are present in this layer, as they are ported from NXP BSPs. HW 2D blit support can be cross-checked in the PDF link above.

[...]

PR #486, opened 2020-09-10
gstreamer1.0-plugins-bad: Remove un-necessary opencv dependency

thochstein:
OpenCV is no longer required, so remove it from PACKAGECONFIG.


otavio (approved):

zandrey:
@thochstein Can we also drop the *dynamic-layers/openembedded-layer/recipes-support/opencv/opencv_4.%.bbappend*?

This was in a report of #475 and looks like it is no longer required...

-- andrey

thochstein:
> @thochstein Can we also drop the _dynamic-layers/openembedded-layer/recipes-support/opencv/opencv_4.%.bbappend_?
> 
> This was in a report of #475 and looks like it is no longer required...

Thanks @zandrey, that does seem to follow. I will do another PR.

PR #502, opened 2020-09-29
weston-imx: remove also xwayland from package config

zandrey:
Commit [5a5c5dd23ea0173ef16073c3c651aec89b5a67c1] removes x11 and
wayland from PACKAGECONFIG, but when wayland with X11 distro is used -
this leaves the xwayland package config option enabled.

Without x11 and wayland in PACKAGECONFIG, xwayland cannot be enabled so
remove this config option as well.

### DIFF
--- a/recipes-graphics/wayland/weston_8.0.0.imx.bb
+++ b/recipes-graphics/wayland/weston_8.0.0.imx.bb
@@ -144,7 +144,7 @@ PACKAGECONFIG_OPENGL              = "opengl"
 PACKAGECONFIG_OPENGL_imxgpu2d     = ""
 PACKAGECONFIG_OPENGL_imxgpu3d     = "opengl"
 
-PACKAGECONFIG_remove = "wayland x11"
+PACKAGECONFIG_remove = "wayland x11 xwayland"
 PACKAGECONFIG_append = " ${@bb.utils.filter('DISTRO_FEATURES', '${PACKAGECONFIG_OPENGL}', d)}"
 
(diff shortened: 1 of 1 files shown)


recipes-graphics/wayland/weston_8.0.0.imx.bb
```diff
@@ -144,7 +144,7 @@ PACKAGECONFIG_OPENGL              = "opengl"
 PACKAGECONFIG_OPENGL_imxgpu2d     = ""
 PACKAGECONFIG_OPENGL_imxgpu3d     = "opengl"
 
-PACKAGECONFIG_remove = "wayland x11"
+PACKAGECONFIG_remove = "wayland x11 xwayland"
```
otavio, line 147: Does it fix a build error? or it is just for consistency?

zandrey: This actually fixes a build error, it has been reported on Yocto IRC. Original description is like that:

>  I'm trying to build the fsl-image-multimedia-full recipe, which relies on weston-8.0.0.imx.bb. That recipe is failing to build with an error about "ERROR: Problem encountered: xwayland requires xcursor which was not found. Or, you can use '-Dxwayland=false'." But I have xcursor, and xcursor-devel installed on my build system, and xcursor recipes exist within the includd meta-layers. Can anyone point me in the right direction here?

It turned out that when both X11 and wayland were disabled - xwayland cannot be enabled.

PR #505, opened 2020-09-30
imx-boot: allow for additional/different u-boot device-tree

MaxKrummenacher:
The current imx-boot recipe has hardcoded u-boot device-tree file names.

With this change the hardcoded value is overridden with whatever is set in
UBOOT_DTB_NAME.

### DIFF
--- a/recipes-bsp/imx-mkimage/imx-boot_1.0.bb
+++ b/recipes-bsp/imx-mkimage/imx-boot_1.0.bb
@@ -126,20 +126,20 @@ compile_mx8x() {
     fi
 }
 do_compile() {
-    compile_${SOC_FAMILY}
     # mkimage for i.MX8
     # Copy TEE binary to SoC target folder to mkimage
     if ${DEPLOY_OPTEE}; then
         cp ${DEPLOY_DIR_IMAGE}/tee.bin ${BOOT_STAGING}
     fi
     for target in ${IMXBOOT_TARGETS}; do
+        compile_${SOC_FAMILY}
         if [ "$target" = "flash_linux_m4_no_v2x" ]; then
            # Special target build for i.MX 8DXL with V2X off
            bbnote "building ${SOC_TARGET} - ${REV_OPTION} V2X=NO ${target}"
-           make SOC=${SOC_TARGET} ${REV_OPTION} V2X=NO  flash_linux_m4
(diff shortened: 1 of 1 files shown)


recipes-bsp/imx-mkimage/imx-boot_1.0.bb
```diff
@@ -126,20 +126,20 @@ compile_mx8x() {
...
 }
 do_compile() {
-    compile_${SOC_FAMILY}
     # mkimage for i.MX8
     # Copy TEE binary to SoC target folder to mkimage
     if ${DEPLOY_OPTEE}; then
         cp ${DEPLOY_DIR_IMAGE}/tee.bin ${BOOT_STAGING}
     fi
     for target in ${IMXBOOT_TARGETS}; do
+        compile_${SOC_FAMILY}
```
zandrey, line 135: @MaxKrummenacher From what I see this recipe does above is - it deploys `${UBOOT_DTB_NAME}` into `${BOOT_STAGING}` _only_ for `mx8m` machines.

Would your patch work for `mx8` and `mx8x`?

MaxKrummenacher: Yes.

The i.MX 8 and i.MX 8X families do not use a separate U-Boot device-tree binary.
For these two families a fix is not needed and the variable dtbs is not used in the soc.mak

P.S. Note that the patch does not deploy anything additionally. It only sets the variable dtbs in the call to make.

P.P.S. For i.MX 8 and i.MX 8X the same 'hardcoded to NXP evaluation hardware' pattern exists in other recipes, however there I'm not yet sure what fix to propose.

zandrey: @MaxKrummenacher @gibsson 

Thanks a lot for fixing this part. I took a look and indeed the `dtbs` parameter was present in the previous update, and has been dropped during the upgrade from `0.2` to `1.0`. Since I used the NXP EVK to verify this - I didn't note the missing parameter since the EVK DTBs are defaulted in the `imx-boot` itself.

I guess the version from @MaxKrummenacher is more preferable here over #507, since it would produce images for additional boot additional targets (e.g. `fspi`) so it could be taken instead.

PR #511, opened 2020-10-05
Morty

himanshugupta30062:
@kergoth morty
-Allow I2C addresses to be set for fxos8700 and fxas21002 and allow d… …
add function to get quaternion


otavio:
This is not acceptable; please use dunfell for new products.

PR #513, opened 2020-10-06
Jailhouse012

panfir:
Jailhouse upgrade to 0.12
Fix userspace missing tools

### DIFF
--- a/recipes-extended/jailhouse/jailhouse_0.12.bb
+++ b/recipes-extended/jailhouse/jailhouse_0.12.bb
@@ -13,8 +13,8 @@ LIC_FILES_CHKSUM = "file://COPYING;md5=9fa7f895f96bde2d47fd5b7d95b6ba4d \
                  file://driver/jailhouse.h;beginline=9;endline=36;md5=2825581c1666c44a17955dc574cfbfb3 \
 "
 
-SRCBRANCH = "lf-5.4.y"
-SRCREV = "82d5698945eafaffa58bab3b4d6da320bf56289c"
+SRCBRANCH = "imx_5.4.47_2.2.0"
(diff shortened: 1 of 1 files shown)


recipes-extended/jailhouse/jailhouse/0002-scripts-include.mk-multilib-Fix-hardcoded-lib-in-fir.patch
```diff
@@ -0,0 +1,28 @@
+From 5d81cd4eeffb5f8a9e73ee0faf374e72cec0d43c Mon Sep 17 00:00:00 2001
+From: Cristinel Panfir <cristinel.panfir@nxp.com>
+Date: Sat, 14 Mar 2020 01:22:05 +0200
+Subject: [PATCH] scripts/include.mk: multilib: Fix hardcoded lib in
+ firmwaredir
+
+Signed-off-by: Cristinel Panfir <cristinel.panfir@nxp.com>
```
otavio, line 7: This patch is not need; pass `firmwaredir=$(base_libdir)/firmware`  in `EXTRA_OEMAKE` and it should do the trick.

otavio (changes requested):

recipes-extended/jailhouse/jailhouse/0002-scripts-include.mk-multilib-Fix-hardcoded-lib-in-fir.patch
panfir: Yes, it make sens, I'll discard that patch for now.
I'll check later the passing via `EXTRA_OEMAKE` and if there needs something else added in recipe

otavio (approved):

PR #515, opened 2020-10-07
imx-boot: add back LD_LIBRARY_PATH exporting

liuming50:
It was dropped by commit 65100fad:
[ imx-mkimage: upgrade to version 1.0 ]

after that, the following error was observed again:
| ./mkimage_uboot -E -p 0x3000 -f u-boot.its u-boot.itb
| ./mkimage_uboot: error while loading shared libraries: libssl.so.1.1: cannot open shared object file: No such file or directory

### DIFF
--- a/recipes-bsp/imx-mkimage/imx-boot_1.0.bb
+++ b/recipes-bsp/imx-mkimage/imx-boot_1.0.bb
@@ -126,6 +126,9 @@ compile_mx8x() {
     fi
 }
 do_compile() {
+    # mkimage_uboot requires libssl.so.1.1 from ${STAGING_LIBDIR_NATIVE}
+    export LD_LIBRARY_PATH=${STAGING_LIBDIR_NATIVE}:$LD_LIBRARY_PATH
+
     # mkimage for i.MX8
     # Copy TEE binary to SoC target folder to mkimage
     if ${DEPLOY_OPTEE}; then


zandrey:
@liuming50 I do not see this build issue on the `imx8mmevk` machine...

Can you share which machine you're building against?

I guess this can also be solved via:

```python
DEPENDS = "openssl-native"
```

liuming50:
I reproduced it on a verdin-imx8mm machine.

DEPENDS = "openssl-native" probably could not fix the problem since mkimage_uboot is being executed in ${BOOT_STAGING}/mkimage_uboot, so it would not look for libraries in ${STAGING_LIBDIR_NATIVE}

zandrey:
I'm wondering whether your `u-boot` recipe still installs the `mkimage_uboot` tool in this case?

Starting from `5.4.24-2.1.0` release it is not necessary to be done anymore, as it would be deployed by `imx-boot` itself. You can see the commit [ac7476ea5aef52eb5fda4ce00756e7e0ddfe1174](https://github.com/Freescale/meta-freescale-3rdparty/commit/ac7476ea5aef52eb5fda4ce00756e7e0ddfe1174) in _meta-freescale-3rdparty_ where this has been dropped for _U-boot_ from Boundary Devices.

In the NXP BSP there is also a [note in the imx-boot recipe about it](https://source.codeaurora.org/external/imx/meta-imx/tree/meta-bsp/recipes-bsp/imx-mkimage/imx-boot_1.0.bb?h=zeus-5.4.24-2.1.0#n112), this was brought in commit [226ab440dd](https://source.codeaurora.org/external/imx/meta-imx/commit/?h=zeus-5.4.24-2.1.0&id=226ab440dd1080497f5a29f417a895e4060b683f) in NXP yocto tree.

I'd tend to lean to the fact that this not an issue with the implementation in this recipe, rather a dependency that has been pulled in incorrect.

liuming50:
@zandrey Arha, maybe I was using mkimage deployed from u-boot from other meta layers, will double-check that.

Close this MR, it might not be needed, thanks for the comments.

PR #537, opened 2020-10-27
imx-seco: Upgrade to 3.7.1

panfir:
Simplify the compatibility rule, all mx8 except mx8m supports SECO.
Otherwise, some machines on the list are not upstream.

Signed-off-by: Cristinel Panfir <cristinel.panfir@nxp.com>

### DIFF
--- a/recipes-bsp/imx-seco/imx-seco_3.7.1.bb
+++ b/recipes-bsp/imx-seco/imx-seco_3.7.1.bb
@@ -4,14 +4,14 @@ SUMMARY = "NXP i.MX SECO firmware"
 DESCRIPTION = "NXP IMX SECO firmware"
 SECTION = "base"
 LICENSE = "Proprietary"
-LIC_FILES_CHKSUM = "file://COPYING;md5=228c72f2a91452b8a03c4cab30f30ef9"
+LIC_FILES_CHKSUM = "file://COPYING;md5=983e4c77621568488dd902b27e0c2143"
 
 inherit fsl-eula-unpack use-imx-security-controller-firmware deploy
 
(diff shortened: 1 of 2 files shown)


recipes-bsp/imx-seco/imx-seco_3.7.1.bb
```diff
@@ -23,4 +23,5 @@ do_deploy () {
     install -m 0644 ${S}/firmware/seco/${SECO_FIRMWARE_NAME} ${DEPLOYDIR}
 }
 
-COMPATIBLE_MACHINE = "(mx8qm|mx8qxp|mx8qxpc0|mx8phantomdxl|mx8dxl)"
```
otavio, line 26: This is bad for tracking of what is compatible, please keep the old way.

otavio (changes requested):

recipes-bsp/imx-seco/imx-seco_3.7.1.bb
panfir: ok, that's fine
note that `imx-seco-libs_git` is using already the simplified rule

panfir: back to list rule ..I'll add also the rest of machines

otavio (approved):

recipes-bsp/imx-seco/imx-seco_3.7.1.bb
otavio: Good; please replace the simplified rule, so we keep it consistent.

otavio (approved):

PR #541, opened 2020-11-01
Master enhancements

varigigi:
general enhancements for iMX6, iMX7 and iMX8 machines

### DIFF
--- a/conf/machine/include/imx-base.inc
+++ b/conf/machine/include/imx-base.inc
@@ -281,6 +281,15 @@ MACHINE_GSTREAMER_1_0_PLUGIN_mx7d   ?= "gstreamer1.0-plugins-imx-meta"
 MACHINE_GSTREAMER_1_0_PLUGIN_mx8mm  ?= "imx-gst1.0-plugin"
 MACHINE_GSTREAMER_1_0_PLUGIN_mx8mn  ?= "imx-gst1.0-plugin"
 MACHINE_GSTREAMER_1_0_PLUGIN_mx8mp  ?= "imx-gst1.0-plugin"
+MACHINE_GSTREAMER_1_0_PLUGIN_mx8mq  ?= "imx-gst1.0-plugin"
+MACHINE_GSTREAMER_1_0_PLUGIN_mx8qm  ?= "imx-gst1.0-plugin"
+MACHINE_GSTREAMER_1_0_PLUGIN_mx8qxp ?= "imx-gst1.0-plugin"
+
+PREFERRED_VERSION_gstreamer1.0_mx8 ?= "1.16.imx"
+PREFERRED_VERSION_gstreamer1.0-plugins-base_mx8 ?= "1.16.imx"
+PREFERRED_VERSION_gstreamer1.0-plugins-good_mx8 ?= "1.16.imx"
+PREFERRED_VERSION_gstreamer1.0-plugins-bad_mx8 ?= "1.16.imx"
+PREFERRED_VERSION_gstreamer1.0-libav_mx8 ?= "1.16.0"
 
 # Determines if the SoC has support for Vivante kernel driver
 SOC_HAS_VIVANTE_KERNEL_DRIVER_SUPPORT        = "0"
--- a/recipes-graphics/wayland/weston-init.bbappend
+++ b/recipes-graphics/wayland/weston-init.bbappend
@@ -15,6 +15,12 @@ SRC_URI_append_mx6sl = " file://weston.config"
 INI_UNCOMMENT_ASSIGNMENTS_append_imx = " \
     ${@bb.utils.contains('DISTRO_FEATURES', 'x11 wayland', 'xwayland=true', '', d)} \
 "
+INI_UNCOMMENT_ASSIGNMENTS_append_mx6dl = " \
+    use-g2d=1 \
+"
+INI_UNCOMMENT_ASSIGNMENTS_append_mx6q = " \
+    use-g2d=1 \
+"
 INI_UNCOMMENT_ASSIGNMENTS_append_mx7ulp = " \
     use-g2d=1 \
 "
--- a/dynamic-layers/qt5-layer/recipes-qt/qt5/qtwayland_%.bbappend
+++ b/dynamic-layers/qt5-layer/recipes-qt/qt5/qtwayland_%.bbappend
@@ -1,5 +1,8 @@
 # etnaviv mesa does not have glx
 PACKAGECONFIG_remove_use-mainline-bsp = "xcomposite-glx"
 
+PACKAGECONFIG_remove_mx6 = "xcomposite-egl xcomposite-glx"
+PACKAGECONFIG_remove_mx7 = "xcomposite-egl xcomposite-glx"
+
 # i.MX8 does never provide native x11, so required dependencies are not met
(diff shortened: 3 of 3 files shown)


recipes-graphics/wayland/weston-init.bbappend
```diff
@@ -15,6 +15,12 @@ SRC_URI_append_mx6sl = " file://weston.config"
 INI_UNCOMMENT_ASSIGNMENTS_append_imx = " \
     ${@bb.utils.contains('DISTRO_FEATURES', 'x11 wayland', 'xwayland=true', '', d)} \
 "
+INI_UNCOMMENT_ASSIGNMENTS_append_mx6dl = " \
+    use-g2d=1 \
+"
+INI_UNCOMMENT_ASSIGNMENTS_append_mx6q = " \
+    use-g2d=1 \
+"
```
otavio, line 23: They do support it but I wonder which we ought to use by default. @thochstein can you check internally?

conf/machine/include/imx-base.inc
```diff
@@ -281,6 +281,15 @@ MACHINE_GSTREAMER_1_0_PLUGIN_mx7d   ?= "gstreamer1.0-plugins-imx-meta"
...
 MACHINE_GSTREAMER_1_0_PLUGIN_mx8mp  ?= "imx-gst1.0-plugin"
+MACHINE_GSTREAMER_1_0_PLUGIN_mx8mq  ?= "imx-gst1.0-plugin"
+MACHINE_GSTREAMER_1_0_PLUGIN_mx8qm  ?= "imx-gst1.0-plugin"
+MACHINE_GSTREAMER_1_0_PLUGIN_mx8qxp ?= "imx-gst1.0-plugin"
+
+PREFERRED_VERSION_gstreamer1.0_mx8 ?= "1.16.imx"
+PREFERRED_VERSION_gstreamer1.0-plugins-base_mx8 ?= "1.16.imx"
+PREFERRED_VERSION_gstreamer1.0-plugins-good_mx8 ?= "1.16.imx"
+PREFERRED_VERSION_gstreamer1.0-plugins-bad_mx8 ?= "1.16.imx"
+PREFERRED_VERSION_gstreamer1.0-libav_mx8 ?= "1.16.0"
```
otavio, line 292: The libav does not have a fork, so we can let it float right?

otavio (changes requested):

recipes-graphics/wayland/weston-init.bbappend
thochstein: Prabhu says this is a good idea.

otavio: That's good; thanks for the feedback.

conf/machine/include/imx-base.inc
```diff
@@ -281,6 +281,15 @@ MACHINE_GSTREAMER_1_0_PLUGIN_mx7d   ?= "gstreamer1.0-plugins-imx-meta"
...
 MACHINE_GSTREAMER_1_0_PLUGIN_mx8mn  ?= "imx-gst1.0-plugin"
 MACHINE_GSTREAMER_1_0_PLUGIN_mx8mp  ?= "imx-gst1.0-plugin"
+MACHINE_GSTREAMER_1_0_PLUGIN_mx8mq  ?= "imx-gst1.0-plugin"
+MACHINE_GSTREAMER_1_0_PLUGIN_mx8qm  ?= "imx-gst1.0-plugin"
+MACHINE_GSTREAMER_1_0_PLUGIN_mx8qxp ?= "imx-gst1.0-plugin"
+
+PREFERRED_VERSION_gstreamer1.0_mx8 ?= "1.16.imx"
+PREFERRED_VERSION_gstreamer1.0-plugins-base_mx8 ?= "1.16.imx"
+PREFERRED_VERSION_gstreamer1.0-plugins-good_mx8 ?= "1.16.imx"
+PREFERRED_VERSION_gstreamer1.0-plugins-bad_mx8 ?= "1.16.imx"
```
zandrey, line 291: `MACHINE_GSTREAMER_1_0_PLUGIN` and `PREFERRED_VERSION_gstreamer1.0` are also defined in [meta-freescale-distro/conf/distro/include/fsl-base.inc](https://github.com/Freescale/meta-freescale-distro/blob/063a082a241772b7675daa258822954554160b65/conf/distro/include/fsl-base.inc#L60)

Since I do not see a use case where someone would pull the _meta-freescale-distro_ without pulling _meta-freescale_ - would it make sense to move all defines in here and clean them up in the disto layer? They seem to be scattered now between 2 layers, and defined differently for mx6/mx7 machines.

otavio: Those serve different purposes:

- `imx-base.inc` aims to provide the BSP default. For all SoCs but the i.MX8 we default to `gstreamer-1.0-imx` which supports NXP BSP and has been developed by the community. It does not yet offer i.MX8 support and @dv1 is working to solve this gap.

- `fsl-*` distros are to provide an NXP BSP very close to their GA releases but using up to date underlying Operating System.

In future the i.MX8 SoCs are also going to use the `gstreamer-1.0-imx` provider and then the `imx-gst1.0-plugin` will be used only on `fsl-*` distros.

zandrey: OK, got it! This actually serves a "reverse logic" where if the _-distro_ is not pulled in - it defaults to what is inside this conf files, otherwise - it opts-in to choose a proper config based on the BSP flavor.

Good to know, thanks a lot!

Then I guess proper `MACHINE_GSTREAMER_1_0_PLUGIN` defines for `mx8mm/mx8mn/mx8mp` are missing in the _-distro_ layer...

conf/machine/include/imx-base.inc
varigigi: meta-freescale provides gstreamer1.0-libav_1.16.0.bb to be compliant with gstreamer 1.16.imx
poky provides gstreamer1.0-libav_1.16.2.bb to compliant with the stock gstreamer 1.16.2
libav should be set according what choosen for Gstreamer, am I wrong?

otavio: You're right. In this case, we ought to rename the recipe, so it is clearly a  `.imx` version as the others.

conf/machine/include/imx-base.inc
otavio: @zandrey can you take care of the distro?

conf/machine/include/imx-base.inc
varigigi: It's not a custom NXP/imx version, it's just the stock 1.16.0 version, no longer available in poky due to updates to 1.16.2.
I'm wondering if it make sense remove it completely libav recipes an references
...

conf/machine/include/imx-base.inc
zandrey: > @zandrey can you take care of the distro?

Sorry I didn't give and update here before, PR [41](https://github.com/Freescale/meta-freescale-distro/pull/41) submitted against _meta-freescale-distro_.

otavio: I merged it already.

conf/machine/include/imx-base.inc
otavio: @varigigi I know it is not a real fork but we could use the specific version to avoid confusion. What do you think?

PR #551, opened 2020-11-09
fsl-kernel-localversion.bbclass: add preconfigure after do_kernel_met…

liuming50:
…adata

do_preconfigure task should run after do_kernel_metadata, or else there
might be no ${WORKDIR}/defconfig present when KBUILD_DEFCONFIG is being
set rather than defconfig added in SRC_URI.

### DIFF
--- a/classes/fsl-kernel-localversion.bbclass
+++ b/classes/fsl-kernel-localversion.bbclass
@@ -40,4 +40,4 @@ do_preconfigure() {
 		printf "%s%s" +g $head > ${S}/.scmversion
 	fi
 }
-addtask preconfigure before do_configure after do_unpack do_patch
+addtask preconfigure before do_configure after do_unpack do_patch do_kernel_metadata


classes/fsl-kernel-localversion.bbclass
```diff
@@ -40,4 +40,4 @@ do_preconfigure() {
 		printf "%s%s" +g $head > ${S}/.scmversion
 	fi
 }
-addtask preconfigure before do_configure after do_unpack do_patch
+addtask preconfigure before do_configure after do_kernel_metadata
```
otavio, line 43: I believe you cannot remove the previous tasks as `kernel_metadata` is a `linux-yocto` specific task and we need to keep backward compatibility.

otavio (changes requested):

classes/fsl-kernel-localversion.bbclass
liuming50: @otavio thanks for the comment

Have modified that as you suggested in patch set V2.

otavio (approved):

PR #556, opened 2020-11-10
gstreamer1.0-plugins-base: add additional patches

zdobersek:
Add a pair of patches that's to be applied over the iMX flavor of the
gstreamer1.0-plugins-base codebase.

First patch comes from upstream, and it fixes glupload behavior by avoiding
mapping the GL buffers into CPU memory. It's only applicable to the current
MM_04.05.05_2005_L5.4.24 branch that is based on the upstream 1.16.1 release,
since the patch itself is already present in 1.16.2.
https://gitlab.freedesktop.org/gstreamer/gst-plugins-base/-/merge_requests/444

Second patch is a fix for Meson builds that enable the viv-fb GL winsys.
When enabled, the GL plugins library should also link against libg2d since
the viv-fb code is using g2d allocators.

Signed-off-by: Zan Dobersek <zdobersek@igalia.com>

### DIFF
--- a/recipes-multimedia/gstreamer/gstreamer1.0-plugins-base/0001-glupload-Add-VideoMetas-and-GLSyncMeta-to-the-raw-up.patch
+++ b/recipes-multimedia/gstreamer/gstreamer1.0-plugins-base/0001-glupload-Add-VideoMetas-and-GLSyncMeta-to-the-raw-up.patch
@@ -1,0 +1,87 @@
+From 8d32de090554cf29fe359f83aa46000ba658a693 Mon Sep 17 00:00:00 2001
+From: Thibault Saunier <tsaunier@igalia.com>
+Date: Fri, 27 Sep 2019 11:10:43 -0300
+Subject: [PATCH] glupload: Add VideoMetas and GLSyncMeta to the raw uploaded
+ buffers
+
+This is done by reusing `gst_gl_memory_setup_buffer` avoiding to
+duplicate code.
+
+Without a VideoMeta, mapping those buffers lead to GstBuffer mapping the
+buffer in system memory even when specifying the GL flags (through the
+buffer merging mechanism) making the result totally broken.
+
+Part-of: https://gitlab.freedesktop.org/gstreamer/gst-plugins-base/-/merge_requests/444
+Upstream-Status: Backport [8d32de090554cf29fe359f83aa46000ba658a693]
+
+Signed-off-by: Zan Dobersek <zdobersek@igalia.com>
+---
+ gst-libs/gst/gl/gstglupload.c  | 32 +++++++++++---------------------
+ tests/check/libs/gstglupload.c |  2 ++
+ 2 files changed, 13 insertions(+), 21 deletions(-)
+
+diff --git a/gst-libs/gst/gl/gstglupload.c b/gst-libs/gst/gl/gstglupload.c
+index bc6db7e0a..67ec4f4e9 100644
+--- a/gst-libs/gst/gl/gstglupload.c
++++ b/gst-libs/gst/gl/gstglupload.c
+@@ -1293,33 +1293,23 @@ _raw_data_upload_perform (gpointer impl, GstBuffer * buffer,
+       (raw->upload->context));
+ 
+   /* FIXME Use a buffer pool to cache the generated textures */
+-  /* FIXME: multiview support with separated left/right frames? */
+   *outbuf = gst_buffer_new ();
+-  for (i = 0; i < n_mem; i++) {
+-    GstGLBaseMemory *tex;
+-
+-    raw->params->parent.wrapped_data = raw->in_frame->frame.data[i];
+-    raw->params->plane = i;
+-    raw->params->tex_format =
+-        gst_gl_format_from_video_info (raw->upload->context, in_info, i);
+-
+-    tex =
+-        gst_gl_base_memory_alloc (allocator,
+-        (GstGLAllocationParams *) raw->params);
+-    if (!tex) {
+-      gst_buffer_unref (*outbuf);
+-      *outbuf = NULL;
+-      GST_ERROR_OBJECT (raw->upload, "Failed to allocate wrapped texture");
+-      return GST_GL_UPLOAD_ERROR;
+-    }
++  raw->params->parent.context = raw->upload->context;
++  if (gst_gl_memory_setup_buffer ((GstGLMemoryAllocator *) allocator, *outbuf,
++          raw->params, NULL, raw->in_frame->frame.data, n_mem)) {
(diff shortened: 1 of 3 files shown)


recipes-multimedia/gstreamer/gstreamer1.0-plugins-base/0001-glupload-Add-VideoMetas-and-GLSyncMeta-to-the-raw-up.patch
```diff
@@ -0,0 +1,82 @@
...
+Date: Fri, 27 Sep 2019 11:10:43 -0300
+Subject: [PATCH] glupload: Add VideoMetas and GLSyncMeta to the raw uploaded
+ buffers
+
+This is done by reusing `gst_gl_memory_setup_buffer` avoiding to
+duplicate code.
+
+Without a VideoMeta, mapping those buffers lead to GstBuffer mapping the
+buffer in system memory even when specifying the GL flags (through the
+buffer merging mechanism) making the result totally broken.
```
otavio, line 12: Please add the Upstream-Status header on both patches.

otavio (changes requested):
Hello, please see the minor comment I did below.

Also, did you verify if a newer NXP release has it? We could upgrade it as well.

zdobersek:
> Also, did you verify if a newer NXP release has it? We could upgrade it as well.

MM_04.05.06_2008_L5.4.47, the newer gst-plugins-base branch, is based on the 1.16.2 upstream release, which already has the gstglupload patch. The second patch doesn't affect upstream, but I am trying to get it into the gst-plugins-base repository.

zdobersek:
Amended the commit, updating the patch files with the Upstream-Status and Signed-off-by headers.

recipes-multimedia/gstreamer/gstreamer1.0-plugins-base/0001-glupload-Add-VideoMetas-and-GLSyncMeta-to-the-raw-up.patch
```diff
@@ -0,0 +1,86 @@
...
+
+Upstream-Status: Backport
```
zandrey, line 14: Can you please put a link to or SHA of the original commit? Would be easier to track when `1.16.2` is integrated.

I believe the link is the one you put in the commit message (https://gitlab.freedesktop.org/gstreamer/gst-plugins-base/-/merge_requests/444), would be useful to have it here as well.

recipes-multimedia/gstreamer/gstreamer1.0-plugins-base/0001-meson-viv-fb-code-must-link-against-libg2d.patch
```diff
@@ -0,0 +1,44 @@
+From 8153e86dac04076fb80fc272c8374d4ba019fa3c Mon Sep 17 00:00:00 2001
+From: Zan Dobersek <zdobersek@igalia.com>
+Date: Wed, 4 Nov 2020 14:02:10 +0100
+Subject: [PATCH] meson: viv-fb code must link against libg2d
+
+Find the libg2d library and link against it, if necessary, avoiding
+linking errors for couple of libg2d symbols.
+
+Upstream-Status: Pending
```
zandrey, line 9: Please also put a link to your MR here, that way it would be visible when it gets merged in.

zandrey (changes requested):
There are missing references in the `Upstream-Status`, can you please fill them in?

recipes-multimedia/gstreamer/gstreamer1.0-plugins-base/0001-meson-viv-fb-code-must-link-against-libg2d.patch
zdobersek: This change is against the NXP fork of gst-plugins-base. I don't know of any public issue trackers for that repository, so I mailed the maintainers for further information.

recipes-multimedia/gstreamer/gstreamer1.0-plugins-base/0001-glupload-Add-VideoMetas-and-GLSyncMeta-to-the-raw-up.patch
zdobersek: Done.

zdobersek:
> We could upgrade it as well.

I suppose the whole GStreamer stack could be upgraded to the newer branches, but I don't know if that's an off-the-cuff change that anyone can propose or whether you need some more coordination.

otavio:
@zdobersek please send the update; this help us a lot and allow us to get it in sooner :)

zandrey:
> 
> 
> > We could upgrade it as well.
> 
> I suppose the whole GStreamer stack could be upgraded to the newer branches, but I don't know if that's an off-the-cuff change that anyone can propose or whether you need some more coordination.

Like @otavio just wrote: if you can bump those up - that would be great, any help is appreciated here. 👍 

I did this in the past, some pieces might be tricky but generally doable.

-- andrey

recipes-multimedia/gstreamer/gstreamer1.0-plugins-base/0001-meson-viv-fb-code-must-link-against-libg2d.patch
zandrey: > This change is against the NXP fork of gst-plugins-base.

So this is rather a change in NXP-specific part of `gstreamer1.0-plugins-base`? In this case the `Upstream-Status:` should be set to `Inappropriate [NXP-specific]`.

> I don't know of any public issue trackers for that repository, so I mailed the maintainers for further information.

Have you posted it to some list or directly to maintainer? If it is the later, then maybe Cc'ing _meta-freescale@yoctoproject.org_ would also be helpful to track the status of that patch.

zdobersek: I'll use `Inappropriate [i.MX specific]`, as already used in a few patch files.

zdobersek:
I'll try and do the update once I have a working-reference environment, and some leeway to do it. Along with GStreamer, I imagine there's also the kernel and a range of other recipes that would need updating.

zdobersek:
Force-pushed the amended patch.

otavio (approved):

zandrey:
> Along with GStreamer, I imagine there's also the kernel and a range of other recipes that would need updating.

Kernel recipe containing 5.4.47_2.2.0 is already in `master-next`, modules I have to see about.

PR #557, opened 2020-11-11
fsl-kernel-localversion.bbclass: add preconfigure after do_kernel_met…

liuming50:
…adata

do_preconfigure task should run after do_kernel_metadata, or else there
might be no ${WORKDIR}/defconfig present when KBUILD_DEFCONFIG is being
set rather than defconfig added in SRC_URI.


liuming50:
Hi, may we cherry pick this to dunfell? The issue is breaking our CI/CD builds. Thanks.

otavio:
Ok but please use `git cherry-pick -x` to backport it.

liuming50:
@otavio thanks for the comment, done in patch set V2.

PR #566, opened 2020-11-23
Add i.MX vulkan driver to image explicitly

thochstein:
The i.MX vulkan driver libvulkan-imx is dynamically loaded, so bitbake won't add it
to the image implicitly, and a missing library error at runtime is the result.

Add the driver explicitly.

### DIFF
--- a/recipes-graphics/vulkan/vulkan-loader_%.bbappend
+++ b/recipes-graphics/vulkan/vulkan-loader_%.bbappend
@@ -1,0 +1,3 @@
+# The i.MX implementation is dynamically loaded, so it requires an
+# explicit runtime dependency.
+RRECOMMENDS_${PN}_append_imxgpu = " libvulkan-imx"


conf/machine/include/imx-base.inc
```diff
@@ -303,6 +303,12 @@ PREFERRED_PROVIDER_virtual/libgles2_imxgpu3d ?= "imx-gpu-viv"
 PREFERRED_PROVIDER_virtual/libg2d            ?= "imx-gpu-g2d"
 PREFERRED_PROVIDER_virtual/libg2d_imxdpu     ?= "imx-dpu-g2d"
 
+# Vulkan libraries are dynamically loaded and so must be added to the
+# rootfs explicitly.
+IMX_GPU_VIV_VULKAN     = ""
+IMX_GPU_VIV_VULKAN_mx8 = "libvulkan-imx"
+MACHINE_EXTRA_RRECOMMENDS_append_imxgpu = " ${@bb.utils.contains('DISTRO_FEATURES', 'vulkan', '${IMX_GPU_VIV_VULKAN}', '', d)}"
```
otavio, line 310: Which application did you found it was not working? We cannot do this, in this way as it ends with adding it to all machines even if the image does not use the graphical stack.

otavio (changes requested):

conf/machine/include/imx-base.inc
thochstein: vulkan-cts

otavio: It depends. Please check the `ldd` output of the binary.

thochstein:
```
root@lsv07069:/opt/work/upstream/fsl-xwayland/tmp/work/cortexa53-crypto-mx8m-fsl-linux/vulkan-cts/1.2.3.2-r0/git# aarch64-fsl-linux-objdump -p /opt/work/upstream/fsl-xwayland/tmp/deploy/rpm/cortexa53_crypto_mx8m/test/usr/lib/vulkan-cts/deqp-vk | grep NEEDED
  NEEDED               libpthread.so.0
  NEEDED               libdl.so.2
  NEEDED               libpng16.so.16
  NEEDED               libGLESv2.so.2
  NEEDED               libEGL.so.1
  NEEDED               libstdc++.so.6
  NEEDED               libm.so.6
  NEEDED               libgcc_s.so.1
  NEEDED               libc.so.6
  NEEDED               ld-linux-aarch64.so.1 
```

otavio:
@thochstein I did a deeper look at this, and this does not seem right. It seems, at least for now, we ought to have it manually added to the required packages, so we avoid adding it to all images or in packages that don't use it.

otavio (approved):

PR #573, opened 2020-12-03
Fix u boot imx tools issues

liuming50:


recipes-bsp/u-boot/u-boot-imx-tools_2020.04.bb
```diff
@@ -6,5 +6,4 @@ PROVIDES_append_class-native = " u-boot-tools-native"
 PROVIDES_append_class-nativesdk = " nativesdk-u-boot-tools"
 
 PACKAGE_ARCH = "${MACHINE_ARCH}"
-COMPATIBLE_MACHINE = "(mx6|mx7|mx8)"
 COMPATIBLE_MACHINE_class-target = "(mx6|mx7|mx8)"
```
otavio, line 9: Well, not really. It shouldn't infect other layers.

What issue this triggers for you?

conf/machine/include/imx-base.inc
```diff
@@ -23,11 +23,11 @@ IMX_DEFAULT_BSP_mx5 ?= "mainline"
...
 PREFERRED_PROVIDER_u-boot ??= "${IMX_DEFAULT_BOOTLOADER}"
-PREFERRED_PROVIDER_u-boot-tools-native ??= "${IMX_DEFAULT_BOOTLOADER}-tools-native"
-PREFERRED_PROVIDER_nativesdk-u-boot-tools ??= "nativesdk-${IMX_DEFAULT_BOOTLOADER}-tools"
-PREFERRED_PROVIDER_u-boot-mkimage-native ??= "${IMX_DEFAULT_BOOTLOADER}-tools-native"
-PREFERRED_PROVIDER_nativesdk-u-boot-mkimage ??= "nativesdk-${IMX_DEFAULT_BOOTLOADER}-tools"
 PREFERRED_PROVIDER_virtual/bootloader ??= "${IMX_DEFAULT_BOOTLOADER}"
+PREFERRED_PROVIDER_u-boot-tools-native ??= "u-boot-imx-tools-native"
+PREFERRED_PROVIDER_nativesdk-u-boot-tools ??= "nativesdk-u-boot-imx-tools"
+PREFERRED_PROVIDER_u-boot-mkimage-native ??= "u-boot-imx-tools-native"
+PREFERRED_PROVIDER_nativesdk-u-boot-mkimage ??= "nativesdk-u-boot-imx-tools"
```
otavio, line 30: We control the u-boot depending on the BSP we are using, so the change you are proposing does not improve it. Also, this might break customers' internal BSPs. Could you share the motivation for it?

otavio (changes requested):

recipes-bsp/u-boot/u-boot-imx-tools_2020.04.bb
liuming50: Without dropping that line, how a end user could run "bitbake u-boot-imx-tools-native", or how could they set DEPENDS = "u-boot-tools-native" while u-boot-imx-tools-native is the preferred provider of u-boot-tools-native? Since as a native recipe, it could not comply only to imx6,imx7,imx8 machines, it needs to comply with x86-64 machine.

conf/machine/include/imx-base.inc
liuming50: Just simply because IMX_DEFAULT_BOOTLOADER could be u-boot-fslc, in which case, ${IMX_DEFAULT_BOOTLOADER}-tools-native would make no sense because there is not a u-boot-fslc-tools-native recipe at all.

otavio: This is right, but not necessarily the unique option as we have customers who do have it.

liuming50: Then maybe they should handle PREFERRED_PROVIDER in their own meta layers? Otherwise the customers dont have u-boot-fslc-tools recipes would be impacted? The truth is there is not a u-boot-fslc-tools recipe in meta-freescale or meta-freescale-3rdparty, then it should not be referred.

liuming50: By the way, without this patch, the end users will get this error:
```
NOTE: Multiple providers are available for u-boot-tools-native (u-boot-tools-native, u-boot-imx-tools-native)
Consider defining a PREFERRED_PROVIDER entry to match u-boot-tools-native
```
that's because PREFERRED_PROVIDER_u-boot-tools-native set to an invalid recipe so it does not take effect, means no PREFERRED_PROVIDER_u-boot-tools-native is being set.

liuming50: I mean for imx6/imx7 machines.

otavio:
@liuming50 I reworked the recipe, and I believe it covers the error case you found. We'll add a tools recipe for `u-boot-fslc` next week, so it also provides it and becomes a valid provider.

liuming50:
@otavio thanks for the update, yes, adding u-boot-fslc-tools is a good idea. Then closing this.

PR #580, opened 2020-12-13
imx-base.inc: add imx-boot as a dependency for mx8m machines

zandrey:
`imx-boot` is required to be provided as a dependency for `mx8m` machines in order to provide a boot container, which WIC is expecting to include into the final image.

Since the only dependency for WIC present in the layer now is targeting mx8 machine and _MACHINEOVERRIDES_EXTENDER_FILTER_OUT_ removes it when `use-mainline-bsp` is set - it leaves no 'imx-boot' container in the path of creating WIC file, and build for mx8m machines fails.

Append _WKS_FILE_DEPENDS_ to include `imx-boot` for `mx8m`, as it is required to be present regardless of the BSP flavor chosen.

Fixes: e2589ccb ("imx-base.inc: Avoid adding 'imx-boot' as dependency for all SoCs")

### DIFF
--- a/conf/machine/include/imx-base.inc
+++ b/conf/machine/include/imx-base.inc
@@ -366,6 +366,7 @@ WKS_FILE_DEPENDS ?= " \
 "
 
 WKS_FILE_DEPENDS_append_mx8 = " imx-boot"
+WKS_FILE_DEPENDS_append_mx8m = " imx-boot"
 
 SOC_DEFAULT_WKS_FILE ?= "imx-uboot-bootpart.wks.in"
 SOC_DEFAULT_WKS_FILE_mx8m ?= "imx-imx-boot-bootpart.wks.in"


conf/machine/include/imx-base.inc
```diff
@@ -366,6 +366,7 @@ WKS_FILE_DEPENDS ?= " \
 "
 
 WKS_FILE_DEPENDS_append_mx8 = " imx-boot"
+WKS_FILE_DEPENDS_append_mx8m = " imx-boot"
```
otavio, line 369: Can we drop line 368? why both?

zandrey: Above definition is kept for `mx8qm` and `mx8x` machines, which are using _imx-boot_ but do not have `mx8m` defined.

Since they are only targeted to build from NXP BSP - removing above line would remove the _imx-boot_ from the WIC path for them.

PR #588, opened 2020-12-16
qoriq optee-os/optee-test/optee-client: upgrade to 3.8.0

ting-liu:


recipes-security/optee/optee-os-qoriq/0001-scripts-sign_encrypt.py-Correct-the-Crypto-module-na.patch
```diff
@@ -0,0 +1,39 @@
...
++    from Crypto.Hash import SHA256
++    from Crypto.PublicKey import RSA
+     import base64
+     import logging
+     import os
+@@ -169,7 +169,7 @@ def main():
+     shdr_version = struct.pack('<I', hdr_version)
+ 
+     if args.enc_key:
+-        from Cryptodome.Cipher import AES
```
otavio, line 32: Why not use the cryptodome and avoid patching the code?

otavio (changes requested):

PR #599, opened 2020-12-17
optee-test_3.7.0.imx: fix optee-test build

petegriffin:
Currently nothing is built for optee-test which means
do_install() fails copying xtest binary.

Align the do_compile() with upstream meta-arm optee-test recipe.
With this patch the build then completes OK.

Fixes: a66dc98
Signed-off-by: Peter Griffin <peter.griffin@linaro.org>
(cherry picked from commit 840176462b2ae175bc06322f32c21d5b117b673e)


petegriffin:
@otavio fyi - as requested

PR #604, opened 2020-12-21
gstreamer1.0-plugins-bad: Let processors with pxp chose the correct Opengl API

zakariazh:
Hello, 
The processors without gpu (imx6ul imx6ull) support only opengl-es, gstreamer1.0-plugins-bad must also be configured with this API instead of the standard gl api. Otherwise the compilation does not pass and generate errors.

Thank you

### DIFF
--- a/recipes-multimedia/gstreamer/gstreamer1.0-plugins-bad_1.16.imx.bb
+++ b/recipes-multimedia/gstreamer/gstreamer1.0-plugins-bad_1.16.imx.bb
@@ -36,12 +36,15 @@ DEPENDS += "gstreamer1.0-plugins-base"
 
 inherit gobject-introspection
 
+PACKAGECONFIG_GL ?= "${@bb.utils.contains('DISTRO_FEATURES', 'opengl', 'gl', '', d)}"
+PACKAGECONFIG_GL_imxpxp = "gles2"
+
 PACKAGECONFIG ??= " \
     ${GSTREAMER_ORC} \
     ${@bb.utils.contains('DISTRO_FEATURES', 'bluetooth', 'bluez', '', d)} \
     ${@bb.utils.filter('DISTRO_FEATURES', 'directfb vulkan', d)} \
     ${@bb.utils.contains('DISTRO_FEATURES', 'wayland', 'wayland', '', d)} \
-    ${@bb.utils.contains('DISTRO_FEATURES', 'opengl', 'gl', '', d)} \
+    ${PACKAGECONFIG_GL} \
     bz2 closedcaption curl dash dtls hls rsvg sbc smoothstreaming sndfile \
     ttml uvch264 webp \
 "
@@ -63,6 +66,7 @@ PACKAGECONFIG[hls]             = "-Dhls=enabled -Dhls-crypto=nettle,-Dhls=disabl
 # to -base yet. They depend on the gstgl library in -base, so we do
 # not add GL dependencies here, since these are taken care of in -base.
 PACKAGECONFIG[gl]              = "-Dgl=enabled,-Dgl=disabled,"
+PACKAGECONFIG[gles2]           = ",,virtual/libgles2"
 PACKAGECONFIG[kms]             = "-Dkms=enabled,-Dkms=disabled,libdrm"
 PACKAGECONFIG[libde265]        = "-Dlibde265=enabled,-Dlibde265=disabled,libde265"
 PACKAGECONFIG[libmms]          = "-Dlibmms=enabled,-Dlibmms=disabled,libmms"


recipes-multimedia/gstreamer/gstreamer1.0-plugins-bad_1.16.imx.bb
```diff
@@ -63,6 +66,7 @@ PACKAGECONFIG[hls]             = "-Dhls=enabled -Dhls-crypto=nettle,-Dhls=disabl
 # to -base yet. They depend on the gstgl library in -base, so we do
 # not add GL dependencies here, since these are taken care of in -base.
 PACKAGECONFIG[gl]              = "-Dgl=enabled,-Dgl=disabled,"
+PACKAGECONFIG[gles2]           = ",,virtual/libgles2"
```
otavio, line 69: Doesn't this use any build option for it? auto detection is bad so if we can "force" it, it is always nicer.

otavio (changes requested):
Hello @zakariazh, thanks for sending the PR. I made a small comment on the change you are proposing, but also we need to improve the commit log. Please follow the guidelines here:

https://www.openembedded.org/wiki/Commit_Patch_Message_Guidelines

recipes-multimedia/gstreamer/gstreamer1.0-plugins-bad_1.16.imx.bb
zakariazh: Hello @otavio, thanks for your reply. I didn't find any build option for this in gstreamer (so i took the same options in gstreamer1.0-plugins-base).

I propose this as commit log, it is good ? : 

gstreamer1.0-plugins-bad: Let processors with pxp chose the correct Opengl API

The processors without gpu (imx6ul/imx6ull) support only opengl-es.The system must be able to chose the correct API instead of forcing it for all processors.

Signed-off-by: ZIDOUH Zakaria <zakaria.zidouh@gmail.com>

otavio: Yes, the commit log makes sense. Please pay attention to the number of columns so it fits on the good practices.

@thochstein please review this commit as well.

recipes-multimedia/gstreamer/gstreamer1.0-plugins-bad_1.16.imx.bb
```diff
@@ -36,12 +36,15 @@ DEPENDS += "gstreamer1.0-plugins-base"
 
 inherit gobject-introspection
 
+PACKAGECONFIG_GL ?= "${@bb.utils.contains('DISTRO_FEATURES', 'opengl', 'gl', '', d)}"
+PACKAGECONFIG_GL_imxpxp = "gles2"
```
thochstein, line 40: The commit log suggests this is a change only for 6UL and 6ULL, but this override `_imxpxp` will change **all** of the 6s and 7s except for 6Q. Perhaps you want `_mx6ul` and `_mx6ull`?

zakariazh: I have not tested this on the other platforms, I think you are right to limit this change only on these two platforms, while waiting to test for the others.

thochstein: @CarolZhuYuqing, can you have a look at this as well?

zakariazh: @thochstein Do you want me to update my commit with your proposal or you will do that later?

zakariazh: Hi @CarolZhuYuqing, what do you think of that?

boucman: hi, is there any feedback ? I can't see any comment from @CarolZhuYuqing

otavio (approved):

recipes-multimedia/gstreamer/gstreamer1.0-plugins-bad_1.16.imx.bb
sapiippo, line 40: This doesn't seem right to me. 
I don't think the PXP provides any opengl (es) support.

recipes-multimedia/gstreamer/gstreamer1.0-plugins-bad_1.16.imx.bb
sapiippo, line 69: also AFAIK the plugins use GstGLAPI instead of directly the opengl API, so the dependency should not be necessary at all.

otavio: Have you prepare, and test, a PR with the proposed change?

recipes-multimedia/gstreamer/gstreamer1.0-plugins-bad_1.16.imx.bb
thochstein: We decided above to change this PR to be specific to 6UL and 6ULL, but I guess that was missed waiting for other feedback.

sapiippo:
@zakariazh What configuration was the issue first seen in? I haven't been able to reproduce it.

zakariazh:
Hi @sapiippo, i was compiling gstreamer1.0-plugins-bad_1.16 (dunfell branch) for imx6ull processor on ubuntu18. tell me if you can reproduce the error please.

sapiippo:
I've tried build without this patch using master branch for imx6ullevk and build works correctly.
The patch, however, breaks a build where opengl is not in DISTRO_FEATURES.

otavio:
@sapiippo So we can reproduce the build error with `fslc-framebuffer` distro?

sapiippo:
No, that still has opengl enabled in DISTRO_FEATURES.

I've removed opengl DISTRO_FEATURE in our distro for all machines not supporting it. This should probably be somehow linked to MACHINE_FEATURES, but nothing for that seem to be available.

otavio:
So please report this issue so we can track it on a specific issue number.

PR #612, opened 2021-01-21
firmware-nxp-wifi: remove packages existing in upstream

zandrey:
linux-firmware package already has firmware for modules, provided in
this recipe.

Drop firmware packages for following modules:
- bcm4339
- bcm43430
- bcm43455
- bcm4356-pcie

Keep the bcm4359-pcie in the recipe as it is not yet made part of
linux-firmware package.

Add recipe description detailing the usage of this package and actions
to be taken once the firmware gets integrated upstream.

### DIFF
--- a/recipes-bsp/firmware-imx/firmware-nxp-wifi_1.0.bb
+++ b/recipes-bsp/firmware-imx/firmware-nxp-wifi_1.0.bb
@@ -1,6 +1,11 @@
 # Copyright 2020-2021 NXP
 
 SUMMARY = "Wi-Fi firmware redistributed by NXP"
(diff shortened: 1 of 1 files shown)


otavio:
Looking at the diff, the files will be included in PN package, am I wrong?

zandrey:
@otavio Good catch, had a small leftover in the commit. Thanks for pointing it out!

v2:
- Change the package addition order to collect all files in a proper package

zandrey:
changes in v3:
- Allow main PN to be empty, this satisfies the requirement during installation of `-dev` package

otavio (approved):

PR #613, opened 2021-01-21
Fixes for OpenCL and mesa.

thochstein:


conf/machine/include/imx-base.inc
```diff
@@ -202,11 +202,11 @@ MACHINE_SOCARCH_FILTER_append_use-mainline-bsp = " \
     qtbase \
 "
 MACHINE_SOCARCH_FILTER_append_mx6q = " \
-    virtual/opencl-icd \
```
thochstein, line 205: Hmm, just noticed that `virtual/opencl-icd` is used by opencv recipe:

```
meta-openembedded/meta-oe/recipes-support/opencv/opencv_4.5.0.bb:PACKAGECONFIG[opencl] = "-DWITH_OPENCL=ON,-DWITH_OPENCL=OFF,opencl-headers virtual/opencl-icd,"
```

I see no provider of this though. Is this a mistake?

otavio (changes requested):
Please improve the commit log. WHY this is being changed?

conf/machine/include/imx-base.inc
otavio: Did you check other layers?

thochstein: I tried google but it doesn't find much. Only this secondary addition https://scuderia.cs.hs-rm.de/andreas.werner/am57xx-bsp/-/blob/1389e2cad79be1a98672168fa2d23904e5ed2d51/meta-hsrm/meta-hsrm-ti/recipes-ti/ocl/opencl_git.bbappend.

otavio: But in this case we'd use the `opencl-icd-loader` right?

recipes-graphics/mesa/mesa_%.bbappend
```diff
@@ -17,10 +17,9 @@ python () {
 # Enable Etnaviv and Freedreno support
 PACKAGECONFIG_append_use-mainline-bsp = " gallium etnaviv kmsro freedreno"
 
-# Enable OSMesa which also requires dri (classic) swrast
-PACKAGECONFIG_remove_use-nxp-bsp = "gallium"
-PACKAGECONFIG_append_use-nxp-bsp = " osmesa"
-DRIDRIVERS_use-nxp-bsp = "swrast"
+# For NXP BSP, enable OSMesa for parts with DRM
+PACKAGECONFIG_remove_use-nxp-bsp_imxdrm = "gallium"
+PACKAGECONFIG_append_use-nxp-bsp_imxdrm = " osmesa"
```
thochstein, line 22: Hi @zandrey, I believe you added `swrast` in the first place. Were you having a build issue without it? Or was it a runtime issue? Prabhu says we should not be using `swrast` when we have hardware acceleration available.

zandrey: @thochstein IIRC, I had a build issue back then when HW acceleration was not available in Mesa (not implemented).

If the HW acceleration support is already fully available in Mesa - then I guess this can be dropped.

-- andrey

conf/machine/include/imx-base.inc
thochstein: I am aligning with `opencl-headers` and `opencl-icd-loader` and removing the virtual versions of these from our recipes. Then by setting the preferred version the `clinfo` recipe will use the `imx-gpu-viv` recipe instead of `opencl-headers` and `opencl-icd-loader` recipes.

otavio (approved):

thochstein:
Sorry I missed this, but removing `PROVIDES = "virtual/opencl-icd"` from `imx-gpu-viv` does break `opencv` when configured for `opencl`. I can add the `PROVIDES` back to `imx-gpu-viv` and then look to cleanup `meta-openembedded` by: a) updating `opencl-icd-loader` to provide `virtual/opencl-icd`, and b) updating `clinfo` to depend on it.

otavio:
Have you prepared a PR for it?

thochstein:
Yes, https://github.com/Freescale/meta-freescale/pull/617.

PR #614, opened 2021-01-25
machines: imx-base.inc: fix failing u-boot builds

MaxKrummenacher:
CC @thochstein

### DIFF
--- a/conf/machine/imx6qdlsabreauto.conf
+++ b/conf/machine/imx6qdlsabreauto.conf
@@ -40,6 +40,10 @@ UBOOT_SUFFIX = "img"
 SPL_BINARY = "SPL"
 WKS_FILE = "imx-uboot-spl-bootpart.wks.in"
 
+# The fsl distro uses u-boot-imx which does not provide unified functionality
+# for dl/q/qp SoC variants. Change the defconfig to the targeted SoC variant.
+UBOOT_MACHINE_fsl = "mx6qsabreauto_defconfig"
+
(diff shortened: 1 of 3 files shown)


otavio (changes requested):
Thanks @MaxKrummenacher for the fix. I agree with the fix and thinking deeper about the issue, I think we ought to remove the hack as you did and move this to `fsl-base` distro, where it belongs.

Could you rework this PR and prepare a new one for meta-freescale-distro?

MaxKrummenacher:
@otavio 
I can try to do so. E.g. move the
```
# u-boot-imx does not use a SPL
UBOOT_MAKE_TARGET_fsl = "u-boot.imx"
UBOOT_SUFFIX_fsl = "imx"
SPL_BINARY_fsl = ""
WKS_FILE_fsl = "imx-uboot-bootpart.wks.in"
```
to fsl-base.conf.

The used defconfig is a combination of machine and u-boot recipe so that I think this would need to stay in the <machine>.conf, rather than having a machine specific override in the distro.
However I'm unsure if this is really a better solution. It distributes the U-Boot configuration even further. And, more importantly if u-boot-imx should decide to move to a SPL solution for some machines for others not then the SPL <machine>.conf would again have to override the override from fsl-base or move the definition of the variables yet again.

What do you think?

MaxKrummenacher:
This is how it looks with the changes moved to the distro. Needs also https://github.com/Freescale/meta-freescale-distro/pull/46.

otavio (approved):

PR #646, opened 2021-02-12
alsa-state sound card updates

thochstein:
@zandrey , please have a look. I don't understand how this works and note that my commit here for `wm8960audio4` has the same justification and content as your recent commit for `sofwm8960audio`. For that matter each of the `*wm8960audio*` blocks have identical content.


thochstein:
Hi @zandrey , can you provide some guidance on this update?

zandrey:
@thochstein Sorry for delay... From the syntax point of view - it looks OK, I cannot test it in runtime though as I do not have a HW.

PR #656, opened 2021-02-18
basler-camera,isp-imx,kernel-module-isp-vvcam: Mark imx specific

kraj:
These recipes are really not meant for non imx architectures, therefore
pin them as such


chrisdimich:
Thanks @kraj ! You beat me to it :)

zandrey (changes requested):
@kraj @chrisdimich

Those packages are NXP-specific, therefore `use-mainline-bsp` override does not apply to them. Please drop it or replace with `use-nxp-bsp` instead.

-- andrey

otavio (approved):

PR #662, opened 2021-02-22
layer: license: fix GPL-2.0 license identifiers changed upstream

zandrey:
Upstream commit 2456f523cf ("licenses: Update license file to match current SPDX names") deprecated the usage of GPL-2.0 SPDX identifier, removed GPL-2.0 license file and replaced it with GPL-2.0-only SPDX identifier and text file.

Adjust layer recipes to use new SDPX identifier and text file.

imx-test recipe has license has been re-captured to use GPL-2.0-or-later identifier, since it is required by the license text accompanying the source code repository through "GPLv2 or later" statement in COPYING file.

-- andrey

### DIFF
--- a/recipes-bsp/imx-test/imx-test_git.bb
+++ b/recipes-bsp/imx-test/imx-test_git.bb
@@ -5,8 +5,8 @@
 SUMMARY = "Test programs for i.MX BSP"
 DESCRIPTION = "Unit tests for the i.MX BSP"
 SECTION = "base"
-LICENSE = "GPLv2"
-LIC_FILES_CHKSUM = "file://${COREBASE}/meta/files/common-licenses/GPL-2.0;md5=801f80980d171dd6425610833a22dbe6"
+LICENSE = "GPLv2+"
+LIC_FILES_CHKSUM = "file://${COREBASE}/meta/files/common-licenses/GPL-2.0-or-later;md5=fed54355545ffd980b814dab4a3b312c"
 
 DEPENDS = "alsa-lib libdrm"
 DEPENDS_append_mx6 = " imx-lib"
--- a/recipes-bsp/imx-mkimage/imx-boot_1.0.bb
+++ b/recipes-bsp/imx-mkimage/imx-boot_1.0.bb
@@ -4,7 +4,7 @@ require imx-mkimage_git.inc
 
 DESCRIPTION = "Generate Boot Loader for i.MX 8 device"
 LICENSE = "GPLv2"
-LIC_FILES_CHKSUM = "file://${COREBASE}/meta/files/common-licenses/GPL-2.0;md5=801f80980d171dd6425610833a22dbe6"
+LIC_FILES_CHKSUM = "file://${COREBASE}/meta/files/common-licenses/GPL-2.0-only;md5=801f80980d171dd6425610833a22dbe6"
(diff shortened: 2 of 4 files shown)


recipes-bsp/imx-test/imx-test_git.bb
```diff
@@ -6,7 +6,7 @@ SUMMARY = "Test programs for i.MX BSP"
 DESCRIPTION = "Unit tests for the i.MX BSP"
 SECTION = "base"
 LICENSE = "GPLv2"
```
otavio, line 8: Here it must be adjusted as well.

otavio (changes requested):

recipes-bsp/imx-test/imx-test_git.bb
zandrey: Correct, missed this part. I've adjusted it to `GPLv2+` in the force-push.

MaxKrummenacher: I think what Otavio meant is that 'GPLv2' and friends have also been deprecated.
So the line should be changed to "LICENSE = "GPL-2.0-only"  (or GPL-2.0-or-later).

https://spdx.github.io/spdx-spec/appendix-I-SPDX-license-list/

zandrey: Hello @MaxKrummenacher 

> 
> 
> I think what Otavio meant is that 'GPLv2' and friends have also been deprecated.

This I've also assumed, but haven't seen. According to [meta/conf/licenses.conf](https://git.openembedded.org/openembedded-core/tree/meta/conf/licenses.conf#n34) `GPLv2+` is still present in _SPDXLICENSEMAP_.

> So the line should be changed to "LICENSE = "GPL-2.0-only" (or GPL-2.0-or-later).
> 
> https://spdx.github.io/spdx-spec/appendix-I-SPDX-license-list/

So far, I haven't seen any occurrence of `LICENSE = "GPL-2.0-or-later"` in OE-Core, recipes are rather using the `LICENSE = "GPLv2+"` syntax.

`GPL-2.0-or-later`  on the other hand is used only in `LIC_FILES_CHKSUM` in several OE-Core recipes, e.g. [meta/recipes-core/util-linux/util-linux_2.36.2.bb](https://git.openembedded.org/openembedded-core/tree/meta/recipes-core/util-linux/util-linux_2.36.2.bb).

In this regard, I was just following the OE-Core approach of not changing the `LICENSE = "GPLv2+"` to `LICENSE = "GPL-2.0-or-later"`.

Please let me know if that makes sense and clarify things?

@otavio Should we change the `LICENSE` to full _SPDX_ identifiers, or stay with current ones (like in OE-Core)?

-- andrey

otavio: I prefer `GPLv2+`. The checksum will need to match the files and this out of recipe control so we need to use whatever the package use. If eventually OE-Core moves to `GPL-2.0-or-later` in the future we can replace it all at once when needed.

otavio (approved):

recipes-bsp/imx-test/imx-test_git.bb
MaxKrummenacher: I'm fine with that.

PR #665, opened 2021-02-25
opencv_4.4.0.imx: add recipe

chrisdimich:
Add recipe to support additional machine learning demos and tiny dnn.

### DIFF
--- a/recipes-support/opencv/opencv_4.4.0.imx.bb
+++ b/recipes-support/opencv/opencv_4.4.0.imx.bb
@@ -1,0 +1,253 @@
+# This recipe is for the i.MX fork of opencv.
+SUMMARY = "Opencv : The Open Computer Vision Library"
+HOMEPAGE = "http://opencv.org/"
+SECTION = "libs"
+
+LICENSE = "BSD-3-Clause"
+LIC_FILES_CHKSUM = "file://LICENSE;md5=19598330421859a6dd353a4318091ac7"
+
+ARM_INSTRUCTION_SET_armv4 = "arm"
+ARM_INSTRUCTION_SET_armv5 = "arm"
+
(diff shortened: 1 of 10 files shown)


recipes-support/opencv/opencv_4.4.0.imx.bb
```diff
@@ -0,0 +1,57 @@
+require recipes-support/opencv/opencv_4.4.0.bb
```
otavio, line 1: We should not require existing recipes. Instead, we should include the original recipe content as is and do the required changes below it, marking it accordingly.

One example of this can be seen at:

https://github.com/Freescale/meta-freescale/blob/master/recipes-graphics/wayland/weston_9.0.0.imx.bb

otavio (changes requested):

chrisdimich:
ah okay, sure no problem. I will do this and update PR.

chrisdimich:
@otavio updated recipe to remove the require, and combine the recipes into one.

otavio:
This is closer, but we need to make it easier to maintain for the future.

For example, https://github.com/Freescale/meta-freescale/blob/master/recipes-graphics/wayland/weston_9.0.0.imx.bb#L1-L6 marks the verbatim copy of the original file and https://github.com/Freescale/meta-freescale/blob/master/recipes-graphics/wayland/weston_9.0.0.imx.bb#L143-L145 the end. Then the i.MX fork changes are added.

Please try to follow this template as close as possible as the next upgrade becomes much easier to do.

chrisdimich:
Oh okay, I will do this. At the top of the recipe where it specifies the "Upstream hash" of the original weston, is this a hash of the original weston.bb recipe itself (md5?)

PR #667, opened 2021-02-26
opencv_4.4.0.imx: add recipe

chrisdimich:
Add recipe to support additional machine learning demos and tiny dnn.

### DIFF
--- a/recipes-support/opencv/opencv_4.4.0.imx.bb
+++ b/recipes-support/opencv/opencv_4.4.0.imx.bb
@@ -1,0 +1,295 @@
+# This recipe is for the i.MX fork of opencv. For ease of
+# maintenance, the top section is a verbatim copy of an OE-core
+# recipe. The second section customizes the recipe for i.MX.
+
+########## meta-openembedded copy ###########
+
+SUMMARY = "Opencv : The Open Computer Vision Library"
+HOMEPAGE = "http://opencv.org/"
+SECTION = "libs"
+
+LICENSE = "BSD-3-Clause"
+LIC_FILES_CHKSUM = "file://LICENSE;md5=19598330421859a6dd353a4318091ac7"
+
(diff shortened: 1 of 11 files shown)


otavio (changes requested):
Awesome, however we need to add the preferences and compatibility here otherwise it'll have multiple providers for it.

chrisdimich:
Right, as we discussed I was going to add that to imx-base.inc, but was waiting for this to be pushed first. Let me add that to this PR.

recipes-support/opencv/opencv_4.4.0.imx.bb
```diff
@@ -0,0 +1,295 @@
...
+    install -d ${D}${datadir}/OpenCV/samples/bin/
+    cp -f bin/example_* ${D}${datadir}/OpenCV/samples/bin/
+    if ${@bb.utils.contains('PACKAGECONFIG', 'test', 'true', 'false', d)}; then
+        cp -r share/opencv4/testdata/cv/face/* ${D}${datadir}/opencv4/testdata/cv/face/
+    fi
+}
+
+FILES_${PN}-samples += "${datadir}/OpenCV/samples"
+
+########## End of i.MX overrides ##########
```
otavio, line 293: We need to restrict the recipe to mx8.

chrisdimich:
@otavio gentle reminder :) I made the change you requested and restricted recipe to mx8 only.

recipes-support/opencv/opencv_4.4.0.imx.bb
otavio: This commit belongs to the recipe patch, not the `imx-base.inc`.

chrisdimich:
@otavio I split the commits. Is this all good now?

otavio (approved):

PR #680, opened 2021-03-05
Boot Container support for mainline BSP and mx8m machines

zandrey:
This PR provides a mechanism to create a full image for `mx8m` derivatives based on Community BSP (when _IMX_DEFAULT_BSP_ is set to `use-mainline-bsp`) and allows creation of runtime images that could be used on those derivatives.

Implementation provides a "boot container", which serves as a replacement of `imx-boot` package used in NXP BSP, and is populated in the final image file via new WIC template file.

Basic mechanism description can be summarized as following:
1. Machines that would like to utilize the boot container defines `boot-container` in their _MACHINEOVERRIDES_, which indicates that components required to produce a container are receiving compatibility option.
2. Machines are also **required** to define _ATF_LOAD_ADDR_ parameter, which is used in compilation process of container.
3. Base machine include file defines a variable _UBOOT_PROVIDES_BOOT_CONTAINER_, which indicates that the U-Boot package build should be extended with additional functionality that creates the boot container. This variable is set to "1" in case if `boot-container` is present in _MACHINEOVERRIDES_.
4. U-Boot package build system is extended via inheritance on the separate class (`imx-boot-container`), which takes care of the steps required before container can be created, and deployment of the result container binary after build completes.
5. U-Boot build system creates a container, which is done via special U-Boot build target: `flash.bin`
6. Boot container is deployed and picked up by a WIC file to the result image. It resides at the same location as a binary produced by `imx-boot` package in NXP BSP.
7. Result image file can be then further used to boot the target using the boot container.

This approach has been built and verified on `imx8mm-lpddr4-evk`, `imx8mn-ddr4-evk` and `imx8mp-evk` machines with _DISTRO=poky_.

Build has also been performed for above machines with `fsl-xwayland` distro to make sure that no impact of this implementation is recorded on NXP BSP flavor.

In addition, build of `DISTRO=poky MACHINE=imx7dsabresd bitbake core-image-base` has been performed to make sure that machines utilizing mainline BSP and not setting `boot-container` in their _MACHINEOVERRIDES_ are also unaffected.

--
Cc: @gibsson @MaxKrummenacher @thochstein @fabioestevam @pberginkonsult @hutchman @RobertBerger

### DIFF
--- a/classes/imx-boot-container.bbclass
+++ b/classes/imx-boot-container.bbclass
@@ -1,0 +1,107 @@
+#
+# This class provides a support to build the boot container for
+# i.MX8M derivatives
+#
+# imx8m machines require a separate build target to be executed
+# due to the fact that final boot image is constructed using flash.bin
+# taget. It produces a boot binary image, which is constructed from
+# various binary components (u-boot with separate dtb, atf, DDR
+# firmware and optional op-tee) into a single image using FIT format.
+# This flash.bin file is then parsed and loaded either via
+# SPL directly (imx8mm), or using bootrom code (imx8mn and imx8mp).
+#
+# In order to use flash.bin binary boot image, it is required that
+# the U-Boot build is to be invoked for an additional build target.
+# This class extendes the U-Boot build targets with the "flash.bin",
+# which effectively serves as a boot container. It replaces the boot
+# container `imx-boot` provided by NXP.
+#
+# Class inheritance is performed in u-boot-fslc recipe, and is controlled
+# by variable UBOOT_PROVIDES_BOOT_CONTAINER, which is defined in the
+# base machine include file (imx-base.inc), and is set to "1" when the
+# 'imx-boot-container' is present in MACHINEOVERRIDES.
+
+# Extend the UBOOT_MAKE_TARGET with additional target for U-Boot build
+# system to produce the boot container
+UBOOT_MAKE_TARGET += "flash.bin"
+
+# Define ATF binary file to be deployed to the U-Boot build folder
+ATF_MACHINE_NAME = "bl31-${ATF_PLATFORM}.bin"
+ATF_MACHINE_NAME_append = "${@bb.utils.contains('MACHINE_FEATURES', 'optee', '-optee', '', d)}"
+
+# This package aggregates output deployed by other packages, so set the
+# appropriate dependencies for populate binaries task
+do_resolve_and_populate_binaries[depends] += " \
+    ${@' '.join('%s:do_deploy' % r for r in '${IMX_EXTRA_FIRMWARE}'.split() )} \
+    imx-atf:do_deploy \
+    ${@bb.utils.contains('MACHINE_FEATURES', 'optee', 'optee-os:do_deploy', '', d)} \
+"
+
+# Append make flags to include ATF load address
+EXTRA_OEMAKE += "ATF_LOAD_ADDR=${ATF_LOAD_ADDR}"
+
+# Define an additional task that collects binary output from dependent packages
+# and deploys them into the U-Boot build folder
+do_resolve_and_populate_binaries() {
+    if [ ! -n "${ATF_LOAD_ADDR}" ]; then
+        bberror "ATF_LOAD_ADDR is undefined, result binary would be unusable!"
+    fi
+
+    if [ -n "${UBOOT_CONFIG}" ]; then
+        for config in ${UBOOT_MACHINE}; do
+            i=$(expr $i + 1);
+            for type in ${UBOOT_CONFIG}; do
+                j=$(expr $j + 1);
+                if [ $j -eq $i ]; then
+                    for ddr_firmware in ${DDR_FIRMWARE_NAME}; do
+                        # Sanitize the FW name as U-Boot expects it to be without version
+                        if [ -n "${DDR_FIRMWARE_VERSION}" ]; then
+                            ddr_firmware_name=$(echo $ddr_firmware | sed s/_${DDR_FIRMWARE_VERSION}//)
+                        else
+                            ddr_firmware_name="$ddr_firmware"
+                        fi
+                        bbnote "Copy ddr_firmware: ${ddr_firmware} from ${DEPLOY_DIR_IMAGE} -> ${B}/${config}/${ddr_firmware_name}"
+                        cp ${DEPLOY_DIR_IMAGE}/${ddr_firmware} ${B}/${config}/${ddr_firmware_name}
+                    done
+                    if [ -n "${ATF_MACHINE_NAME}" ]; then
+                        cp ${DEPLOY_DIR_IMAGE}/${BOOT_TOOLS}/${ATF_MACHINE_NAME} ${B}/${config}/bl31.bin
+                    else
+                        bberror "ATF binary is undefined, result binary would be unusable!"
+                    fi
+                fi
+            done
+            unset  j
+        done
+        unset  i
+    fi
+}
+addtask do_resolve_and_populate_binaries before do_compile after do_configure
+
+# Append the u-boot do_deploy task to deploy also the result flash.bin
+# boot container as a replacement for the one provided by NXP BSP.
+#
+# Deploy task renames the target file from flash.bin to imx-boot to match
+# the name WKS file requires.
+#
+# This effectively would allow the usage of the same WKS file for those
+# derivatives that are using the boot container from U-Boot and those
+# that are not yet have support for it enabled.
+do_deploy_append() {
+    # Deploy the resulted flash.bin for WIC to pick it up
+    if [ -n "${UBOOT_CONFIG}" ]; then
+        for config in ${UBOOT_MACHINE}; do
+            i=$(expr $i + 1);
+            for type in ${UBOOT_CONFIG}; do
+                j=$(expr $j + 1);
+                if [ $j -eq $i ]
+                then
+                    install -m 0644 ${B}/${config}/flash.bin  ${DEPLOYDIR}/flash.bin-${MACHINE}-${UBOOT_CONFIG}
+                    ln -sf flash.bin-${MACHINE}-${UBOOT_CONFIG} imx-boot
+                    ln -sf flash.bin-${MACHINE}-${UBOOT_CONFIG} flash.bin
+                fi
+            done
+            unset  j
+        done
+        unset  i
+    fi
+}
--- a/conf/machine/imx8mp-evk.conf
+++ b/conf/machine/imx8mp-evk.conf
@@ -4,7 +4,7 @@
 #@DESCRIPTION: Machine configuration for NXP i.MX 8M Plus Evaluation Kit
 #@MAINTAINER: Jun Zhu <junzhu@nxp.com>
 
-MACHINEOVERRIDES =. "mx8:mx8m:mx8mp:"
+MACHINEOVERRIDES =. "imx-boot-container:mx8:mx8m:mx8mp:"
 
 require conf/machine/include/imx-base.inc
 require conf/machine/include/tune-cortexa53.inc
@@ -59,8 +59,12 @@ DDR_FIRMWARE_NAME = " \
 # Set u-boot DTB
 UBOOT_DTB_NAME = "imx8mp-evk.dtb"
 
-# Set ATF platform name
+# Set ATF platform name and load address
 ATF_PLATFORM = "imx8mp"
+ATF_LOAD_ADDR = "0x970000"
+
+# Extra firmware package name, that is required to build boot container for fslc bsp
+IMX_EXTRA_FIRMWARE = "firmware-imx-8m"
 
 # Set imx-mkimage boot target
 IMXBOOT_TARGETS = "${@bb.utils.contains('UBOOT_CONFIG', 'fspi', 'flash_evk_flexspi', 'flash_evk', d)}"
@@ -71,10 +75,11 @@ SERIAL_CONSOLES = "115200;ttymxc1"
 
 LOADADDR = ""
 UBOOT_SUFFIX = "bin"
-UBOOT_MAKE_TARGET = ""
+UBOOT_MAKE_TARGET = "all"
 IMX_BOOT_SEEK = "32"
 
 OPTEE_BIN_EXT = "8mp"
+TEE_LOAD_ADDR = "0x56000000"
 
 # Add additional firmware
 MACHINE_FIRMWARE_append = " linux-firmware-ath10k"
--- a/conf/machine/imx8mp-evk.conf
+++ b/conf/machine/imx8mp-evk.conf
@@ -9,8 +9,11 @@ MACHINEOVERRIDES =. "imx-boot-container:mx8:mx8m:mx8mp:"
 require conf/machine/include/imx-base.inc
 require conf/machine/include/tune-cortexa53.inc
 
-MACHINE_FEATURES += "pci wifi bluetooth optee jailhouse"
-MACHINE_FEATURES_append_use-nxp-bsp = " mrvl8997"
+MACHINE_FEATURES += "pci wifi bluetooth jailhouse"
+
+# OP-TEE is also applicable to NXP BSP, mainline BSP seems not to have
+# a full support for it yet.
+MACHINE_FEATURES_append_use-nxp-bsp = " optee mrvl8997"
 
 # Mainline kernel contains only one DTB file for
 # imx8mpevk machine
--- a/conf/machine/include/imx-base.inc
+++ b/conf/machine/include/imx-base.inc
@@ -57,6 +57,12 @@ UBOOT_ENTRYPOINT_mx7ulp = "0x60008000"
 UBOOT_ENTRYPOINT_mx8m   = "0x40480000"
 UBOOT_ENTRYPOINT_vf = "0x80008000"
 
+# Some derivates can utilize the boot container provided by U-Boot,
+# below variable sets that those machines which have a imx-boot-container
+# in their MACHINEOVERRIDES can inherit a imx-boot-container class
+UBOOT_PROVIDES_BOOT_CONTAINER = "0"
+UBOOT_PROVIDES_BOOT_CONTAINER_imx-boot-container = "1"
+
 PREFERRED_PROVIDER_virtual/xserver = "xserver-xorg"
 XSERVER_DRIVER                  = "xf86-video-fbdev"
 XSERVER_DRIVER_imxgpu2d         = "xf86-video-imx-vivante"
@@ -390,13 +396,37 @@ WKS_FILE_DEPENDS_append_mx8 = " imx-boot"
 WKS_FILE_DEPENDS_append_mx8m = " imx-boot"
 
 # We need to restrict the append so we don't add this for other i.MX SoC's.
-WKS_FILE_DEPENDS_append_use-mainline-bsp_aarch64 = " imx-boot"
+# Derivatives that are not yet adopted the usage of boot container provided
+# by U-Boot build are still targeted to use 'imx-boot' package provided by
+# NXP. Moving those derivatives to mainline BSP would require to define an
+# 'imx-boot-container' override, and test if the U-Boot built 'flash.bin'
+# binary is used a replacement.
+# Note, that the results binary name of the boot container is set to 'imx-boot'
+# for both NXP and Mainline BSP.
+# For Mainline BSP: the 'flash.bin' boot container is renamed during the
+# deployment task extesion execution defined in imx-boot-container class.
+# For NXP BSP: rename is done in 'imx-boot' recipe at the execution of compile
+# task.
+WKS_FILE_DEPENDS_append_use-mainline-bsp_aarch64 = " \
+    ${@oe.utils.ifelse(d.getVar('UBOOT_PROVIDES_BOOT_CONTAINER') == '0', 'imx-boot', '')} \
+"
 
 SOC_DEFAULT_WKS_FILE ?= "imx-uboot-bootpart.wks.in"
 SOC_DEFAULT_WKS_FILE_mx8m ?= "imx-imx-boot-bootpart.wks.in"
 
 SOC_DEFAULT_WKS_FILE_mx8 ?= "imx-imx-boot-bootpart.wks.in"
 SOC_DEFAULT_WKS_FILE_mxs ?= "imx-uboot-mxs-bootpart.wks.in"
+
+# Boot container built as a part of mainline U-Boot uses the same WKS
+# file as the entire mx8m series, as it renames flash.bin binary to
+# imx-boot before it is packed into the boot partition.
+# This operation is performed in imx-boot-container class as a part of
+# delopyment task.
+# flash.bin binary is produced by U-Boot build itself, and is serves as a
+# direct replacement of imx-boot from NXP.
+# Creation of the flash.bin is controlled by UBOOT_PROVIDES_BOOT_CONTAINER
+# variable defined above
+SOC_DEFAULT_WKS_FILE_imx-boot-container ?= "imx-imx-boot-bootpart.wks.in"
 
 WKS_FILE ?= "${SOC_DEFAULT_WKS_FILE}"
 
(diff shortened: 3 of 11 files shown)


otavio:
@zandrey, I'd like to thank you for this work, publically. We've been talking about this behind the scenes for a while, and it is exciting to see it being sent for formal review.

There is one thing I'd like to explore, which is to use the override as `imx-boot-container`. We should keep it in a namespace, so it doesn't conflict with a future `qoriq-boot-container` feature or another NIP under development.

RobertBerger:
@zandrey, I am so proud of you! I would also like to thank you and the people who back your great work. Did you notice that people are referring to meta-freescale as a reference for how to make a proper BSP?

MaxKrummenacher:
@zandrey Great work Andrey, thank you.

conf/machine/include/imx-base.inc
```diff
@@ -400,6 +406,8 @@ SOC_DEFAULT_WKS_FILE_mxs ?= "imx-uboot-mxs-bootpart.wks.in"
 # WKS file, as it packs the flash.bin binary into the boot partition.
 # This binary is produced by U-Boot build itself, and is serves as a
 # direct replacement of imx-boot from NXP.
+# Creation of the flash.bit is controlled by UBOOT_PROVIDES_BOOT_CONTAINER
```
MaxKrummenacher, line 409: I guess that would be s/flash.bit/flash.bin/

conf/machine/imx8mp-evk.conf
```diff
@@ -4,7 +4,7 @@
 #@DESCRIPTION: Machine configuration for NXP i.MX 8M Plus Evaluation Kit
 #@MAINTAINER: Jun Zhu <junzhu@nxp.com>
 
-MACHINEOVERRIDES =. "mx8:mx8m:mx8mp:"
+MACHINEOVERRIDES =. "boot-container:mx8:mx8m:mx8mp:"
```
MaxKrummenacher, line 7: You unconditionally use now U-Boot to provide the boot container if I understand this line correctly.
I'd propose to rename this commit as this commit does actually use the support the other commits are providing. E.g.

conf: machine: mx8m[m,n,p]: use boot container class

conf/machine/include/imx-base.inc
```diff
@@ -390,15 +390,18 @@ WKS_FILE_DEPENDS ?= " \
 WKS_FILE_DEPENDS_append_mx8 = " imx-boot"
 WKS_FILE_DEPENDS_append_mx8m = " imx-boot"
 
-# We need to restrict the append so we don't add this for other i.MX SoC's.
-WKS_FILE_DEPENDS_append_use-mainline-bsp_aarch64 = " imx-boot"
```
MaxKrummenacher, line 394: What if someone does not change to the _boot-container override?
Also I guess this drops the depend for the i.MX 8/i.MX 8X SoC.

WKS_FILE_DEPENDS_append_use-mainline-bsp_aarch64 = "${@bb.utils.contains('OVERRIDES', 'boot-container', '', ' imx-boot', d)} "

classes/imx-boot-container.bbclass
```diff
@@ -0,0 +1,106 @@
...
+# by variable UBOOT_PROVIDES_BOOT_CONTAINER, which is defined in the
+# base machine include file (imx-base.inc), and is set to "1" when the
+# boot-container is present in MACHINEOVERRIDES.
+
+# Define additional dependencies required to produce a boot container
+DEPENDS_append = " \
+    ${IMX_EXTRA_FIRMWARE} \
+    imx-atf \
+    ${@bb.utils.contains('MACHINE_FEATURES', 'optee', 'optee-os', '', d)} \
+"
```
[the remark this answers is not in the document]
otavio: I believe @MaxKrummenacher is right. We have this in line 44, so those wouldn't need it. Have you found any errors without this?

classes/imx-boot-container.bbclass
```diff
@@ -0,0 +1,106 @@
...
+    # Deploy the resulted flash.bin for WIC to pick it up
+    if [ -n "${UBOOT_CONFIG}" ]; then
+        for config in ${UBOOT_MACHINE}; do
+            i=$(expr $i + 1);
+            for type in ${UBOOT_CONFIG}; do
+                j=$(expr $j + 1);
+                if [ $j -eq $i ]
+                then
+                    install -m 0777 ${B}/${config}/flash.bin  ${DEPLOYDIR}/flash.bin-${MACHINE}-${UBOOT_CONFIG}
+                    ln -sf flash.bin-${MACHINE}-${UBOOT_CONFIG} flash.bin
```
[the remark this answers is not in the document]
otavio: I believe the `flash.bin` here is closer to what U-Boot mainline uses. I have mixed opinions about this.

conf/machine/include/imx-base.inc
```diff
@@ -390,15 +390,18 @@ WKS_FILE_DEPENDS ?= " \
...
 
 SOC_DEFAULT_WKS_FILE_mx8 ?= "imx-imx-boot-bootpart.wks.in"
 SOC_DEFAULT_WKS_FILE_mxs ?= "imx-uboot-mxs-bootpart.wks.in"
 
+# Boot container built as a part of mainline U-Boot uses a separate
+# WKS file, as it packs the flash.bin binary into the boot partition.
+# This binary is produced by U-Boot build itself, and is serves as a
+# direct replacement of imx-boot from NXP.
+SOC_DEFAULT_WKS_FILE_boot-container ?= "imx-boot-container-bootpart.wks.in"
+
```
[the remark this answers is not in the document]
otavio: Or we could move imx-boot to use this?

recipes-bsp/u-boot/u-boot-fslc_2020.10.bb
```diff
@@ -6,7 +6,9 @@ order to provide support for some backported features and fixes, or because it \
 was submitted for revision and it takes some time to become part of a stable \
 version, or because it is not applicable for upstreaming."
 
-DEPENDS_append = " bc-native dtc-native lzop-native"
+inherit ${@oe.utils.ifelse(d.getVar('UBOOT_PROVIDES_BOOT_CONTAINER') == '1', 'imx-boot-container', '')}
```
MaxKrummenacher, line 9: Couldn't we drop the new UBOOT_PROVIDES_BOOT_CONTAINER. This location seems to be the only user of it and it could directly test for the override.

inherit ${@bb.utils.contains('OVERRIDES', 'boot-container', 'imx-boot-container', '')}

conf/machine/include/imx-base.inc
MaxKrummenacher: Additionally I guess the `WKS_FILE_DEPENDS_append_mx8m = " imx-boot"` should also get the check for boot-container override or with use-nxp-bsp imx-boot will be additionally built and probably clash somehow with the u-boot provided boot container.

MaxKrummenacher: "${@bb.utils.contains('OVERRIDES', 'boot-container', '', ' imx-boot', d)} "
That will not work as OVERRIDES is a : separated variable.

recipes-bsp/u-boot/u-boot-fslc_2020.10.bb
MaxKrummenacher: Bad idea as OVERRIDES use : separator.

zandrey:
Hello all,

First of, thanks a lot for your kind words! I was hoping that this implementation would be beneficial for a lot of people, especially those who would like to build Community BSP for `mx8m` derivatives (myself included). Hope this implementation would open up the road to perform this, and also bring NXP and Community BSP closer here.

@MaxKrummenacher Thanks for the review! I'll comment on your points above, and would address some in _V2_ further.

I would also appreciate if anyone interested in this implementation drop their comments here, as more reviews this would go through - better the final implementation would be, and all other derivatives would be able to use it with as little modifications as possible.

Thanks a lot!

-- andrey

recipes-bsp/u-boot/u-boot-fslc_2020.10.bb
zandrey: Correct, this I've already tried out and realized _OVERRIDES_ would not be usable here. It would require `.split()` function and then the construct would become quite unusable.

zandrey:
> 
> There is one thing I'd like to explore, which is to use the override as `imx-boot-container`. We should keep it in a namespace, so it doesn't conflict with a future `qoriq-boot-container` feature or another NIP under development.

It make total sense, would address this in _V2_. 👍🏻

classes/imx-boot-container.bbclass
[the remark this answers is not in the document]
zandrey: This was a leftover block, would address this in _V2_

classes/imx-boot-container.bbclass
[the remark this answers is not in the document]
zandrey: @MaxKrummenacher This could potentially be done, thus reducing the implementation.

My original idea however was to completely separate `imx-boot` from `flash.bin` all the way down to WKS file. By this, we would have a independent implementation for boot container, and in case NXP would change their WKS generation - we would not be dependent on it.

Second, renaming `flash.bin` to `imx-boot` in `do_deploy` task would rather obscure the Community part of boot container, thus making it not so visible for those who utilizes it. It is not a major issue, rather a clean readability.

I tend to agree with @otavio here that the path from U-Boot providing `flash.bin` should rather be kept here.

If there is a strong objection to those points above - I can address it further in _V2_ though.

conf/machine/include/imx-base.inc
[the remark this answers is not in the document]
zandrey: It would completely depend on the above point, and I really would like to collect opinions on it before switching to a unified WKS file.

conf/machine/include/imx-base.inc
zandrey: Thanks, typos are evil! 😉

conf/machine/include/imx-base.inc
zandrey: I'd change this part to `use-nxp-bsp` in _V2_, overrides are not going to fully work here.

conf/machine/imx8mp-evk.conf
zandrey: Yes, for Community BSP - it is added unconditionally. NXP BSP is unaffected here, as it would use `u-boot-imx` instead of -u-boot-fslc`.

Would clarify the commit message in _V2_

classes/imx-boot-container.bbclass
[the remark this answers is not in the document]
MaxKrummenacher: Well, flash.bin is also  the file name created by imx-boot/imx-mkimage. It only gets renamed in the OE recipe.
And the boot containers from a build directly from U-Boot vs. the one where the assembly is done with imx-boot/imx-mkimage should be identical.
Thus I think for what comes after the creation of the boot container having one name simplifies matters and does not complicate it. E.g. one wks file instead of two, one uuu script instead of two ...

conf/machine/imx8mp-evk.conf
```diff
@@ -4,13 +4,13 @@
...
 
 require conf/machine/include/imx-base.inc
 require conf/machine/include/tune-cortexa53.inc
 
-MACHINE_FEATURES += "pci wifi bluetooth optee jailhouse"
-MACHINE_FEATURES_append_use-nxp-bsp = " mrvl8997"
+MACHINE_FEATURES += "pci wifi bluetooth jailhouse"
+MACHINE_FEATURES_append_use-nxp-bsp = " optee mrvl8997"
```
pberginkonsult, line 13: There is a comment in imx8m{m,n}-evk.conf files that I think is valid here too?
"
    # OP-TEE is also applicable to NXP BSP, mainline BSP seems not to have
    # a full support for it yet.
"

pberginkonsult (commented):
I have reviewed this from my perspective and knowledge. No big comments mostly cosmetic stuff.

One general question/thought; would it be possible and of value to get rid of the imx-boot_1.0.bb recipe usage for imx8m completely? Looking in to the imx-boot_1.0.bb recipe I find a lot of similarities with the new class imx-boot-container.bbclass. From maintenance and ease of understanding it would be good if there was a common way to create flash.bin from u-boot-fslc and u-boot-imx. I have not digged up all details around this if it is feasible or not. I just tried to check out u-boot-imx and there it was possible to 'make flash.bin' for imx8mp_evk_defconfig and a it looked like the container was created. (I have not verified it nor run-time tested it just a quick test to build.) Someone that have more knowledge and can comment on this?

zandrey:
Hello @pberginkonsult 

> 
> 
> I have reviewed this from my perspective and knowledge. No big comments mostly cosmetic stuff.

Thanks a lot for the review, I would address all your findings in _V2_ also.

> 
> One general question/thought; would it be possible and of value to get rid of the imx-boot_1.0.bb recipe usage for imx8m completely? Looking in to the imx-boot_1.0.bb recipe I find a lot of similarities with the new class imx-boot-container.bbclass. From maintenance and ease of understanding it would be good if there was a common way to create flash.bin from u-boot-fslc and u-boot-imx. I have not digged up all details around this if it is feasible or not. I just tried to check out u-boot-imx and there it was possible to 'make flash.bin' for imx8mp_evk_defconfig and a it looked like the container was created. (I have not verified it nor run-time tested it just a quick test to build.) Someone that have more knowledge and can comment on this?

One problem here is: `imx-boot` is used for variety of _mx8m_ derivatives, and does include more binary blobs in result container than `flash.bin` does (HDMA, SECO, SCFW to name the few) based on what specific SoC might require. `flash.bin` on the other hand only packs DDR, ATF and optional (and currently unused) OP-TEE - those that are required to boot the chip.

Indeed, the `u-boot-imx` is capable of producing the `flash.bin` target as well, but if you look at the compile log, you might messages that stating something like "_DDR firmware not found, result binary is unusable_", which points to the fact that necessary firmware files are not deployed prior to compile step, making the result `flash.bin` produced not the be bootable.

The new class addresses exactly this: it collects and deploys necessary binaries _before_ the compile step in order to produce a bootable image.

Those two mechanisms, even though the are similar in cause - are currently used "orthogonal", thus making it hard to simply replace `imx-boot` with `flash.bin`.

If one can find a way to make it - I would be really happy help with adoption though!

-- andrey

zandrey:
@MaxKrummenacher @pberginkonsult @otavio 

Regarding the `imx-boot` <-> `flash.bin` name usage for the result boot container, I see there are now several opinions that I would like to get clarified before I would address them in _V2_ further.

The way I see it, there are now 3 alternatives to proceed, namely:
1. Unify the naming at `imx-boot` by renaming the `flash.bin` produced by U-Boot and continue to use WKS file from NXP BSP
2. Unify the naming at `flash.bin` by renaming the `imx-boot` binary produced by NXP BSP, and use new WKS file from this layer
3. Keep both result binary names as they are produced by respective recipes and use separate WKS files

All of those alternatives are technically possible to implement, but my take on this would be (referred to the list above):
1. This would make the layer dependent on any future changes NXP might introduce to their WKS file (additions, removals, partition scheme modifications), which could lead to a situation when the compatibility between NXP and mainline implementation would not hold anymore, and we would be forced to either address this with NXP or switch to use separate WKS files (as in 3)
2. This might bring additional maintenance efforts as `flash.bin` would be made dependent on potential changes NXP would do on their end in either WKS file or in container itself. Any conflicts arising here would need to be addressed in upstream U-Boot build system to make sure that the unification point still holds, otherwise - separate WKS for `flash.bin` would be required (again, as in 3).
3. This approach brings synchronization maintenance, and changes on both sides might need to be ported across (although - not necessary). In addition, it would be required that two WKS files are kept, which might require some additional work from SOM/Board vendors.

Again, this is my take on this and now I'm a bit puzzled on which road to take further on. To me, in a long shot - the option 3 is a preference, but I might be decision-biased here since I've brought it on the first place. 😉 

Considering all above, can we perhaps work out a common agreement on which option to take?

-- andrey

conf/machine/include/imx-base.inc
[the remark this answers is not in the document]
otavio: I believe in the long term, we should get rid of imx-boot completely and adjust existing u-boot-imx users to use this new mechanism.

otavio:
I'd like to preferably let this implementation as clean as possible (1) and later see if we can port the existing boards to use the new schema and then drop the old one. This allows a clean and lean base for work in the next PRs. Does it make sense?

pberginkonsult:
I support going for 3) now. I would say it is good if we can go with the u-boot community and their naming (flash.bin). But either way works. Option 3) will add some maintenance, but are there really that much changes in the wks files expected? I don't think so. From experience they are mostly for demo on EVKs and then you have to make your own for your project that fits your need. I also think that vendors pick either u-boot-imx or u-boot-fslc and have only one setup for their BSP. I think it could be more work if we first go with 1) and then later change naming for both.

In the long run we should try to influence NXP to also follow this path and the community and end up in 2. But that is probably a challenge. ;-)

zandrey:
Changes in _V2_:
- Replace `boot-container` override with `imx-boot-container`
- Drop `DEPENDS` from the class as it is not required
- Drop separate WKS file for boot container, deployed `flash.bin` as `imx-boot` and use WKS file from NXP BSP
- Drop compatibility of `imx-boot` package with mainline BSP; `mx8` derivatives that are built using `use-mainline-bsp` override should use the new boot container mechanism instead of one provided by NXP BSP
- Return `WKS_FILE_DEPENDS` variable append when `use-mainline-bsp` is present and made it dependent on `UBOOT_PROVIDES_BOOT_CONTAINER` variable. This ensures that derivative currently not switched to new boot container scheme are using `imx-boot` container 
- Fix typos here and there

Thanks to all of you who provided a review here!

For the boot container deployment into WIC image: I decided to pursue the [Option 1 mentioned above](https://github.com/Freescale/meta-freescale/pull/680#issuecomment-793089953), as mainly it does simplifies the implementation. If this approach would not work out in the long run - there is always a possibility to introduce the separate WKS file and got for _Option 3_.

As a consequence, I also added in _V2_: compatibility of `imx-boot` package has been returned to the state, where it can only be used with NXP BSP. This would also ensure that we do not pack `imx-boot` container accidentally instead of the renamed `flash.bin` one when building for Mainline BSP.

-- andrey

MaxKrummenacher:
V2 looks good to me.

Just to be picky:
`This would also ensure that we do not pack imx-boot container accidentally instead of the renamed flash.bin one when building for Mainline BSP.`
Please note that the imx-boot/imx-mkimage recipes do also generate a flash.bin when calling make and then rename it as part of the OE recipe.

Thanks again Andrey.

conf/machine/include/imx-base.inc
zandrey: This has been addressed in _V2_.

zandrey:
@MaxKrummenacher Thanks a lot for looking into _V2_!

The quote you entered below is only in the "Cover Message" describing what was changed in _V2_. You're right - indeed the `imx-boot` recipe does this in `do_compile()` task after the binary is produced, it can be found in [imx-boot_1.0.bb#L129](https://github.com/Freescale/meta-freescale/blob/0a435d64e588843f6943d6e5723e5ecc12990aa2/recipes-bsp/imx-mkimage/imx-boot_1.0.bb#L129). But I guess this is not really relevant for this implementation.

Do you want me to put this note in a commit message as well?

-- andrey

conf/machine/include/imx-base.inc
```diff
@@ -391,14 +397,32 @@ WKS_FILE_DEPENDS_append_mx8 = " imx-boot"
 WKS_FILE_DEPENDS_append_mx8m = " imx-boot"
 
 # We need to restrict the append so we don't add this for other i.MX SoC's.
-WKS_FILE_DEPENDS_append_use-mainline-bsp_aarch64 = " imx-boot"
+# Derivatives that are not yet adopted the usage of boot container provided
+# by U-Boot build are still targeted to use 'imx-boot' package provided by
+# NXP. Moving those derivatives to mainline BSP would require to define an
+# 'imx-boot-container' override, and test if the U-Boot built 'flash.bin'
+# binary is used a replacement.
```
otavio, line 404: Here we mention `flash.bin`. I believe it's worth mentioning it is renamed later.

classes/imx-boot-container.bbclass
```diff
@@ -0,0 +1,106 @@
...
+do_deploy_append() {
+    # Deploy the resulted flash.bin for WIC to pick it up
+    if [ -n "${UBOOT_CONFIG}" ]; then
+        for config in ${UBOOT_MACHINE}; do
+            i=$(expr $i + 1);
+            for type in ${UBOOT_CONFIG}; do
+                j=$(expr $j + 1);
+                if [ $j -eq $i ]
+                then
+                    install -m 0777 ${B}/${config}/flash.bin  ${DEPLOYDIR}/flash.bin-${MACHINE}-${UBOOT_CONFIG}
```
otavio, line 98: Please change the mode to `0644`. Also, add the `flash.bin` name as the final link next to `imx-boot`, so if we decide to clean this up, we can use it as an artifact during the development work.

conf/machine/include/imx8mn-evk.inc
```diff
@@ -38,14 +42,15 @@ SERIAL_CONSOLES = "115200;ttymxc1"
 BOOT_SPACE = "65536"
 LOADADDR = ""
 UBOOT_SUFFIX = "bin"
-UBOOT_MAKE_TARGET = ""
```
otavio, line 41: Couldn't the variable here be dropped and use the default used in `imx-base.inc`?

otavio (changes requested):

MaxKrummenacher:
@zandrey 

> The quote you entered below is only in the "Cover Message" describing what was changed in _V2_. You're right - indeed the `imx-boot` recipe does this in `do_compile()` task after the binary is produced, it can be found in [imx-boot_1.0.bb#L129](https://github.com/Freescale/meta-freescale/blob/0a435d64e588843f6943d6e5723e5ecc12990aa2/recipes-bsp/imx-mkimage/imx-boot_1.0.bb#L129). But I guess this is not really relevant for this implementation.
> 
> Do you want me to put this note in a commit message as well?

In this thread there are several hints that u-boot produces flash.bin while imx-mkimage produces imx-boot. I just wanted to point out that this is not the case.
I don't think that this info is needed anywhere in a commit message or in a actual commit.

conf/machine/include/imx-base.inc
zandrey: Sure, would do.

conf/machine/include/imx8mn-evk.inc
zandrey: _imx-base.inc_ sets it as default to:
```python
UBOOT_MAKE_TARGET ?= "u-boot.${UBOOT_SUFFIX}"
```

There are 2 overrides next to this assignment: one for `mxs` and one for `mx8`, which would not apply here as they are filtered out.

Besides, there are several derivatives, which are setting the `UBOOT_MAKE_TARGET` to empty string, and I'm not sure that change to "all" would not break them.

And (as a cherry on top of the pie): `imx8qm-mek` and `imx8x-mek` have it assigned conditionally based on whether `optee` feature is enabled or not.

I'd prefer to keep the definition of the U-Boot build target as it is now, changing it to all as default in _imx-base.inc_ might break things.

-- andrey

zandrey:
Changes in _V3_:
- Fix permissions of `flash.bin` binary in deploy task
- Add symlink from deployed boot container to also point to it's original name
- Extend description with the comment about boot container renaming scheme
- Re-base PR branch on top of master

-- andrey

conf/machine/include/imx8mn-evk.inc
otavio: That's fine. I want to clean it later but we ought to have something merged first.

otavio (approved):

otavio:
@zandrey please rebase this on top of current master.

zandrey:
> 
> 
> @zandrey please rebase this on top of current master.

@otavio Done. No changes in this update, only rebase on top of current `master`.

PR #693, opened 2021-03-18
optee-os: Cleanup build options

thochstein:
- Move log level variables to EXTRA_OEMAKE to simplify overriding
- Set LDFLAGS and CFLAGS globally, like normal
- Set -C in EXTRA_OEMAKE
- Drop do_compile after all of the above
- Drop un-necessary override of B variable
- Simplify creation of tee.bin link in DEPLOYDIR

### DIFF
--- a/recipes-security/optee-imx/optee-os_3.10.0.imx.bb
+++ b/recipes-security/optee-imx/optee-os_3.10.0.imx.bb
@@ -17,7 +17,6 @@ SRC_URI = "\
 SRCREV = "a991c90475bb1c715651e5fe27f7f32cbe61aef9"
 
 S = "${WORKDIR}/git"
-B = "${WORKDIR}/build.${PLATFORM_FLAVOR}"
 
 inherit deploy python3native autotools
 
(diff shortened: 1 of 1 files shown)


recipes-security/optee-imx/optee-os_3.10.0.imx.bb
```diff
@@ -51,30 +50,25 @@ EXTRA_OEMAKE = " \
...
 
-do_compile () {
-    unset LDFLAGS
-    export CFLAGS="${CFLAGS} --sysroot=${STAGING_DIR_HOST}"
-    oe_runmake -C ${S} all CFG_TEE_TA_LOG_LEVEL=0 CFG_TEE_CORE_LOG_LEVEL=0
-}
-
+LDFLAGS = ""
+CFLAGS += "--sysroot=${STAGING_DIR_HOST}"
+CXXFLAGS += "--sysroot=${STAGING_DIR_HOST}"
```
otavio, line 60: Is those two need? I believe this is the default in toolchain arguments.

thochstein: I think so. I think the problem is the build files use `HOST_PREFIX` instead of using the tool variables where `--sysroot` resides. I will look into fixing that root issue.

thochstein: Looks non-trivial to rework the optee build files for this.

otavio (approved):

otavio:
This is a step in the right direction so I am merging it now.

thochstein:
This commit is not on `master`. Did history get overwritten?

otavio:
I added it back; it was my mistake when I removed the reworked bootloader change which I sent by mistake. We are good now.

Thanks for reporting.

PR #696, opened 2021-03-19
Rework bootloader choosing logic

otavio:
We need to make it more obvious what is in use, when we support both bootloaders and for it we are now using the _pn suffix for both.

This also fixes the WIC_FILE depending on the choice otherwise image build fails.

### DIFF
--- a/conf/machine/imx23evk.conf
+++ b/conf/machine/imx23evk.conf
@@ -9,6 +9,13 @@ MACHINEOVERRIDES =. "mxs:mx23:"
 include conf/machine/include/imx-base.inc
 include conf/machine/include/tune-arm926ejs.inc
 
+# This machine is not supported by u-boot-imx as it is not tested by NXP on this
+# board. So we force it to use u-boot-fslc which is based on mainline here.
+IMX_DEFAULT_BOOTLOADER = "u-boot-fslc"
+
+UBOOT_MAKE_TARGET = "u-boot.sb"
+UBOOT_SUFFIX = "sb"
+
 UBOOT_MACHINE = "mx23evk_config"
 
 KERNEL_DEVICETREE = "imx23-evk.dtb"
--- a/conf/machine/imx51evk.conf
+++ b/conf/machine/imx51evk.conf
@@ -11,4 +11,11 @@ include conf/machine/include/tune-cortexa8.inc
 
 KERNEL_DEVICETREE = "imx51-babbage.dtb"
 
+# This machine is not supported by u-boot-imx as it is not tested by NXP on this
+# board. So we force it to use u-boot-fslc which is based on mainline here.
+IMX_DEFAULT_BOOTLOADER = "u-boot-fslc"
+
+UBOOT_MAKE_TARGET = "u-boot.imx"
+UBOOT_SUFFIX = "imx"
+
 UBOOT_MACHINE = "mx51evk_config"
--- a/conf/machine/imx51evk.conf
+++ b/conf/machine/imx51evk.conf
@@ -9,9 +9,6 @@ MACHINEOVERRIDES =. "mx5:mx51:"
 include conf/machine/include/imx-base.inc
 include conf/machine/include/tune-cortexa8.inc
 
-# Disable perf features as 2.6.35.3 fails to build otherwise
-PERF_FEATURES_ENABLE = ""
-
 KERNEL_DEVICETREE = "imx51-babbage.dtb"
 
 UBOOT_MACHINE = "mx51evk_config"
--- a/conf/machine/imx6qdlsabreauto.conf
+++ b/conf/machine/imx6qdlsabreauto.conf
@@ -40,9 +40,9 @@ UBOOT_SUFFIX = "img"
 SPL_BINARY = "SPL"
 WKS_FILE = "imx-uboot-spl-bootpart.wks.in"
 
-# The fsl distro uses u-boot-imx which does not provide unified functionality
-# for dl/q/qp SoC variants. Change the defconfig to the targeted SoC variant.
-UBOOT_MACHINE_fsl = "mx6qsabreauto_defconfig"
+# The u-boot-imx does not provide unified functionality for DL/Q/QP SoC
+# variants. Change the defconfig to the targeted SoC variant.
+UBOOT_MACHINE_pn-u-boot-imx = "mx6qsabreauto_defconfig"
 
 SERIAL_CONSOLES = "115200;ttymxc3"
 
--- a/conf/machine/imx6qdlsabreauto.conf
+++ b/conf/machine/imx6qdlsabreauto.conf
@@ -34,12 +34,11 @@ KERNEL_DEVICETREE_use-mainline-bsp = " \
 
 UBOOT_MACHINE ?= "mx6sabreauto_defconfig"
 
-# These u-boot variables default to the correct settings for u-boot-imx.
-# Set overrides for u-boot-fslc.
-UBOOT_MAKE_TARGET_pn-u-boot-fslc = "all"
-UBOOT_SUFFIX_pn-u-boot-fslc = "img"
-SPL_BINARY_pn-u-boot-fslc = "SPL"
-WKS_FILE_pn-u-boot-fslc = "imx-uboot-spl-bootpart.wks.in"
+# Use fslc u-boot by default. See also imx-base.inc.
+UBOOT_MAKE_TARGET = "all"
+UBOOT_SUFFIX = "img"
+SPL_BINARY = "SPL"
+WKS_FILE = "imx-uboot-spl-bootpart.wks.in"
 
 # The fsl distro uses u-boot-imx which does not provide unified functionality
 # for dl/q/qp SoC variants. Change the defconfig to the targeted SoC variant.
--- a/conf/machine/imx6qdlsabreauto.conf
+++ b/conf/machine/imx6qdlsabreauto.conf
@@ -32,17 +32,25 @@ KERNEL_DEVICETREE_use-mainline-bsp = " \
     imx6dl-sabreauto.dtb \
 "
 
-UBOOT_MACHINE ?= "mx6sabreauto_defconfig"
+### u-boot-fslc settings ###
 
-# Use fslc u-boot by default. See also imx-base.inc.
-UBOOT_MAKE_TARGET = "all"
-UBOOT_SUFFIX = "img"
-SPL_BINARY = "SPL"
-WKS_FILE = "imx-uboot-spl-bootpart.wks.in"
+SPL_BINARY_pn-u-boot-fslc = "SPL"
+UBOOT_MACHINE_pn-u-boot-fslc ?= "mx6sabreauto_defconfig"
+UBOOT_SUFFIX_pn-u-boot-fslc = "img"
+
+### u-boot-imx settings ###
 
 # The u-boot-imx does not provide unified functionality for DL/Q/QP SoC
 # variants. Change the defconfig to the targeted SoC variant.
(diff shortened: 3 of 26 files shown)


otavio:
@thochstein @zandrey @fbertux 

Please take a look at this PR. I believe this move us to the right direction and allow for a much easier way to set the bootloader.

For the converted machines, we now can do:

```
IMX_DEFAULT_BOOTLOADER = 'u-boot-imx'
```

and it does the right thing.

@dv1 this PR should fix the error you were getting during the image generation.

dv1:
Indeed, it works now, thanks!

conf/machine/imx6qdlsabreauto.conf
```diff
@@ -32,18 +32,25 @@ KERNEL_DEVICETREE_use-mainline-bsp = " \
...
+UBOOT_MAKE_TARGET_pn-u-boot-imx = "u-boot.imx"
+UBOOT_SUFFIX_pn-u-boot-imx = "imx"
 
-# The fsl distro uses u-boot-imx which does not provide unified functionality
-# for dl/q/qp SoC variants. Change the defconfig to the targeted SoC variant.
-UBOOT_MACHINE_fsl = "mx6qsabreauto_defconfig"
+WKS_FILE = " \
+    ${@bb.utils.contains('IMX_DEFAULT_BOOTLOADER', \
+                         'u-boot-fslc', 'imx-uboot-spl-bootpart.wks.in', \
+                                        'imx-uboot-bootpart.wks.in', d)}"
```
thochstein, line 53: Can you use `WKS_FILE_pn-` instead?

conf/machine/imx6qdlsabresd.conf
```diff
@@ -41,18 +41,27 @@ KERNEL_DEVICETREE_use-mainline-bsp = " \
     imx6dl-sabresd.dtb \
 "
 
-UBOOT_MACHINE ?= "mx6sabresd_defconfig"
+WKS_FILE = "imx-uboot-spl-bootpart.wks.in"
```
thochstein, line 44: Drop this line.

conf/machine/include/imx-base.inc
```diff
@@ -32,13 +32,11 @@ PREFERRED_PROVIDER_virtual/bootloader ??= "${IMX_DEFAULT_BOOTLOADER}"
...
-# Set specific make target and binary suffix
+# UBOOT_BINARY is used inside the wks files to dynamically find the required
+# U-Boot file.
 UBOOT_BINARY ?= "u-boot.${UBOOT_SUFFIX}"
-UBOOT_MAKE_TARGET ?= "u-boot.${UBOOT_SUFFIX}"
-UBOOT_MAKE_TARGET_mxs ?= "u-boot.sb"
-UBOOT_MAKE_TARGET_mx8 ?= ""
 
-UBOOT_SUFFIX ?= "imx"
+UBOOT_SUFFIX ?= "${UBOOT_SUFFIX_pn-${IMX_DEFAULT_BOOTLOADER}}"
```
thochstein, line 39: This looks like a tricky assignment. Can you explain it? Add a comment?

conf/machine/imx6qdlsabreauto.conf
otavio: You can't. It is used in another context and not inside the U-Boot recipe. This was the root cause of @dv1 error, in fact.

conf/machine/imx6qdlsabresd.conf
otavio: Thanks, good catch. Fixed.

conf/machine/imx6qdlsabreauto.conf
thochstein: I see, perhaps you can make a comment then so it's clear why it doesn't follow the same pattern?

otavio: I added some other comments that try to make things clearer, but I don't see how it'd help as it is done this way all around the layer, so people would need to comment on many recipes.

conf/machine/include/imx-base.inc
otavio: I did. Please take a look now.

otavio:
Hello all, we checked build on:

- [x] imx23evk
- [x] imx25pdk
- [x] imx28evk
- [x] imx51evk
- [x] imx53ard
- [x] imx53qsb
- [x] imx6qdlsabreauto
- [x] imx6qdlsabresd
- [x] imx6slevk
- [x] imx6sllevk
- [x] imx6sxsabreauto
- [x] imx6sxsabresd
- [x] imx6ulevk
- [x] imx6ullevk
- [x] imx6ulz-14x14-evk
- [x] imx7dsabresd
- [x] imx7ulpevk
- [x] imx8dxl-evk
- [x] imx8dx-mek
- [x] imx8mm-ddr4-evk
- [x] imx8mm-lpddr4-evk
- [x] imx8mn-ddr4-evk
- [x] imx8mn-lpddr4-evk
- [x] imx8mp-evk
- [x] imx8mq-evk
- [x] imx8qm-mek
- [x] imx8qxp-mek

thochstein:
Which distros?

otavio:
> Which distros?

Our internal one. Now checking the mainline Linux build with it. For this, the distro should not have much impact as we're messing up with U-Boot. I used `core-image-minmal` to confirm it builds fine.

Can you do some tests at your end as well?

thochstein:
> Can you do some tests at your end as well?

Will do.

conf/machine/imx23evk.conf
```diff
@@ -9,6 +9,13 @@ MACHINEOVERRIDES =. "mxs:mx23:"
 include conf/machine/include/imx-base.inc
 include conf/machine/include/tune-arm926ejs.inc
 
+UBOOT_MAKE_TARGET = "u-boot.sb"
+UBOOT_SUFFIX = "sb"
+
+# This machine is not supported by u-boot-imx as it is not tested by NXP on this
+# board. So we force it to use u-boot-fslc which is based on mainline here.
+IMX_DEFAULT_BOOTLOADER = "u-boot-fslc"
```
thochstein, line 17: Doesn't it make more sense to place this above `UBOOT_MAKE_TARGET`?

thochstein: Oh, I guess that is what you do with the next files.

conf/machine/imx51evk.conf
```diff
@@ -9,9 +9,13 @@ MACHINEOVERRIDES =. "mx5:mx51:"
...
-PERF_FEATURES_ENABLE = ""
-
 KERNEL_DEVICETREE = "imx51-babbage.dtb"
 
+# This machine is not supported by u-boot-imx as it is not tested by NXP on this
+# board. So we force it to use u-boot-fslc which is based on mainline here.
+IMX_DEFAULT_BOOTLOADER = "u-boot-fslc"
+
+UBOOT_MAKE_TARGET = "u-boot.imx"
+UBOOT_SUFFIX = "imx"
```
thochstein, line 19: These are correct for `u-boot-fslc`?

conf/machine/imx6qdlsabresd.conf
```diff
@@ -41,18 +41,25 @@ KERNEL_DEVICETREE_use-mainline-bsp = " \
...
+### u-boot-fslc settings ###
 
-# These u-boot variables default to the correct settings for u-boot-imx.
-# Set overrides for u-boot-fslc.
-UBOOT_MAKE_TARGET_pn-u-boot-fslc = "all"
-UBOOT_SUFFIX_pn-u-boot-fslc = "img"
 SPL_BINARY_pn-u-boot-fslc = "SPL"
-WKS_FILE_pn-u-boot-fslc = "imx-uboot-spl-bootpart.wks.in"
+UBOOT_MACHINE_pn-u-boot-fslc ?= "mx6sabresd_defconfig"
+UBOOT_SUFFIX_pn-u-boot-fslc = "img"
```
thochstein, line 48: Should the ?= assignment be consistent for all these variables?

conf/machine/include/imx8x-mek.inc
```diff
@@ -54,6 +54,11 @@ UBOOT_CONFIG[fspi] = "${UBOOT_CONFIG_BASENAME}_fspi_defconfig"
 
 IMX_BOOT_SEEK = "32"
 
+# This machine is not supported by u-boot-fslc, so we force it to use
+# u-boot-imx here.
+IMX_DEFAULT_BOOTLOADER = "u-boot-imx"
```
thochstein, line 59: Why no overrides here like the other `imx8*.inc`?

conf/machine/imx51evk.conf
otavio: Those are; this does not use SPL so the image file is `u-boot.imx`.

conf/machine/imx6qdlsabresd.conf
otavio: Good question. Do we want it to be overridable in `local.conf`?

conf/machine/include/imx8mm-evk.inc
```diff
@@ -22,6 +22,11 @@ KERNEL_DEVICETREE_append_use-nxp-bsp = " \
 "
 UBOOT_DTB_NAME = "${KERNEL_DEVICETREE_BASENAME}.dtb"
 
+IMX_DEFAULT_BOOTLOADER_use-nxp-bsp = "u-boot-imx"
+IMX_DEFAULT_BOOTLOADER_use-mainline-bsp = "u-boot-imx"
```
zandrey, line 26: I might be missing a big picture here, but this line (and the same for `imx8mn`) has caught my attention: why do we enforce the _u-boot-imx_ for mainline BSP?

I believe that current `u-boot-fslc` does have a support for those machines, so why not just use it?

otavio: Because it is wrong :-) Good catch.

conf/machine/include/imx8x-mek.inc
otavio: This is not supported by `u-boot-fslc` yet so we force it to use the same in any case.

otavio:
I believe all issues have been addressed. Please take a look.

conf/machine/imx6qdlsabreauto.conf
thochstein: This is all I had in mind:

```
# For WKS_FILE we can't use _pn-u-boot-* override like above since WKS_FILE is used outside of u-boot context.
```

I guess that fix gets kind of buried.

thochstein:
> > Can you do some tests at your end as well?
> 
> Will do.

Having a build problem again with mesa, `imx8mn-lpddr4-evk`...

```
| ../mesa-21.0.0/meson.build:21:0: ERROR: Options "swrast" are not in allowed choices: "auto, i915, i965, r100, r200, nouveau"
```

otavio:
Yes, newer mesa dropped swrast. Please open an issue so we can look at this together.

thochstein:
A problem I'm running into with this rework is my local machines are not defaulting to the same values.

```
cp: cannot stat '/home/bamboo/build/5.10-hn-full/fsl-imx-internal-fb/temp_build_dir/build_fsl-imx-internal-fb/tmp/work/imx6ul7d-poky-linux-gnueabi/u-boot-imx/1_2020.04-r0/git/mx6ul_14x14_evk_optee_config/u-boot.u-boot-imx': No such file or directory
```

We fixed it locally:

```
UBOOT_MAKE_TARGET_pn-u-boot-imx = "u-boot.imx"
UBOOT_SUFFIX_pn-u-boot-imx = "imx"
```

Is it a bug?

otavio:
Yes, please open an issue so we look at it.

PR #701, opened 2021-03-29
Add 8M Plus EVK with DDR4

thochstein:


otavio (changes requested):
Please add the ddr4 variant here as well.

thochstein:
> Please add the ddr4 variant here as well.

Do you mean you want me to squash the 2 commits?

otavio:
When I saw it, it didn't have the ddr4 variant. I think it is fine in two commits.

angolini (approved):

PR #708, opened 2021-04-07
conf: imx8mp-evk.inc: correct u-boot make target

zandrey:
U-Boot make target set to empty string does not expand correctly for mainline BSP build, which append make target with 'flash.bin' entry.
This causes the build to fail, since the u-boot.bin target is not produced when only 'flash.bin' is present in the make target list.

Define the default make target to 'all', which allows proper build target expansion.

Fixes: 823a97cf ("imx8mp-lpddr4-evk: Rework 8M Plus EVK config file")
Cc: @thochstein


otavio (approved):
Have you tried with u-boot-imx and u-boot-fslc?

thochstein (approved):

PR #712, opened 2021-04-08
Add 8 DXL EVK with DDR3

thochstein:


otavio (approved):
@zandrey mind to take a look here as well?

conf/machine/include/imx8dxl-evk.inc
```diff
@@ -57,8 +49,14 @@ IMXBOOT_TARGETS_SD = \
 IMXBOOT_TARGETS_FSPI = \
     "${@bb.utils.contains('MACHINE_FEATURES', 'optee', 'flash_spl_flexspi', \
                                                        'flash_flexspi', d)}"
+IMXBOOT_TARGETS_NAND = \
+    "${@bb.utils.contains('MACHINE_FEATURES', 'optee', 'flash_spl_nand', \
+                                                       'flash_nand flash_nand_fw', d)}"
```
zandrey, line 54: @thochstein

Just for my understanding: `optee` feature is checked here, but is not added to `MACHINE_FEATURES` above. Is it in _WIP_ state with the support planned for this derivative?

thochstein: We supported `optee` in the release, and I had it in the initial PR for the machine, but, as I remember, @otavio asked me to remove it. I'll let him explain further.

otavio: We have it not enabled by default as the mainline doesn't support it, and to add it is a matter of adding it to the machine features. I think we have this more homogeneous around the machines we support to allow a more predictable experience.

PR #720, opened 2021-04-13
weston-init: Rework uncomment function logic

Ossanes:
Fix error "Commented setting '#use-g2d=1' not found"
when building for some imx machines.

Signed-off-by: Vinicius Aquino <voa.aquino@gmail.com>

### DIFF
--- a/recipes-graphics/wayland/weston-init.bbappend
+++ b/recipes-graphics/wayland/weston-init.bbappend
@@ -40,7 +40,7 @@ INI_UNCOMMENT_ASSIGNMENTS_append_mx8mq = " \
 "
 
 uncomment() {
-    if ! (grep "^#$1" $2); then
+    if ! grep -q "^#$1" $2 && ! grep -q "^$1" $2; then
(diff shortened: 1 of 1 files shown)


otavio (approved):

recipes-graphics/wayland/weston-init.bbappend
```diff
@@ -40,7 +40,7 @@ INI_UNCOMMENT_ASSIGNMENTS_append_mx8mq = " \
 "
 
 uncomment() {
-    if ! (grep "^#$1" $2); then
+    if ! grep -q "^#$1" $2 && ! grep -q "^$1" $2; then
```
thochstein, line 43: I think if the line is already uncommented there should be a `bbdebug` and return.

otavio: We didn't do that as some (e.g., i.MX6QDL) will run this twice. So the code guarantee it is as intended (uncommented).

PR #722, opened 2021-04-14
imx53qsb: U-Boot fixup

twoerner:
A change was made in oe-core, that forced a change in the BSP layer, that
highlighted an issue with the U-Boot variable for this machine.

Build tested for imx53qsb.

### DIFF
--- a/conf/machine/imx53qsb.conf
+++ b/conf/machine/imx53qsb.conf
@@ -12,8 +12,8 @@ include conf/machine/include/tune-cortexa8.inc
 KERNEL_DEVICETREE = "imx53-qsb.dtb imx53-qsrb.dtb"
 
 # This machine is not supported by u-boot-imx as it is not tested by NXP on this
-# board. So we force it to use u-boot-fslc which is based on mainline here.
-IMX_DEFAULT_BOOTLOADER = "u-boot-fslc"
+# board. Use mainline instead.
+IMX_DEFAULT_BOOTLOADER = "u-boot"
 
 UBOOT_MAKE_TARGET = "u-boot.imx"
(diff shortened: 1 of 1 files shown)


otavio (changes requested):
Same comments as made to olinxino change.

otavio (commented):
Why do you prefer to use u-boot instead of u-boot-fslc?

twoerner:
> Why do you prefer to use u-boot instead of u-boot-fslc?

That's a good question, and one I'll have to dig into somewhat. I don't think I'm explicitly setting the PREFERRED_PROVIDER in my settings, but that's what I'm getting:

```
$ bitbake core-image-minimal -e | less
…
# $PREFERRED_PROVIDER_virtual/bootloader
#   set /opt/oe/configs/z/build-master/imx53qsb/layers/meta-freescale/conf/machine/include/imx-base.inc:79
#     [_defaultval] "${IMX_DEFAULT_BOOTLOADER}"
PREFERRED_PROVIDER_virtual/bootloader="u-boot"
```

I'll try a "fresh" build to make sure I'm not getting some sort of interference, but I _think_ I'm doing a basic build with no changes from the defaults and ending up with a "u-boot" binary when U-Boot is done building and not a "u-boot-fslc" binary.

otavio:
We set it for you based on the default bootloader. Check `imx-base.inc`.

twoerner:
Seems I've got a mystery on my hands. A completely fresh build produces:

```
PREFERRED_PROVIDER_virtual/bootloader="u-boot-fslc"
```

And all my builds for however long have been using u-boot as the PREFERRED_PROVIDER for the bootloader… but I'm not sure why.

Sorry for the noise!

Issue #704, opened 2021-04-01
Bootloader choosing logic breaks compatibility for existing machines


thochstein:
A problem I'm running into after https://github.com/Freescale/meta-freescale/issues/696  is my local machines are not defaulting to the same values.

```
cp: cannot stat '/home/bamboo/build/5.10-hn-full/fsl-imx-internal-fb/temp_build_dir/build_fsl-imx-internal-fb/tmp/work/imx6ul7d-poky-linux-gnueabi/u-boot-imx/1_2020.04-r0/git/mx6ul_14x14_evk_optee_config/u-boot.u-boot-imx': No such file or directory
```

We fixed it locally:

```
UBOOT_MAKE_TARGET_pn-u-boot-imx = "u-boot.imx"
UBOOT_SUFFIX_pn-u-boot-imx = "imx"
```

But it would be nice to fix this so we don't break compatibility.

_Originally posted by @thochstein in https://github.com/Freescale/meta-freescale/issues/696#issuecomment-811933523_

otavio:
I've successfully built the `u-boot-fslc`and `u-boot-imx` for ìmx6ulevk`.

You can see this below:
```
% ls tmp/*/deploy/images/*
tmp/fslc/deploy/images/imx6ulevk:
SPL
SPL-imx6ulevk
SPL-imx6ulevk-sd
SPL-imx6ulevk-v2020.10+gitAUTOINC+80c23498b4-r0-sd-v2020.10+gitAUTOINC+80c23498b4-r0
SPL-sd
u-boot.img
u-boot-fslc-initial-env-imx6ulevk-sd
u-boot-fslc-initial-env-imx6ulevk-sd-v2020.10+gitAUTOINC+80c23498b4-r0
u-boot-fslc-initial-env-sd
u-boot-imx6ulevk.img
u-boot-imx6ulevk.img-sd
u-boot-sd-v2020.10+gitAUTOINC+80c23498b4-r0.img
u-boot.img-sd

tmp/imx/deploy/images/imx6ulevk:
u-boot.imx                           u-boot-imx-initial-env-imx6ulevk-sd-2020.04-r0
u-boot-imx6ulevk.imx                 u-boot-imx-initial-env-sd
u-boot-imx6ulevk.imx-sd              u-boot-sd-2020.04-r0.imx
u-boot-imx-initial-env-imx6ulevk-sd  u-boot.imx-sd
```

[...]

PR #723, opened 2021-04-14
imx-base.inc: Provide a default value for SPL_BINARY

otavio:
This fixes the build case when machine does not use SPL and does not set
it. We couldn't use dynamic expansion or we trigger a recursive
expansion so we opted to use an utility function to do the proper logic.

Fixes: f98059fe ("imx-base.inc: Allow use of SPL_BINARY variable inside wic images")


otavio:
@thochstein please check this. This ought to fix the error with imx6sllevk.

thochstein:
This looks good, I wonder if you can do the same for `UBOOT_SUFFIX` and `UBOOT_MAKE_TARGET` to make those backwards compatible (see issue #704).

otavio:
This is not exactly the same thing. We are exporting this for use outside the u-boot recipe (on the wks file) and that's the reason we had to handle it in a special way.

PR #732, opened 2021-04-16
u-boot-imx-tools: Drop i.MX specific tools

otavio:
We are building same tools as the original U-Boot thus we should rely on
them.

This is critical as native tools cannot vary between machines as there
are no clear reason to convert this to be a cross tool as we build no
machine specific code.


thochstein:
This builds tools from https://source.codeaurora.org/external/imx/uboot-imx. Are you sure there is nothing custom there? Does it build just in the folder `tools`. When I check there it looks like it only contains an update to `freescale.bmp`.

thochstein:
`u-boot-tools` does in fact fail to build for me, distro `fsl-xwayland`.

PR #750, opened 2021-04-26
NXP release 5.4.70-2.3.2

thochstein:
Start 5.4.70-2.3.2 upstreaming with EULA, SCR, kernel


otavio:
@zandrey please take a look if we need to update our branch, please.

zandrey:
> 
> 
> @zandrey please take a look if we need to update our branch, please.

Sure, I had a look and there are 48 patches out of 49 that need to be applied to `5.4-2.3.x-imx`, one patch was cherry-picked from upstream which is already present in `linux-fslc` via stable update.

I've cherry-picked all of remaining 48 patches, and building the tree for both `arm32` and `aarch64` configs. If it turns our to be OK - I'll prepare a PR for this upgrade in `linux-fslc` tree.

-- andrey

PR #760, opened 2021-04-29
Fixes for weston and gpu drivers

dv1:
This fixes two issues that caused weston to not work. The second one is particularly noteworthy since it partially reverts a previous commit. @thochstein , can you take a look as well, since that previous commit was yours? When I used strace on weston to find out what's wrong, I noticed attempts at accessing `/dev/galcore`, which just wasn't present. Adding the GPU kernel module adds that device node, fixing the second issue.


otavio (approved):

thochstein:
Thanks, @dv1! I dropped the change since I could find no justification for it internally and was suspicious. Sorry it bit you!

PR #767, opened 2021-05-01
firmware-imx: Upgrade to 8.10.1

thochstein:


otavio (approved):

zandrey:
@thochstein I see that the `8.11` archive is already available. Is there ant reason you did not switch to it?

I tried it our with the `mx8mm`, `mx8mn` and `mx8mp` - they all boot fine with the mainline. As I'm now working on the update for new _U-Boot_ version, I was planning to bump the firmware package to `8.11`. But seeing this PR - I'm not sure if I would have any issues with it... 

Can you please clarify if it would be safe to use `8.11` version or is it better to stay with `8.10.1`?

Thanks a lot!

-- andrey

thochstein:
Hi @zandrey , I am focused on our 5.4.70-2.3.2 release which has 8.10.1. I don't know whether 8.11 would be good or problematic. I can look for advice internally if you want.

zandrey:
> 
> 
> Hi @zandrey , I am focused on our 5.4.70-2.3.2 release which has 8.10.1. I don't know whether 8.11 would be good or problematic.

Got it, thanks for your reply here! I can say from my side that at least DDR firmware seems to be OK, as I had it tested with U-Boot already. Not sure if all other firmware parts are operable though...

> I can look for advice internally if you want.

Would really appreciate any statement here, thanks a lot! 👍🏻 

-- andrey

thochstein:
@zandrey, so the best I can gather is there is some risk with 8.11 since that has only been tested with a later release using 5.10 kernel. Our China team is on holiday this week, so hopefully I'll hear more and will let you know.

zandrey:
@thochstein

> 
> 
> @zandrey, so the best I can gather is there is some risk with 8.11 since that has only been tested with a later release using 5.10 kernel.

Thanks a lot for the clarification note! I held the upgrade to `8.11` from the _U-Boot_ PR #768, since it was fully functional with updated `8.10.1` version. I guess this can wait till fully tested at NXP end and officially announced.

> Our China team is on holiday this week, so hopefully I'll hear more and will let you know.

Would be great to get any news here anyways, thanks! 👍🏻 

-- andrey

PR #768, opened 2021-05-03
U-Boot upgrade related changes (update to v2021.07-rc1 and layer adaptions)

zandrey:
Since the introduction of new _U-Boot_ version (past `v2021.04`), several major functional upgrades for `imx8m` family are done that require layer to be adapted.

Following functionality has been enabled in the the _U-Boot_ past `v2021.04` tag:
- _Distro boot_ option has been enabled an selected as a default boot option
- Binary boot image generation has been switched to use _binman_ instead of custom `flash.bin` target

It has been also found, that even though upstream `v2021.04` version contains several improvements targeted `imx8m` family - it does not provide a full support for above features across all derivatives in the family, hence the later `v2021.07-rc1` revision of _U-Boot_ is taken which contains the support level that is re-usable for all.

This requires modification to the layer, which this PR introduce:
- _Distro boot_ option require _Syslinux_ to be enabled, as it is considered to be preferred way of booting distro images.
- _binman_ produces now two binary images that should be populated onto the raw boot block of a boot device, does not use the custom `flash.bin` build target anymore, and uses fully-qualified DDR firmware names so the sanitation is not required anymore.

All those points are addressed in corresponding commit messages.

-- andrey

Cc: @thochstein 
Cc: @gibsson 
Cc: @MaxKrummenacher


conf/machine/include/imx-base.inc
```diff
@@ -403,6 +403,21 @@ PREFERRED_PROVIDER_virtual/kernel ??= "${IMX_DEFAULT_KERNEL}"
...
+# We opt-in for Syslinux, since it is designated as a preferred
+# distro boot mode according to the U-Boot documentation.
+UBOOT_EXTLINUX_use-mainline-bsp ?= "1"
+UBOOT_EXTLINUX_LABELS_use-mainline-bsp ?= "default"
+UBOOT_EXTLINUX_DEFAULT_LABEL_use-mainline-bsp ?= "i.MX8M Family EVK"
+
+UBOOT_EXTLINUX_FDT_default_use-mainline-bsp     ?= "../${KERNEL_DEVICETREE_BASENAME}.dtb"
+UBOOT_EXTLINUX_CONSOLE_default_use-mainline-bsp ?= "console=${console}"
+UBOOT_EXTLINUX_ROOT_default_use-mainline-bsp    ?= "root=/dev/mmcblk1p2"
+UBOOT_EXTLINUX_MENU_DESCRIPTION_default_use-mainline-bsp ?= "i.MX8M Family EVK"
```
otavio, line 419: I'd prefer this to be added to every machine as it is hard to come with a generic solution here.

otavio (changes requested):

conf/machine/include/imx-base.inc
zandrey: Sure, no problem! Funny enough - I had it originally duplicated in every machine include file, but decided to unify parts that were identical across. 😄 

I would move it back to respective machine include files and update this PR with _V2_.

zandrey:
Changes in _V2_:
- Drop Syslinux definitions from `imx-base.inc` file and duplicate those definitions in machine-specific include files

-- andrey

otavio (approved):

Issue #795, opened 2021-05-26
Mesa build is broken after the upstream upgrade to 21.1.1


zandrey:
Upstream OE commit 8579142183 ("mesa: update 21.0.3 -> 21.1.1") has bumped up the version of _Mesa_ to `21.1.1.`

This causes the patch [0002-dri-bring-back-use-flags-for-createImageWithModifier.patch](../blob/master/recipes-graphics/mesa/mesa/0002-dri-bring-back-use-flags-for-createImageWithModifier.patch) not to apply clean and build breaks.

Patches were merged with commit bfb9e699 ("mesa/etnaviv: Add patches to fix redraw"), which lists `Upstream-Status: Submitted [https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/7603]` in patches submitted into the layer.

The mentioned _Mesa_ PR however does not have the patch included in the layer, as it has been updated several times after the original submission, which eventually lead to several patches dropped (including the "offending" one).

I'm not sure if the commit bfb9e699 ("mesa/etnaviv: Add patches to fix redraw") applies to the new version, and should be either re-worked or reverted. Reverting the patch does solve the compilation issue, but I cannot test that the runtime is sane.

-- andrey

otavio:
@schnitzeltony is it possible for you to take a look at this?

schnitzeltony:
Yes I should do that because there were promising changes at the etnaviv area. Frankly spoken: I was waiting for this update. 
Bad news: Daily business makes it impossible for me to announce some schedule - hope this is something for weekend...

otavio:
Fixed in #797.

PR #797, opened 2021-05-31
mesa: Refresh patches for mesa 21.1.1

schnitzeltony:
* Addresses [1]
* all patches were taken from mesa's main branch and applied cleanly
* On X11/xfce first tests were performed without any patches: At least we do not see
  a mouse pointer which is similar behavior as seen when introducing first
  patch series.
* On X11/xfce moving windows is faster now / gtk-based applications still show
  tiny artifacts / Qt/QML based application run without issues / glmark2-es2
  results in 158 which is similar to mesa 21.0.3
* Could not test my weston images (use sddm as dm): All native recipes inheriting
  cmake_qt5 are broken due to MACHINE_SOCARCH_FILTER - will open an issue for
  that.

[1] https://github.com/Freescale/meta-freescale/issues/795

Signed-off-by: Andreas Müller <schnitzeltony@gmail.com>


otavio (approved):

otavio:
Thanks for looking at this.

PR #804, opened 2021-06-06
linux-fslc: defconfig: Select the Wifi ath10k sdio driver

fabioestevam:
Select the ath10k sdio driver so that Wifi can work by default on
boards like imx6dl-pico-pi with a QCA9377 chip.

Signed-off-by: Fabio Estevam <festevam@gmail.com>


fabioestevam:
I have sent this patch usptream:
https://www.spinics.net/lists/arm-kernel/msg899845.html

otavio:
Thanks for the patch. Do you mind sending this to our fork as well? so we keep it in sync with what we are using in the layer.

I'll merge it anyway.

fabioestevam:
Hi Otavio,

On Sun, Jun 6, 2021 at 2:48 PM Otavio Salvador ***@***.***> wrote:
>
> Thanks for the patch. Do you mind sending this to our fork as well? so we keep it in sync with what we are using in the layer.

Sure , will submit it the patch against linux-fslc as well.

> I'll merge it anyway.

Thanks!

PR #814, opened 2021-06-17
Consolidated Kernel update (v5.4.126 + v5.10.44 + 5.12.11)

zandrey:
Kernel branches were updated up to and including following versions for recipes from stable korg:
- `linux-fslc-imx`: _v5.4.126_
- `linux-fslc-lts`: _v5.10.44_
- `linux-fslc`: _v5.12.11_

Update recipe `SRCREV` to point to those versions now.

Upstream commits and resolved conflicts are recorded in corresponding recipe commit messages.

-- andrey

### DIFF
--- a/recipes-kernel/linux/linux-fslc-imx_5.4.bb
+++ b/recipes-kernel/linux/linux-fslc-imx_5.4.bb
@@ -28,7 +28,7 @@ Latest stable Kernel patchlevel is applied and maintained by Community."
 # ------------------------------------------------------------------------------
 # 1. Stable (tag or SHA(s))
 # ------------------------------------------------------------------------------
-#    tag: v5.4.125
+#    tag: v5.4.126
 #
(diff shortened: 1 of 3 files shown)


otavio:
@zandrey can you handle the backport manually?

zandrey:
> 
> 
> @zandrey can you handle the backport manually?

Sure, would do!

-- andrey

zandrey:
@otavio 

It looks like that kernel updates cannot be applied as backport to `hardknott`, and should be applied separately. I'll have a look at it, and have it uprev'ed to the latest kernel in the _linux-fslc_.

I think it does not make sense to include the kernel PRs in GitHub actions, so I wanted to ask you if this can be excluded? I guess it would always be different in between branches, since there are several different kernel branches that are used in between Yocto releases, and backports are always be a rather manual update.

-- andrey

PR #818, opened 2021-06-29
License and kernel updates for NXP 5.10.35-2.0.0 release

thochstein:


recipes-kernel/linux/linux-imx_5.10.bb
```diff
@@ -13,20 +13,20 @@ i.MX Family Reference Boards. It includes support for many IPs such as GPU, VPU
...
-LIC_FILES_CHKSUM = "file://COPYING;md5=bbea815ee2795b2f4230826c0c6b8814"
+LIC_FILES_CHKSUM = "file://COPYING;md5=6bc538ed5bd9a7fc9398086aedcd7e46"
 
 DEPENDS += "lzop-native bc-native"
 
-SRCBRANCH = "imx_5.4.70_2.3.0"
-LOCALVERSION = "-2.3.2"
-SRCREV = "5f2220e87dd8a8fb86f524ea7945ab1d4155bc73"
+SRCBRANCH = "lf-5.10.y"
+LOCALVERSION = "-lts-5.10.y"
```
otavio, line 21: The local version is better to point to the GA release it is using.

otavio (changes requested):

recipes-kernel/linux/linux-imx_5.10.bb
thochstein: Okay, so to follow the last change, set it to `-2.0.0`?

otavio (approved):

otavio:
@thochstein the change looks good; how this has been tested?

thochstein:
I only built it for 8QM and 6Q. How do you want me to test? I can easily do a sanity check for 8M Quad EVK and 6Q SabreSD.

otavio:
This would be great; just to avoid merging something which wouldn't boot.

recipes-kernel/linux/linux-imx_5.10.bb
```diff
@@ -13,20 +13,20 @@ i.MX Family Reference Boards. It includes support for many IPs such as GPU, VPU
...
+LIC_FILES_CHKSUM = "file://COPYING;md5=6bc538ed5bd9a7fc9398086aedcd7e46"
 
 DEPENDS += "lzop-native bc-native"
 
-SRCBRANCH = "imx_5.4.70_2.3.0"
-LOCALVERSION = "-2.3.2"
-SRCREV = "5f2220e87dd8a8fb86f524ea7945ab1d4155bc73"
+SRCBRANCH = "lf-5.10.y"
+LOCALVERSION = "-2.0.0"
+SRCREV = "ef3f2cfc6010c13feb40cfb7fd7490832cf86f45"
```
zandrey, line 22: It is a bit odd to see that version goes backwards here: from `2.3.0` back to `2.0.0`.

I known that it is rather a "cosmetic" point, but I hope that people do not rely much on the output of `uname` in this case.

thochstein: Would `-imx-5.10.35-2.0.0` be better? What about `linux-fslc-imx`, it currently follows the same scheme? Should I update it like this?

`linux-fslc-imx_5.4.bb:LOCALVERSION = "-fslc-imx-5.4.70-2.3.0"`

zandrey: @thochstein 

I would not mind doing this change.

I can include it in the kernel PR I'm currently working on for `linux-fslc-imx`, and change _LOCALVERSION_ from `-imx-5.4.70-2.3.0` to `-fslc-imx-5.4.70-2.3.0` with consecutive change to `-imx-5.10.35-2.0.0` once kernel branch would be created.

@otavio what do you think about this?

-- andrey

otavio: `5.10.35-2.0.0` is shorten.

thochstein: So @zandrey should update fslc version to `-fslc-5.10.35-2.0.0`?

otavio: I believe `-5.4.70-2.3.0` is enough as this is the BSP version it is based on.

zandrey: @otavio OK, no problem!

I'll push one more commit on the upcoming kernel PR to change _LOCALVERSION_ to `-5.4.70-2.3.0`

@thochstein I believe you would then use the `-5.10.35-2.0.0` for _LOCALVERSION_, right? In this case we would have both recipes aligned.

-- andrey

zandrey: Done, please see a1ff57ba ("linux-fslc-imx: sanitize local version") in #820

thochstein:
> This would be great; just to avoid merging something which wouldn't boot.

I booted `core-image-minimal` on `imx8mq-evk` and `imx6qdlsabresd`.

recipes-bsp/imx-atf/imx-atf_2.4.bb
```diff
@@ -27,6 +27,8 @@ EXTRA_OEMAKE += " \
 
 BUILD_OPTEE = "${@bb.utils.contains('MACHINE_FEATURES', 'optee', 'true', 'false', d)}"
 
+CFLAGS_remove_mx8mq = "-O2"
```
thochstein, line 30: I've got a ticket open for a better fix if you want to wait for this.

thochstein: Actually, looks like this fix is fine, the problem is the 8M Quad has limited OCRAM. I updated the commit message.

pberginkonsult:
Tested to boot linux-imx on a imx8mm_lpddr4_evk together with u-boot-imx 2021.04.
Are there any blocking issues for merging this?

PR #829, opened 2021-07-20
udev: Allow normal user to use GPU drivers

thochstein:
A normal user account cannot run the GPU drivers. Fix this by giving
permission to the galcore device.


recipes-core/udev/udev-rules-imx/10-imx.rules
```diff
@@ -20,4 +20,4 @@ KERNEL=="mxc_vpu",  MODE="0666"
 SUBSYSTEM=="video", MODE="0660"
 KERNEL=="fb[0-9]",  MODE="0660", GROUP="video"
 KERNEL=="gsl_kmod", MODE="0660", GROUP="video"
-KERNEL=="galcore",  MODE="0660", GROUP="video"
+KERNEL=="galcore",  MODE="0666", GROUP="video"
```
otavio, line 23: This change is conceptually wrong as this allows any user to write to the GPU device.

The current setting allows any user which _belongs_ to the `video` group, the access to the GPU device. This allows a more granular control about the security of users inside the system.

otavio (changes requested):

recipes-core/udev/udev-rules-imx/10-imx.rules
thochstein: Makes sense.

PR #835, opened 2021-07-28
isp-imx, basler-camera

thochstein:


recipes-bsp/isp-imx/basler-camera_4.2.2.13.0.bb
```diff
@@ -2,14 +2,14 @@
```
otavio, line 2: Changlog please.

otavio (changes requested):

recipes-bsp/isp-imx/basler-camera_4.2.2.13.0.bb
thochstein: The Changelog I can see is just 'drop from VSI', no detail.

PR #841, opened 2021-08-10
gtk+3: Fix incorrect PACKAGECONFIG_remove

sfoster1:
Overridding PACKAGECONFIGs with PACKAGECONFIG_remove for a
different package than the current context requires the spelling
PACKAGECONFIG_pn-packagename (where packagename is replaced with
the packagename, but pn is a literal pn). The gtk+3 bbappend
uses PACKAGECONFIG_remove_imxgpu2d when it should be
PACKAGECONFIG_remove_pn-imxgpu2d, which makes the remove apply
to the recipe this is appended to (gtk+3 from openembedded-core),
which breaks gtk+3 when both wayland and x11 are in your distro
features as they are when requesting x11 backcompat for wayland.
It also makes the remove not actually apply to imxgpu2d.

Fixing the spelling here fixes both issues.

Signed-off-by: Seth Foster <seth@opentrons.com>

### DIFF
--- a/recipes-graphics/gtk+/gtk+3_%.bbappend
+++ b/recipes-graphics/gtk+/gtk+3_%.bbappend
@@ -1,5 +1,5 @@
 DEPENDS_append_imxgpu2d = " virtual/egl"
 
-PACKAGECONFIG_remove_imxgpu2d = " \
+PACKAGECONFIG_remove_pn-imxgpu2d = " \
     ${@bb.utils.contains("DISTRO_FEATURES", "wayland", "x11", "", d)} \
 "


sfoster1:
Please let me know if you'd like any specific testing on this patch (it builds, runs, and correctly allows the openembedded-core gtk+3 to build with both `x11` and `wayland` in `DISTRO_FEATURES`, thereby providing the gtk+3 x11 backend) or if I should submit this via the list instead of as a github pull request.

otavio:
This seems to be not correct. What machine and distro did you check?

sfoster1:
> This seems to be not correct. What machine and distro did you check?

I was using a custom machine and distro; the distro is here https://github.com/Opentrons/meta-opentrons/blob/main/conf/distro/opentrons.conf and based on one from Toradex, which is a SOM vendor integrating (in this case) imx8mm - this is the `verdin-imx8mm` machine. The Toradex distro is here: http://git.toradex.com/cgit/meta-toradex-distro.git/tree/conf/distro/tdx-xwayland.conf .

The direct problem is that when running a build with the linked distro and `verdin-imx8mm`, with `x11 wayland` in `DISTRO_FEATURES` courtesy of the Toradex distro above, gtk+3 does not have an x11 backend generated when running `bitbake gtk+3`. Doing `bitbake -e gtk+3 | grep '^PACKAGECONFIG'` reveals that `x11` is not present in the `PACKAGECONFIG` arguments and thus the required autoconf args aren't passed to build the `x11` backend.

By making only the change in this pull request, rebuilding `bitbake gtk+3` includes the x11 backend and `bitbake -e gtk+3 | grep '^PACKAGECONFIG'` shows `x11` in the packageconfig and the appropriate configure args being passed.

I definitely understand this could be wrong; I see both `PACKAGECONFIG_remove_pn-packagename` and `PACKAGECONFIG_remove_packagename` in various recipes throughout my layer tree, with some examples of both in meta-freescale-distro and meta-openembedded. I can't really find documentation or source on how this remove works, so I'm not sure which is correct, but the patch does fix the issue for me.

otavio:
@sfoster1 that is intended as i.MX8MM BSP doesn't offer X11 support and you'd use the Wayland backend instead.

Using the Wayland backend for GTK+3 didn't work?

sfoster1:
> @sfoster1 that is intended as i.MX8MM BSP doesn't offer X11 support and you'd use the Wayland backend instead.
> 
> Using the Wayland backend for GTK+3 didn't work?

Ah, I understand. No, unfortunately I was trying to run an electron app that runs on an older version of electron which links directly to the x backend for gtk+. I'll close this PR I suppose, though I will say that it does build and work with the x backend via xwayland. My hope is to update the electron version to run directly on wayland anyway.

PR #844, opened 2021-08-18
imx-gpu-apitrace: exclude from builds with glibc 2.34+

zandrey:
Upstream project has issues with new glibc dropping `__libc_dlopen_mode` and `__libc_dlsym` functions, which were initially marked as **GLIBC_PRIVATE**.

Since according to https://github.com/apitrace/apitrace/issues/756 there is no solution for upstream at the moment, exclude the NXP fork of _apitrace_ from builds.

meta-oe has similar exclusion commit ba94b0fef ("apitrace: Exclude from builds with glibc 2.34+") for upstream component.

Cc: @thochstein 
Cc: @kraj


otavio (approved):

thochstein:
Thanks for the info, @zandrey

PR #855, opened 2021-08-31
Kernel upgrade (linux-fslc to 5.14)

zandrey:
Kernel branch of `linux-fslc` was updated to _v5.14_ and config files were synchronized with upstream.

-- andrey


recipes-kernel/linux/linux-fslc/defconfig
```diff
@@ -162,6 +162,7 @@ CONFIG_WL18XX=m
 CONFIG_WLCORE_SDIO=m
 # CONFIG_WILINK_PLATFORM_DATA is not set
 CONFIG_INPUT_EVDEV=y
+CONFIG_INPUT_EVBUG=m
```
otavio, line 165: This is problematic; please drop it or add it to the loading blacklist.

otavio (changes requested):

recipes-kernel/linux/linux-fslc/defconfig
zandrey: Oh, this change was not intended but somehow slipped through.

Thanks for pointing it out, would drop and send a _V2_ now.

zandrey:
Changes in _V2_:
- Drop config re-sync for arch/arm/configs/imx_v6_v7_defconfig -> defconfig as the only option changed was `CONFIG_INPUT_EVBUG`, which is undesired.

PR #856, opened 2021-09-01
Gstreamer: Upgrade to 1.18.0[.imx]

thochstein:


otavio (approved):

kraj (commented):
This breaks musl builds see https://errors.yoctoproject.org/Errors/Details/603647/
@otavio I think we should add a musl build to CI which could help identify these issues ahead of time.

otavio:
@kraj can you prepare a PR fixing those?

PR #857, opened 2021-09-02
imx-gpu-apitrace: enable back for glibc >= 2.34

zandrey:
[Upsteam suggested](https://github.com/apitrace/apitrace/issues/756) that for glibc >= 2.34, dlsym wrappers can be disabled.

Similar commit is introduced in meta-oe:
`9fdc7960b ("apitrace: Enable on glibc >= 2.34")`

Cc: @thochstein


otavio (approved):

thochstein:
Thanks @zandrey. It's in apitrace master now: https://github.com/apitrace/apitrace/commit/d28a980802ad48568c87da02d630c8babfe163bb

zandrey:
@thochstein  Thanks for pointing it out!

I've backported the patch to the version `9.0`, which is currently in the NXP fork.
I'll keep in mind that when upgrade would be released - this patch would need to be replaced with the one from upstream.

-- andrey

PR #858, opened 2021-09-02
weston-init: modify weston.service file in do_install_append

chrisdimich:
As per:

https://source.codeaurora.org/external/imx/meta-imx/tree/meta-bsp/recipes-graphics/wayland/weston-init.bbappend?h=hardknott-5.10.35-2.0.0

Replace weston user/group with root. weston should be run as weston,
not as root, but needed for now to run things like GPU demos or
gstreamer pipelines as the default root user in the BSP.

Signed-off-by: Chris Dimich <Chris.Dimich@boundarydevices.com>

### DIFF
--- a/recipes-graphics/wayland/weston-init.bbappend
+++ b/recipes-graphics/wayland/weston-init.bbappend
@@ -46,6 +46,13 @@ uncomment() {
     sed -i -e 's,^#'"$1"','"$1"',g' $2
 }
 
+update_file() {
+    if ! grep -q "$1" $3; then
+        bbfatal $1 not found in $3
+    fi
+    sed -i -e "s,$1,$2," $3
+}
+
 do_install_append() {
     if [ -f "${WORKDIR}/weston.config" ]; then
         install -Dm0755 ${WORKDIR}/weston.config ${D}${sysconfdir}/default/weston
@@ -53,4 +60,8 @@ do_install_append() {
     for assignment in ${INI_UNCOMMENT_ASSIGNMENTS}; do
         uncomment "$assignment" ${D}${sysconfdir}/xdg/weston/weston.ini
     done
+
+    # FIXME: weston should be run as weston, not as root
+    update_file "User=weston" "User=root" ${D}${systemd_system_unitdir}/weston.service
+    update_file "Group=weston" "Group=root" ${D}${systemd_system_unitdir}/weston.service
 }


otavio:
Not really; what need to be done is add the weston user to the video group, I believe.

chrisdimich:
There is no video group that I see in the BSP. Can you please elaborate?

otavio:
The udev-imx-rules set it: https://github.com/Freescale/meta-freescale/blob/41a108b3453b5855d090a62dd451855ab6f9dad9/recipes-core/udev/udev-rules-imx/10-imx.rules#L21-L23

chrisdimich:
I set the user and group back to weston in weston.service and added this udev rule recipe to the image, yet still for example I cannot run the demo "/opt/imx-gpu-sdk/GLES2/Blur/GLES2.Blur_Wayland", I get the error "ERROR: demo setup failed with: wl_display_connect Failure". Or another example is pipeline fails for gst-launch + playbin for a movie file because user is not root.

thochstein:
Hi @chrisdimich, can you try adding video group to weston?

`$ usermod -a -G video weston`

chrisdimich:
Unfortunately that did not help either. When I login as root, I do not see the video group. When I login as weston, I see the video group but again I face the same issues.

chrisdimich:
@otavio @thochstein gentle reminder, this solution is not working for me. Can you please advise?

thochstein:
I wonder if the group is created properly. Did you try adding it?

`$ groupadd video`

thochstein:
Well, at least I do see it on my board in `/etc/group`.

chrisdimich:
The group exists, but still I face same issue. 

root@nitrogen8mp:~#  groupadd video
groupadd: group 'video' already exists

root@nitrogen8mp:~# cat /etc/group | grep video
video:x:44:weston

root@nitrogen8mp:~# cat /etc/group | grep weston
input:x:19:weston
video:x:44:weston
weston-launch:x:995:
weston:x:1000:

otavio:
And how does the /dev/galcore file looks?

thochstein:
> And how does the /dev/galcore file looks?

```
root@imx8mq-evk:~# ls -l /dev/galcore
crw-rw----    1 root     video     199,   0 Aug  6 14:41 /dev/galcore
```

thochstein:
@chrisdimich, where do you run from? I'm somewhat limited at the moment and get an error when I try to run from the console:

```
root@imx8mq-evk:~# su - weston
imx8mq-evk:~$ /opt/imx-gpu-sdk/GLES2/Blur/GLES2.Blur_Wayland
error: XDG_RUNTIME_DIR not set in the environment.
ERROR: demo setup failed with: wl_display_connect Failure
```

I need to figure out how to set `XDR_RUNTIME_DIR` appropriately, but wanted to make sure you're not also seeing this.

chrisdimich:
@thochstein I see the same error. "ERROR: demo setup failed with: wl_display_connect Failure". Whether I run from the service or manually I see the same.

otavio:
@chrisdimich @thochstein, you might be mixing things here. The user won't have it exported (as it doesn't normally have `DISPLAY` in X case).

Inside `weston-terminal` does it work?

The `weston-start` set it accordingly. See:
https://github.com/openembedded/openembedded-core/blob/50154869b496cf63a4fae2c8278350fcf9a2591d/meta/recipes-graphics/wayland/weston-init/weston-start#L74-L84

thochstein:
Running from `weston-terminal` works fine for me. Running from `console` or from `ssh` is problematic because the shell does not have `XDG_RUNTIME_DIR`.

The problem is avoided in `meta-imx` layers by updating the user profile:

https://source.codeaurora.org/external/imx/meta-imx/commit/meta-bsp/recipes-graphics/wayland/weston-init/profile?h=hardknott-5.10.35-2.0.0&id=e569336c75a521b605ddaada0ba52088ec68097e

I confirmed that setting `XDG_RUNTIME_DIR` does allow me to run demo successfully from the console.

@chrisdimich does this fix your issue?

@otavio assuming it does fix the issue, do you want me to submit a patch? I submitted the patch in the past and it was rejected.

otavio:
On meta-freescale we need to follow OE-Core. In meta-freescale-distro we can do it.

chrisdimich:
Yes, this fixes the issue if I switch user to weston "su - weston".  Thanks!

PR #865, opened 2021-09-14
imx53: handle non-DPU device

twoerner:
Wouter Vanhauwaert discovered that the latest linux-fslc kernel would panic
when booting on an imx53-based device. Fabio Estevam prepared the following
patch. We'll carry this patch locally until it is applied upstream.

Reported-by: Wouter Vanhauwaert <wouter.vanhauwaert@gmail.com>
Fixed-by: Fabio Estevam <festevam@gmail.com>


otavio:
@twoerner, @fabioestevam sent the patch to our fork; in this case, bumping the recipe should do the trick.

PR #866, opened 2021-09-14
linux-fslc: bump revision to include msm fix

twoerner:
Wouter Vanhauwaert discovered that the latest linux-fslc kernel would panic
when booting on an imx53-based device:
https://github.com/Freescale/meta-freescale/issues/864

Fabio Estevam added a fix to linux-fslc, therefore we bump the revision to
pick up this fix.


twoerner:
oops

```
ERROR: linux-fslc-5.14+gitAUTOINC+4b88a1ee28-r0 do_kernel_version_sanity_check: Package Version (5.14+gitAUTOINC+4b88a1ee28) does not match of kernel being built (5.14.3). Please update the PV variable to match the kernel source or set KERNEL_VERSION_SANITY_SKIP="1" in your recipe.
```

PR #870, opened 2021-09-17
imx-boot-container.bbclass: support having more than one UBOOT_CONFIG

Villemoes:
The whole purpose of the UBOOT_CONFIG machinery is to allow one to
build multiple versions of U-Boot within the same recipe, e.g. one for
writing to an eMMC boot partition, and one for use with an SD card for
bootstrapping/recovery. But the current code here assumes UBOOT_CONFIG
consists of just a single word (and normally leading whitespace in that
variable would be harmless).

So use the variable ${type} which gets set in turn to each individual
word in UBOOT_CONFIG. For compatibility, keep creating the u-boot.itb
and flash.bin symlinks, but only for the first type in
UBOOT_CONFIG.

It might be better (a little less arbitrary) to make the test
something like

  if [ ${type} = "sd" ]

but that's not entirely equivalent to the current code if UBOOT_CONFIG
does consist of a single word, but not the word "sd". Long-term, it's
likely better to just update the wic definition to point at the proper
file instead of adding compatibility symlinks.

### DIFF
--- a/classes/imx-boot-container.bbclass
+++ b/classes/imx-boot-container.bbclass
@@ -85,10 +85,17 @@ do_deploy_append() {
                 j=$(expr $j + 1);
                 if [ $j -eq $i ]
                 then
-                    install -m 0644 ${B}/${config}/u-boot.itb  ${DEPLOYDIR}/u-boot.itb-${MACHINE}-${UBOOT_CONFIG}
-                    install -m 0644 ${B}/${config}/flash.bin  ${DEPLOYDIR}/flash.bin-${MACHINE}-${UBOOT_CONFIG}
-                    ln -sf u-boot.itb-${MACHINE}-${UBOOT_CONFIG} u-boot.itb
-                    ln -sf flash.bin-${MACHINE}-${UBOOT_CONFIG} flash.bin
+                    install -m 0644 ${B}/${config}/u-boot.itb  ${DEPLOYDIR}/u-boot.itb-${MACHINE}-${type}
+                    install -m 0644 ${B}/${config}/flash.bin  ${DEPLOYDIR}/flash.bin-${MACHINE}-${type}
+                    # When there's more than one word in UBOOT_CONFIG,
+                    # this will overwrite the links created in
+                    # previous loop iterations, effectively making
+                    # u-boot.itb and flash.bin correspond to the _last_
+                    # word in UBOOT_CONFIG. This is also how all other
+                    # artifacts handled by oe-core's u-boot.inc are
+                    # treated.
+                    ln -sf u-boot.itb-${MACHINE}-${type} u-boot.itb
+                    ln -sf flash.bin-${MACHINE}-${type} flash.bin
                 fi
             done
             unset  j


classes/imx-boot-container.bbclass
```diff
@@ -85,10 +85,12 @@ do_deploy_append() {
                 j=$(expr $j + 1);
                 if [ $j -eq $i ]
                 then
-                    install -m 0644 ${B}/${config}/u-boot.itb  ${DEPLOYDIR}/u-boot.itb-${MACHINE}-${UBOOT_CONFIG}
-                    install -m 0644 ${B}/${config}/flash.bin  ${DEPLOYDIR}/flash.bin-${MACHINE}-${UBOOT_CONFIG}
-                    ln -sf u-boot.itb-${MACHINE}-${UBOOT_CONFIG} u-boot.itb
-                    ln -sf flash.bin-${MACHINE}-${UBOOT_CONFIG} flash.bin
+                    install -m 0644 ${B}/${config}/u-boot.itb  ${DEPLOYDIR}/u-boot.itb-${MACHINE}-${type}
+                    install -m 0644 ${B}/${config}/flash.bin  ${DEPLOYDIR}/flash.bin-${MACHINE}-${type}
+                    if [ $i -eq 1 ] ; then
```
zandrey, line 90: This _magic number_ does rely on the fact that the _SD_ build configuration is always set as the first element in the `UBOOT_MACHINE` (which is effectively default `imx8mm_evk_defconfig` for eMMC/SD boot config).

If there was an intention to link only the first U-Boot _build_ configuration, then perhaps the **j** should've been used in the `if` statement.

This is the extract from the unfolded `do_deploy()` task, where one can see the usage of those iterators:
```
    # Deploy the resulted flash.bin for WIC to pick it up
    if [ -n "sd" ]; then
        for config in  imx8mm_evk_defconfig; do
            i=$(expr $i + 1);
            for type in sd; do
                j=$(expr $j + 1);
                if [ $j -eq $i ]
                then
                    install -m 0644 /mainline/build-output/work/imx8mm_lpddr4_evk-leica-linux/u-boot-fslc/v2021.07+gitAUTOINC+691e634bfd-r0/build/${config}/u-boot.itb  /mainline/build-output/work/imx8mm_lpddr4_evk-leica-linux/u-boot-fslc/v2021.07+gitAUTOINC+691e634bfd-r0/deploy-u-boot-fslc/u-boot.itb-imx8mm-lpddr4-evk-sd
                    install -m 0644 /mainline/build-output/work/imx8mm_lpddr4_evk-leica-linux/u-boot-fslc/v2021.07+gitAUTOINC+691e634bfd-r0/build/${config}/flash.bin  /mainline/build-output/work/imx8mm_lpddr4_evk-leica-linux/u-boot-fslc/v2021.07+gitAUTOINC+691e634bfd-r0/deploy-u-boot-fslc/flash.bin-imx8mm-lpddr4-evk-sd
                    ln -sf u-boot.itb-imx8mm-lpddr4-evk-sd u-boot.itb
                    ln -sf flash.bin-imx8mm-lpddr4-evk-sd flash.bin
                fi
            done
            unset  j
        done
        unset  i
    fi

```

Besides, creating custom build configuration and adding them to `UBOOT_MACHINE` would imply that users do create their own separate _WIC_ files, where they control the layout and content of the final image.

In this case, I do not see any reason why the current implementation would not work - default links pointing to binaries from `imx8mm_evk_defconfig` build output do not collide with those deployed from additional build configurations defined in `UBOOT_MACHINE`.

Perhaps, what you intended to do is covered in the [OE-Core U-Boot implementation include file](https://github.com/openembedded/openembedded-core/blob/6450138afebffcc55ab32afadd5fb979274fff2b/meta/recipes-bsp/u-boot/u-boot.inc#L229) and just needs to be ported here.

This would allow to have several independent boot files to be deployed, each having the `${type}` in their names, while keeping default links from `imx8mm_evk_defconfig` pointing to `flash.bin` and `u-boot.itb` in order for default _WIC_ file to operate proper.

-- andrey

zandrey (changes requested):
@Villemoes 

There is a `Signed-off-by:` tag missing, perhaps you didn't use `--signoff` option to GIT?

Besides, there are some style problems with the patch:

classes/imx-boot-container.bbclass
Villemoes: > This _magic number_ does rely on the fact that the _SD_ build configuration is always set as the first element in the `UBOOT_MACHINE` (which is effectively default `imx8mm_evk_defconfig` for eMMC/SD boot config).

No, it does not, it simply preserves compatibility with existing setups, whether that setup has the word sd or not. The current code falls completely apart if UBOOT_CONFIG had more than one word (just see what those command lines would expand to, it's completely nonsensical).
 
> If there was an intention to link only the first U-Boot _build_ configuration, then perhaps the **j** should've been used in the `if` statement.

We're inside an `if [ $i -eq $j ]` block, so how does that matter?

> This is the extract from the unfolded `do_deploy()` task, where one can see the usage of those iterators:
> 
> ```
>     # Deploy the resulted flash.bin for WIC to pick it up
>     if [ -n "sd" ]; then
>         for config in  imx8mm_evk_defconfig; do
>             i=$(expr $i + 1);
>             for type in sd; do
>                 j=$(expr $j + 1);
>                 if [ $j -eq $i ]
>                 then
>                     install -m 0644 /mainline/build-output/work/imx8mm_lpddr4_evk-leica-linux/u-boot-fslc/v2021.07+gitAUTOINC+691e634bfd-r0/build/${config}/u-boot.itb  /mainline/build-output/work/imx8mm_lpddr4_evk-leica-linux/u-boot-fslc/v2021.07+gitAUTOINC+691e634bfd-r0/deploy-u-boot-fslc/u-boot.itb-imx8mm-lpddr4-evk-sd
>                     install -m 0644 /mainline/build-output/work/imx8mm_lpddr4_evk-leica-linux/u-boot-fslc/v2021.07+gitAUTOINC+691e634bfd-r0/build/${config}/flash.bin  /mainline/build-output/work/imx8mm_lpddr4_evk-leica-linux/u-boot-fslc/v2021.07+gitAUTOINC+691e634bfd-r0/deploy-u-boot-fslc/flash.bin-imx8mm-lpddr4-evk-sd
>                     ln -sf u-boot.itb-imx8mm-lpddr4-evk-sd u-boot.itb
>                     ln -sf flash.bin-imx8mm-lpddr4-evk-sd flash.bin
>                 fi
>             done
>             unset  j
>         done
>         unset  i
>     fi
> ```
> 
> Besides, creating custom build configuration and adding them to `UBOOT_MACHINE` would imply that users do create their own separate _WIC_ files, where they control the layout and content of the final image.

Absolutely true, and in fact I have no intention of using WIC at all for anything. For the sd image, I use genimage, and for the stuff that is to be written raw to an eMMC boot partition, well, I don't need anything other than the flash.bin as deployed.

But the thing is, as noted above, the moment I do add a second UBOOT_CONFIG, this shell snippet breaks apart. So I could just not use the imx-boot-container class and do my own deploy, of course, but I actually do want the "stage the ddr firmware" logic.

> In this case, I do not see any reason why the current implementation would not work - default links pointing to binaries from `imx8mm_evk_defconfig` build output do not collide with those deployed from additional build configurations defined in `UBOOT_MACHINE`.

Please try to read the code again and see what happens when, for example, UBOOT_CONFIG is "sd emmc".

zandrey: > No, it does not, it simply preserves compatibility with existing setups, whether that setup has the word sd or not. 


This should be extended with the wording: "provided that the _sd_ is at the first location in the `UBOOT_CONFIG `", isn't it? This is what I was referring to with the statement above.

> The current code falls completely apart if UBOOT_CONFIG had more than one word (just see what those command lines would expand to, it's completely nonsensical).

I did not state that the block is perfect, I actually pointed out the link in my original message, which shows the proper implementation done in `u-boot.inc` file. Now I tend to think that this block needs re-work in order accommodate for multiple `IMAGE_FSTYPES` which actually you brought up with this commit.

What I do not agree with is the way you propose, not the idea itself.

> 
> > If there was an intention to link only the first U-Boot _build_ configuration, then perhaps the **j** should've been used in the `if` statement.
> 
> We're inside an `if [ $i -eq $j ]` block, so how does that matter?

For the sake of readability, but this is irrelevant as the entire `if` statement makes little sense.

> >   
> > Besides, creating custom build configuration and adding them to `UBOOT_MACHINE` would imply that users do create their own separate _WIC_ files, where they control the layout and content of the final image.
> 
> Absolutely true, and in fact I have no intention of using WIC at all for anything. For the sd image, I use genimage, and for the stuff that is to be written raw to an eMMC boot partition, well, I don't need anything other than the flash.bin as deployed.
> 

I believe that for both of your boot sources you would need `u-boot.itb` unless you go for an SPL+DDR bundle, right?

> But the thing is, as noted above, the moment I do add a second UBOOT_CONFIG, this shell snippet breaks apart. So I could just not use the imx-boot-container class and do my own deploy, of course, but I actually do want the "stage the ddr firmware" logic.

This is fine, this was the intention of this class on the first place - collect all build prequesites in order to produce a bootable binary image. Please note, that ATF would also be required, as `binman` has it listed in.

> 
> > In this case, I do not see any reason why the current implementation would not work - default links pointing to binaries from `imx8mm_evk_defconfig` build output do not collide with those deployed from additional build configurations defined in `UBOOT_MACHINE`.
> 
> Please try to read the code again and see what happens when, for example, UBOOT_CONFIG is "sd emmc".

Here, the question is: what is defined in your `UBOOT_CONFIG[emmc]` then? There is no such variable defined in the layer for `mx8m` machine...

If you need additional `IMAGE_FSTYPES` for the eMMC boot, then it should be defined as:
```
UBOOT_CONFIG[emmc]="imx8mm_evk_defconfig,emmc"
```

In such case, the deploy block should be modified as in the link I've posted above (from OE-Core U-Boot include) to include ${type} in the deployment naming scheme.

Villemoes: > > > Besides, creating custom build configuration and adding them to `UBOOT_MACHINE` would imply that users do create their own separate _WIC_ files, where they control the layout and content of the final image.
> > 
> > 
> > Absolutely true, and in fact I have no intention of using WIC at all for anything. For the sd image, I use genimage, and for the stuff that is to be written raw to an eMMC boot partition, well, I don't need anything other than the flash.bin as deployed.
> 
> I believe that for both of your boot sources you would need `u-boot.itb` unless you go for an SPL+DDR bundle, right?

Yes, _but I may not want the same u-boot.itb if I have used different configurations to build it_. Please read what I've written, otherwise conversation is impossible. So it makes no sense for the plain `u-boot.itb` to point at one or the other artifact - I need the (for example) `u-boot.itb-sd` for one downstream purpose (creating a recovery SD card), and `u-boot.itb-emmc` as an artifact to write to an emmc boot partition. Those two binary files may not be functionally or binary equivalent, and I very well may have configured them differently.

> 
> > > In this case, I do not see any reason why the current implementation would not work - default links pointing to binaries from `imx8mm_evk_defconfig` build output do not collide with those deployed from additional build configurations defined in `UBOOT_MACHINE`.

[Can you _please_ stop with saying UBOOT_MACHINE when you mean UBOOT_CONFIG. UBOOT_MACHINE is, when using UBOOT_CONFIG at all, re-purposed internally by uboot-config.bbclass (or whatever its called). That's a silly mess, but one we have to live with.] Let me try once again: The current code _completely breaks_ when one has more than one word in UBOOT_CONFIG. So it's not a case of build output colliding or not, because that situation simply doesn't exist currently because of the naive use of `-${UBOOT_CONFIG}` in the current shell script. You probably _meant_ to say `-${type}` all along, but you didn't. So stop pretending that your code would work at all with UBOOT_CONFIG set to anything other than a single word.

> > Please try to read the code again and see what happens when, for example, UBOOT_CONFIG is "sd emmc".
> 
> Here, the question is: what is defined in your `UBOOT_CONFIG[emmc]` then? There is no such variable defined in the layer for `mx8m` machine...

So what? I'm not building for some eval kit board, I'm building for a custom board based on an imx8 SOC. So I have my own machine `.conf` file, with my own UBOOT_CONFIG[...] settings. And yes, I'm also pointing at my own `_defconfig` files.

> If you need additional `IMAGE_FSTYPES` for the eMMC boot, then it should be defined as:
> 
> ```
> UBOOT_CONFIG[emmc]="imx8mm_evk_defconfig,emmc"
> ```
> 
> In such case, the deploy block should be modified as in the link I've posted above (from OE-Core U-Boot include) to include ${type} in the deployment naming scheme.

Which is _precisely_ what the first two lines of my patch does, TYVM. Then there's the question of what the unsuffixed (as in, does not contain `${type}` in their names) links should point at, which is a completely different discussion. I now see that the core u-boot files "solve" that problem the other way around, by just ignoring the issue and thus effectively ending up with the unsuffixed link pointing at the artifact coming from the _last_ word in UBOOT_CONFIG. I don't care either way, so I'll just remove the inner `if`, possibly adding a comment that yes, that clobbers links created just microseconds earlier, but that's how it is...

zandrey: > 
> 
> > > > Besides, creating custom build configuration and adding them to `UBOOT_MACHINE` would imply that users do create their own separate _WIC_ files, where they control the layout and content of the final image.
> > > 
> > > 
> > > Absolutely true, and in fact I have no intention of using WIC at all for anything. For the sd image, I use genimage, and for the stuff that is to be written raw to an eMMC boot partition, well, I don't need anything other than the flash.bin as deployed.
> > 
> > 
> > I believe that for both of your boot sources you would need `u-boot.itb` unless you go for an SPL+DDR bundle, right?
> 
> Yes, _but I may not want the same u-boot.itb if I have used different configurations to build it_. 

Agree, this is what I was pointing on myself: the current code in the class does not handle that case. But your originally proposed solution was just a specialization that you need, not a solution for all cases.

> Please read what I've written, otherwise conversation is impossible. So it makes no sense for the plain `u-boot.itb` to point at one or the other artifact - I need the (for example) `u-boot.itb-sd` for one downstream purpose (creating a recovery SD card), and `u-boot.itb-emmc` as an artifact to write to an emmc boot partition. Those two binary files may not be functionally or binary equivalent, and I very well may have configured them differently.

This is exactly what I meant with using `-{type}` in the suffix, like upstream does. You updated patch does exactly that, and this is a better way of handling all the cases.

> 
> > > > In this case, I do not see any reason why the current implementation would not work - default links pointing to binaries from `imx8mm_evk_defconfig` build output do not collide with those deployed from additional build configurations defined in `UBOOT_MACHINE`.
> 
> [Can you _please_ stop with saying UBOOT_MACHINE when you mean UBOOT_CONFIG. UBOOT_MACHINE is, when using UBOOT_CONFIG at all, re-purposed internally by uboot-config.bbclass (or whatever its called). That's a silly mess, but one we have to live with.] 

Exactly for that reason I was referring to `UBOOT_CONFIG` as `UBOOT_MACHINE` - it is the upstream implementation that does the re-purposing, so I was using upstream terminology. There is no discussion point here, if you feel that upstream is broken - you're welcomed to send patches to OE-Core.

> Let me try once again: The current code _completely breaks_ when one has more than one word in UBOOT_CONFIG. So it's not a case of build output colliding or not, because that situation simply doesn't exist currently because of the naive use of `-${UBOOT_CONFIG}` in the current shell script. You probably _meant_ to say `-${type}` all along, but you didn't. So stop pretending that your code would work at all with UBOOT_CONFIG set to anything other than a single word.

Yes, I was expecting the `-${type}` suffix in this patch which would accommodate for all cases, and this is what you've introduced in the updated commit. And if you pay a bit more attention to my replies, I never pretended that the code worked for multiple `UBOOT_CONFIG`, I suggested it needs a change in here: 
```
Perhaps, what you intended to do is covered in the OE-Core U-Boot implementation include file and just needs to be ported here.

This would allow to have several independent boot files to be deployed, each having the ${type} in their names, while keeping default links from imx8mm_evk_defconfig pointing to flash.bin and u-boot.itb in order for default WIC file to operate proper.
```

Hope it's clear now.

> 
> > > Please try to read the code again and see what happens when, for example, UBOOT_CONFIG is "sd emmc".
> > 
> > 
> > Here, the question is: what is defined in your `UBOOT_CONFIG[emmc]` then? There is no such variable defined in the layer for `mx8m` machine...
> 
> So what? I'm not building for some eval kit board, I'm building for a custom board based on an imx8 SOC. So I have my own machine `.conf` file, with my own UBOOT_CONFIG[...] settings. And yes, I'm also pointing at my own `_defconfig` files.

Like all of us, but EvalKit is a reference point for those who just start to build their designs, right?

> 
> > If you need additional `IMAGE_FSTYPES` for the eMMC boot, then it should be defined as:
> > ```
> > UBOOT_CONFIG[emmc]="imx8mm_evk_defconfig,emmc"
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > In such case, the deploy block should be modified as in the link I've posted above (from OE-Core U-Boot include) to include ${type} in the deployment naming scheme.
> 
> Which is _precisely_ what the first two lines of my patch does, TYVM. Then there's the question of what the unsuffixed (as in, does not contain `${type}` in their names) links should point at, which is a completely different discussion. I now see that the core u-boot files "solve" that problem the other way around, by just ignoring the issue and thus effectively ending up with the unsuffixed link pointing at the artifact coming from the _last_ word in UBOOT_CONFIG. I don't care either way, so I'll just remove the inner `if`, possibly adding a comment that yes, that clobbers links created just microseconds earlier, but that's how it is...

Which is _precisely_ what I meant with all my comment above: if you need multiple deploy targets - please use `-${type}` suffix and craft it in your image file after the package is built. Which is what your updated patch handles well now, but not the originally proposed one.

zandrey (approved):
For the _V2_:

`Reviewed-by: Andrey Zhizhikin <andrey.z@gmail.com>`

classes/imx-boot-container.bbclass
Villemoes: > Which is _precisely_ what I meant with all my comment above: if you need multiple deploy targets - please use `-${type}` suffix and craft it in your image file after the package is built. Which is what your updated patch handles well now, but not the originally proposed one.

This is simply false, and I'm sorry that we've had all this back-and-forth because you didn't read that patch properly. That was precisely what I did (and still do). Here's the original commit: https://github.com/villemoes/meta-freescale/commit/b68ef55c48924bda690a3b04deaa8a8c6023ad4d . Lines 88 and 89 simply consist of changing `-${UBOOT_CONFIG}` to `-$type`. And as `git range-diff 72f59d35 b68ef55c 079e2936` clearly shows, that part is unchanged from v1 to v2.

Then the second part tried to deal with what to do with unsuffixed symlinks. And yes, it's better to be consistent with how other artifacts are handled, so the new version's "let the final one take precedence" is clearly better. Not because it avoids an inner `if`, but merely because that's how all the other `do_deploy` fragments are written. It's still mostly arbitrary, and if one does have multiple entries in UBOOT_CONFIG _and_ relies on the `u-boot.itb` or `flash.bin` names, one needs to know that there's a difference between `UBOOT_CONFIG = "sd emmc"` and `UBOOT_CONFIG = "emmc sd"`.

Thanks for the R-b. I think we can usefully stop here.

zandrey: > 
> 
> > Which is _precisely_ what I meant with all my comment above: if you need multiple deploy targets - please use `-${type}` suffix and craft it in your image file after the package is built. Which is what your updated patch handles well now, but not the originally proposed one.
> 
> This is simply false, and I'm sorry that we've had all this back-and-forth because you didn't read that patch properly. That was precisely what I did (and still do). Here's the original commit: [Villemoes@b68ef55](https://github.com/Villemoes/meta-freescale/commit/b68ef55c48924bda690a3b04deaa8a8c6023ad4d) . Lines 88 and 89 simply consist of changing `-${UBOOT_CONFIG}` to `-$type`. And as `git range-diff 72f59d35 b68ef55c 079e2936` clearly shows, that part is unchanged from v1 to v2.

Just for the record: I was not objecting the addition of the `-${type}` suffix, in contrary - this is needed as I understood the problem straight away and realized the fault. If you read my original comment - it starts with line 90, where the inner `if` was introduced, which brings us the second part:

> 
> Then the second part tried to deal with what to do with unsuffixed symlinks. And yes, it's better to be consistent with how other artifacts are handled, so the new version's "let the final one take precedence" is clearly better. Not because it avoids an inner `if`, but merely because that's how all the other `do_deploy` fragments are written. It's still mostly arbitrary, and if one does have multiple entries in UBOOT_CONFIG _and_ relies on the `u-boot.itb` or `flash.bin` names, one needs to know that there's a difference between `UBOOT_CONFIG = "sd emmc"` and `UBOOT_CONFIG = "emmc sd"`.

This is clearly a miss in the upstream implementation, and I also saw that. But there is no easy fix for this ordering issue, other than try to perhaps solve it in the way `uboot-extlinux-config.bbclass` does: create a `UBOOT_CONFIG_DEFAULT` variable, which would point to default entry in `UBOOT_CONFIG` that in turn would point to a configuration which should be unsuffixed... This is definitely something that needs to be brought up to upstream, and once accepted - this layer would pick it up.

> 
> Thanks for the R-b. I think we can usefully stop here.

Agree. 👍🏻

thochstein (approved):
LGTM

PR #889, opened 2021-10-18
linux-qoriq and qoriq-atf: add back old version

ting-liu:
add back linux-qoriq_5.4.bb and qoriq-atf_1.5.bb to not break 3rdparty build.


otavio:
As this is used on 3dparty only, it'd be better to add those there.

zandrey:
@ting-liu Is it possible you proceed with this PR? I guess the suggestion from @otavio is really legit, so the old versions are to me moved to where bbappends now reside.

Without this fixed, the entire CI/CD is broken on master, and I guess this is not only for me.

@otavio As an alternative, I can make an RFC PR in `meta-freescale-3dparty` to simply delete those dangling appends.

-- andrey

ting-liu:
https://github.com/Freescale/meta-freescale-3rdparty/pull/173 is created to replace this one.

PR #902, opened 2021-10-29
Enable ION memory manager for i.MX6 and 7

dv1:
The linux-imx and linux-fslc-imx kernels have the ION memory manager, which can be used for allocating DMA-BUF memory. ION is already enabled for the i.MX8/8m/8qm/etc. SoCs. On i.MX6 and 7, it isn't, so there is no way to allocate DMA-BUF memory on these platforms. Physically contiguous memory allocators are available, but these don't use DMA-BUF.

Rectify this by enabling ION in these two kernels.

This is only relevant for Yocto Dunfell. The linux-imx and linux-fslc-imx kernels in Yocto Hardknott and Yocto Honister are at version 5.10, where the dma-heap allocator is available. That is the successor to the ION allocator, and should be preferred over ION. A future pull request will be opened separately which enables dma-heap for i.MX6 and 7 SoCs.

The libimxdmabuffer recipe is adjusted accordingly to enable ION on all i.MX SoCs, since with this kernel configuration change, ION will be enabled on all SoCs that are supported by this library.


recipes-kernel/linux/linux-fslc-imx/imx/defconfig
```diff
@@ -577,3 +577,5 @@ CONFIG_MAGIC_SYSRQ=y
 # CONFIG_DEBUG_PREEMPT is not set
 # CONFIG_DEBUG_BUGVERBOSE is not set
 # CONFIG_FTRACE is not set
+CONFIG_ION=y
+CONFIG_ION_CMA_HEAP=y
```
otavio, line 581: I am fine in adding this but please do the same as a PR to linux-fslc repo so the original defconfig files are in sync.

otavio (changes requested):

recipes-kernel/linux/linux-fslc-imx/imx/defconfig
dv1: Alright, then I'll append these two lines to the linux-fslc defconfig as well.

zandrey:
@dv1 I've been looking into the ION for other derivative and figured out that it is actually comes to the state where it is planned to be abandoned.

Google [moved away from ION in favor of DMA-BUF](https://source.android.com/devices/architecture/kernel/dma-buf-heaps), and [this LWN article](https://lwn.net/Articles/792733/) suggests that ION would be subjected to de-staging.

Perhaps you can also consider to move `libimxdmabuffer` to use DMA-BUF instead, and drop the ION support. I'm certainly doing this for the other SoC I have on my table now.

-- andrey

dv1:
@zandrey As I wrote:

> This is only relevant for Yocto Dunfell. The linux-imx and linux-fslc-imx kernels in Yocto Hardknott and Yocto Honister are at version 5.10, where the dma-heap allocator is available. That is the successor to the ION allocator, and should be preferred over ION. A future pull request will be opened separately which enables dma-heap for i.MX6 and 7 SoCs.

libimxdmabuffer already supports DMA-BUF. That's the whole point of this pull request - to be able to use DMA-BUF on imx6 and imx7 machines in Dunfell.

otavio:
@dv1 on master it already works, right? Please rebase.

zandrey:
> @zandrey As I wrote:
> 
> > This is only relevant for Yocto Dunfell. The linux-imx and linux-fslc-imx kernels in Yocto Hardknott and Yocto Honister are at version 5.10, where the dma-heap allocator is available. That is the successor to the ION allocator, and should be preferred over ION. A future pull request will be opened separately which enables dma-heap for i.MX6 and 7 SoCs.
> 
> libimxdmabuffer already supports DMA-BUF. That's the whole point of this pull request - to be able to use DMA-BUF on imx6 and imx7 machines in Dunfell.

I see it now, thanks @dv1!

It was a bit confusing to me to see that even though `libimxdmabuffer` moves towards DMA-BUF, ION is still needed here. Now I see that you do this only for old Kernel releases, which is the only available for `dunfell`. Should've look better at the target branch. 😉 

BTW, just as a small clarification: what do you mean saying that the whole point of this PR to use DMA-BUF is to switch ION on? I do not see any connection between those 2 frameworks here, other than that ION is a predecessor of DMA-BUF... Can you give a bit more background here? Why do you need ION to move to DMA-BUF? Or is it just because `dunfell` Kernels do not have full support of DMA-BUF that you'd rather lean towards?

Overall, LGTM. 👍🏻 

-- andrey

dv1:
> @dv1 on master it already works, right? Please rebase.

No, this is pointless on master. Ultimately, it is all about being able to allocate DMA-BUF memory. Kernel 5.6 and newer support the dma-heap allocator, which replaces ION. Older kernels have to rely on ION. The kernels in meta-freescale Dunfell are at version 5.4. ION is already enabled for imx8 machines in Dunfell. This PR enables it for imx6 and imx7 as well to have a DMABUF capable allocator available on imx6, imx7, and imx8. Right now, only imx8 has this, even though this is purely a configuration issue and not hardware specific.

For git master, a different PR is necessary to enable dma-heap on imx6 and imx7. The goal there is to achieve the same, that is, an DMA-BUF allocator being available on imx6, imx7, and imx8.

dv1:
> BTW, just as a small clarification: what do you mean saying that the whole point of this PR to use DMA-BUF is to switch ION on? I do not see any connection between those 2 frameworks here, other than that ION is a predecessor of DMA-BUF... Can you give a bit more background here? Why do you need ION to move to DMA-BUF? Or is it just because `dunfell` Kernels do not have full support of DMA-BUF that you'd rather lean towards?

ION is a predecessor of dma-heap, not DMA-BUF. dma-heap is a userspace API for allocating DMA-BUF memory, and is available starting at kernel 5.6. Since meta-freescale has 5.4 kernels for dunfell, dma-heap is not an option in dunfell - only ION is available.

> 
> Overall, LGTM. 👍🏻

Cool, thanks :)

otavio (approved):

PR #906, opened 2021-10-31
Draft: Master isp imx

MaxKrummenacher:
isp-imx currently does not build, These two patches do address the do_patch and do_configure failures.

However there seems to be a binary only file deployed linked against an older version of libtinyxml2 and thus a QA error is generated. (libtinyxml2.so.9 is now provided from meta-openembedded) resulting in:

```
ERROR: isp-imx-4.2.2.15.0-r0 do_package_qa: QA Issue: /usr/lib/libcam_device.so contained in package isp-imx requires libtinyxml2.so.8()(64bit), but no providers found in RDEPENDS:isp-imx? [file-rdeps]
ERROR: isp-imx-4.2.2.15.0-r0 do_package_qa: QA run found fatal errors. Please consider fixing them.
ERROR: Logfile of failure stored in: /build/krm/oe-core_master/build/tmp/work/aarch64-mx8mp-tdx-linux/isp-imx/4.2.2.15.0-r0/temp/log.do_package_qa.6780
ERROR: Task (/build/krm/oe-core_master/build/../layers/meta-freescale/recipes-bsp/isp-imx/isp-imx_4.2.2.15.0.bb:do_package_qa) failed with exit code '1'
```

For this I have no solution yet, so I don't think we should merge this. I just wanted to make the first part of the fix public.

### DIFF
--- a/recipes-bsp/isp-imx/isp-imx/0001-isp-imx-drop-use-of-__TIME__-__DATE__.patch
+++ b/recipes-bsp/isp-imx/isp-imx/0001-isp-imx-drop-use-of-__TIME__-__DATE__.patch
@@ -1,6 +1,6 @@
-From accaeb1ae77eb40b89e70df3efcb00d0dc2af758 Mon Sep 17 00:00:00 2001
+From ea5ce303fa5bc7b6a17c522b4caf49d349adb12b Mon Sep 17 00:00:00 2001
 From: Max Krummenacher <max.krummenacher@toradex.com>
-Date: Mon, 14 Jun 2021 08:20:48 +0000
+Date: Sat, 30 Oct 2021 17:26:42 +0200
 Subject: [PATCH] isp-imx: drop use of __TIME__, __DATE__
 
 With reproducible build enabled the compiler is configured to throw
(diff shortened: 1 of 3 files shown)


otavio:
Have you built it for imx8?

thochstein:
Unfortunately `libtinyxml2` is upgraded to v9, while our archive built on `hardknott` was linked against v8. Can we add `libtinyxml2_8.0.0.bb` to the layer?

thochstein:
FYI, it looks like we're finally making some progress on integrating the patches. Sorry it is taking so long...

MaxKrummenacher:
I built for a i.MX 8M Plus based target. I think this is the only i.MX with that specific ISP.

thochstein:
LGTM

MaxKrummenacher:
@thochstein, @otavio 
I tried to add a libtinyxml2 8.0.0 recipe. That makes isp-imx build and with some magic to not deploy the '-dev' files can even be installed with the development package alongside libtinyxml2 9.0.0 as may be needed to e.g. build an SDK.

IMHO this is a terrible hack, it is way better to either build from source or make an abstraction library between libtinyxml2 and whatever closed source there is. Then the abstraction can be compiled against whatever libtinyxml2 version available.

Anyway, with the above 4 commits it builds an image and an sdk with isp-imx and libtinyxml in a 8.0.0 and 9.0.0 flavour.

otavio:
@MaxKrummenacher totally agree. It'd be statically linked to avoid this mess ... @thochstein please raise this internally as this is too ugly.

PR #917, opened 2021-11-04
opencv: Backport updates from OE-Core

thochstein:
2178fd7386 opencv: remove setup_vars_opencv4.sh
cafcc65e74 opencv: fix build with protobuf-3.18 when dnn PACKAGECONFIG is enabled
3c022cd50d opencv: Do not lock to gcc only compiler


thochstein:
This should be on honister as well.

otavio (approved):

PR #924, opened 2021-11-15
imx8qm-mek: remove obsolete dtb

migronof:
imx8qm-mek-dsp.dtb is no longer available in the kernel,
it was removed in de78ae45bfffed2c9490ab733d11ebab2672984d


otavio:
I believe this has been fixed on master so we ought to do a `git cherry-pick -x` so it is clear where the change is being copied from.

PR #927, opened 2021-11-17
packagegroup-fsl-isp: Mark as TUNE_PKGARCH specific

MaxKrummenacher:
Fixes non fatal error with latest openembedded-core master branch.

### DIFF
--- a/recipes-fsl/packagegroups/packagegroup-fsl-isp.bb
+++ b/recipes-fsl/packagegroups/packagegroup-fsl-isp.bb
@@ -1,5 +1,7 @@
 DESCRIPTION = "Add packages for ISP build"
 
+# basler-camera* gets dynamically renamed
+PACKAGE_ARCH = "${MACHINE_SOCARCH}"
 inherit packagegroup
 
 ISP_PKGS      ?= ""


recipes-fsl/packagegroups/packagegroup-fsl-isp.bb
```diff
@@ -1,5 +1,7 @@
 DESCRIPTION = "Add packages for ISP build"
 
+# basler-camera* gets dynamically renamed
+PACKAGE_ARCH = "${TUNE_PKGARCH}"
```
otavio, line 4: Shouldn't it to be `MACHINE_SOCARCH`?

otavio (changes requested):

recipes-fsl/packagegroups/packagegroup-fsl-isp.bb
MaxKrummenacher: I just took what what Richard proposed in his commit without thinking it through.
The packagegroup today is only used for iMX8MP, but I would think that the packagroup, i.e. the RRECOMMENDS and RDEPENDS in it would be applicable for any MACHINE_SOCARCH, even if the packages installed would have a different content depending on SOC.

MaxKrummenacher: On the hand, building it per soc arch will not eat much time and if the lib* packages get a different name per SOCARCH it cannot hurt. Will change to MACHINE_SOCARCH?

otavio (approved):

kraj:
I am seeing these errors on non fsl platforms

https://errors.yoctoproject.org/Errors/Build/136077/

Can we use something that will make it unbreak non-fsl platforms too ?

otavio:
Ohh, indeed. It is undefined in this case. I think it could be `MACHINE_ARCH` as it is very small and intended to change depending on machine. @MaxKrummenacher or @kraj, can you send a PR?

MaxKrummenacher:
I will send a PR.

otavio:
It has been merged already.

PR #933, opened 2021-12-03
imx-atf improvements (Clang ready)

zandrey:
This PR is generally a set of patches ported from _meta-arm_, which provides recipe for upstream TF-A.

It addresses the match of toolchain which is provided in the sysroot to what TF-A expects in it's Makefile, specifically:
- Unsets `CFLAGS`, `LDFLAGS`, `AS` and `LD` variables which are defined by TF-A itself
- Drops dependency on C-runtime as it is not required for TF-A (baremetal)
- Enables support to use _Clang_ for TF-A build by sanitizing `CC` and `LD` variables.

Original commits used for this port are linked in commits to this layer.

@thochstein This PR also drops the 79ba26618a075332ac9106f55c1aad573e42c9e0 as `CFLAGS` will be set by TF-A itself. I've tested this on `imx8mq_evk` machine and observed no build issues.

-- andrey

### DIFF
--- a/recipes-bsp/imx-atf/imx-atf_2.4.bb
+++ b/recipes-bsp/imx-atf/imx-atf_2.4.bb
@@ -25,13 +25,16 @@ EXTRA_OEMAKE += " \
     PLAT=${ATF_PLATFORM} \
 "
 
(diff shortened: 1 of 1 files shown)


thochstein (approved):
LGTM

otavio:
Please open an issue with this; are you willing to work on a fix for it?

PR #934, opened 2021-12-03
[RFC]: Use upstream U-Boot for imx8m Mini, Nano and Plus machines (Mainline BSP)

zandrey:
Since recent _U-Boot_ releases, Mini, Nano and Plus are supported in the upstream. The delta that is kept in `u-boot-fslc` repository is negligible, and to my knowledge does not contain any specific commit that are relevant to those derivatives.

Therefore, I wanted to post this RFC here with the attempt to switch _u-boot_ provider from `u-boot-fslc` to `u-boot` provided in the OE-Core.

Current version in upstream recipe is already newer that the one inside `u-boot-fslc` repository, so this would also effectively provide an upgrade of _U-Boot_ package to _2021.10_.

This would also allow to pickup upgrades from upstream rather than maintaining it in a separate repository (at least for those derivatives).

i.MX8M Quad does lack essential _binman_ support, therefore it cannot be used now with upstream U-Boot. Patch to address it is submitted, so once accepted - that machine can be switched over as well at the release this fix would be merged in.

Cc: @fabioestevam 

-- andrey

### DIFF
--- a/recipes-bsp/u-boot/u-boot_%.bbappend
+++ b/recipes-bsp/u-boot/u-boot_%.bbappend
@@ -1,0 +1,8 @@
+# So far, we just need a container to be inherited
+inherit ${@oe.utils.ifelse(d.getVar('UBOOT_PROVIDES_BOOT_CONTAINER') == '1', 'imx-boot-container', '')}
+
+# Location known to imx-boot component, where U-Boot artifacts
+# should be additionally deployed.
+# See below note above do_deploy:append:mx8m for the purpose of
+# this delopyment location
+BOOT_TOOLS = "imx-boot-tools"
--- a/conf/machine/include/imx8mm-evk.inc
+++ b/conf/machine/include/imx8mm-evk.inc
@@ -28,7 +28,7 @@ KERNEL_DEVICETREE:append:use-nxp-bsp = " \
 UBOOT_DTB_NAME = "${KERNEL_DEVICETREE_BASENAME}.dtb"
 
 IMX_DEFAULT_BOOTLOADER:use-nxp-bsp = "u-boot-imx"
-IMX_DEFAULT_BOOTLOADER:use-mainline-bsp = "u-boot-fslc"
+IMX_DEFAULT_BOOTLOADER:use-mainline-bsp = "u-boot"
 
 UBOOT_SUFFIX = "bin"
 
--- a/conf/machine/include/imx8mn-evk.inc
+++ b/conf/machine/include/imx8mn-evk.inc
@@ -28,7 +28,7 @@ KERNEL_DEVICETREE:append:use-nxp-bsp = " \
 UBOOT_DTB_NAME = "${KERNEL_DEVICETREE_BASENAME}.dtb"
 
 IMX_DEFAULT_BOOTLOADER:use-nxp-bsp = "u-boot-imx"
-IMX_DEFAULT_BOOTLOADER:use-mainline-bsp = "u-boot-fslc"
+IMX_DEFAULT_BOOTLOADER:use-mainline-bsp = "u-boot"
 
 UBOOT_SUFFIX = "bin"
 
--- a/conf/machine/include/imx8mp-evk.inc
+++ b/conf/machine/include/imx8mp-evk.inc
@@ -24,7 +24,7 @@ KERNEL_DEVICETREE = " \
 UBOOT_DTB_NAME = "${KERNEL_DEVICETREE_BASENAME}.dtb"
 
 IMX_DEFAULT_BOOTLOADER:use-nxp-bsp = "u-boot-imx"
-IMX_DEFAULT_BOOTLOADER:use-mainline-bsp = "u-boot-fslc"
+IMX_DEFAULT_BOOTLOADER:use-mainline-bsp = "u-boot"
 
 UBOOT_SUFFIX = "bin"
 


otavio:
I'd love to do it however, we still have patches that we need. How do you intend to do with them?

fabioestevam:
Hi Otavio,

On Fri, Dec 3, 2021 at 2:36 PM Otavio Salvador ***@***.***> wrote:
>
> I'd love to do it however, we still have patches that we need. How do you intend to do with them?

The last time we looked at the differences between u-boot-fslc versus
mainline, the result was:
https://github.com/Freescale/u-boot-fslc/pull/41#issuecomment-871343058

The first 4 have been added upstream.

2d107f6 ("imx: iMX6: Define disconnect_from_pc function")
560a81e ("MLK-21854-2 imx: uuu: Update autoboot to run uuu or fastboot")

Andrey commented on these two:
https://github.com/Freescale/u-boot-fslc/pull/41#issuecomment-871548961

81e786c ("pico-imx6ul, pico-imx7d: Explain how to use eMMC user partition")
A simple one, Otavio. You should upstream it :-)

2f7d389 ("mx6sabre_common: Dynamically change the video output")
6b4b7e0 ("mx6sabresd: Enable video interfaces in bootargs")

Only affects the NXP kernel, not the mainline kernel. Both should be dropped.

IMHO Andrey's patch moves us in the right direction.

recipes-bsp/u-boot/u-boot_%.bbappend
```diff
@@ -0,0 +1,8 @@
+# So far, we just need a container to be inherited
+inherit ${@oe.utils.ifelse(d.getVar('UBOOT_PROVIDES_BOOT_CONTAINER') == '1', 'imx-boot-container', '')}
+
+# Location known to imx-boot component, where U-Boot artifacts
+# should be additionally deployed.
+# See below note above do_deploy:append:mx8m for the purpose of
+# this delopyment location
+BOOT_TOOLS = "imx-boot-tools"
```
MaxKrummenacher, line 8: Hi Andrey

I wonder if we should move the setting of BOOT_TOOLS to imx-base for all recipes using it.
The variable is used to coordinate the location of artifact providers (imx-atf imx-sc-firmware u-boot-imx) to its consumers (u-boot-fslc, imx-boot  and now u-boot).
I.e. all of these recipes must agree on the same location or the build will fail.

Max

zandrey: Agree, this would be the right thing to do. I can try this out and either push it to this PR or make a separate one.

Thanks for pointing it out! 👍🏻

zandrey:
Hello @otavio!

> I'd love to do it however, we still have patches that we need. How do you intend to do with them?

I'm glad that you've appreciated this!

As I've mentioned in the PR cover letter, the patch delta is (a) not significant, and this is what @fabioestevam has commented out already on; and (b) those patches do not affect `imx8m` derivatives, or at least those that are included in this PR.

I would really prefer those patches to be up-streamed, so the layer can drop this technical dept and move on with vanilla _U-Boot_. I however cannot upstream those myself since I simply do not have a HW to test those patches before submission (I do not have i.MX6 or i.MX7 boards), hence if somebody willing to step up and help here - it would be really appreciated.

Boards that I could verify I've included in this PR, they are all bootable all the way with vanilla _U-Boot_, which I tested. _i.MX8MQ_ is on the way, I've [submitted the patch](https://lists.denx.de/pipermail/u-boot/2021-December/468983.html) directly upstream in an effort to include it the same way as I did for all other derivatives here.

I guess there is nothing that is currently preventing the Mainline BSP to move on to vanilla U-Boot for those derivatives, at least nothing that I can see.

-- andrey

otavio:
Some patches are NXP specific which we need to maintain. I wonder if we do not face conflicts for the patches we'd need to keep in the layer too.

zandrey:
> Some patches are NXP specific which we need to maintain. I wonder if we do not face conflicts for the patches we'd need to keep in the layer too.

This might be true for derivatives, which are not covered by this RFC.

I'm not proposing to drop the fork now, rather to set the trend and having those boards using vanilla U-Boot would serve as a good example IMHO.

Besides, 3rd parties who already have their boards supported in upstream can use the same mechanism to provide Mainline BSPs.

fabioestevam:
On Fri, Dec 3, 2021 at 6:31 PM Otavio Salvador ***@***.***> wrote:
>
> Some patches are NXP specific which we need to maintain. I wonder if we do not face conflicts for the patches we'd need to keep in the layer too.

Which specific patches are you referring to?

Do you mean these two?

2d107f6 ("imx: iMX6: Define disconnect_from_pc function")
560a81e ("MLK-21854-2 imx: uuu: Update autoboot to run uuu or fastboot")

otavio:
The two were referred by @fabioestevam as well as maybe others. We'd need to keep those patches on the layer.

fabioestevam:
On Fri, Dec 3, 2021 at 7:06 PM Otavio Salvador ***@***.***> wrote:
>
> The two were referred by @fabioestevam as well as maybe others. We'd need to keep those patches on the layer.

I would like to understand the need for these two patches. Not sure
what "maybe others" mean.

Looking at:

commit 560a81e9cb729cb08f539852db77db604600f657
Author: Ye Li ***@***.***>
Date:   Fri May 17 01:52:27 2019 -0700

    MLK-21854-2 imx: uuu: Update autoboot to run uuu or fastboot

    When booting from USB device, change u-boot autoboot to default
    run UUU bootcmd or enter fastboot directly for android.

    Signed-off-by: Ye Li ***@***.***>
    (cherry picked from commit eba8710106348a283940b851080b049bc7fbe466)
    (cherry picked from commit decfa984826db8ae39e93f2968b8a04d009eef0d)

It seems to me that the logic added inside common/autoboot.c should be placed
inside board files instead, just like it is done at the boards below
in U-Boot mainline:

board/out4/o4-imx6ull-nano/o4-imx6ull-nano.c:
board/toradex/apalis_imx6/apalis_imx6.c
board/toradex/colibri-imx6ull/colibri-imx6ull.c
board/toradex/colibri_imx6/colibri_imx6.c
board/toradex/colibri_imx7/colibri_imx7.c

otavio:
Doesn't make much sense to me to every board to implement the logic; instead we ought to have a default behavior.

fabioestevam:
On Sat, Dec 4, 2021 at 8:48 AM Otavio Salvador ***@***.***> wrote:
>
> Doesn't make much sense to me to every board to implement the logic; instead we ought to have a default behavior.

Is there a  valid "default behaviour" for all i.MX boards that enter
USB serial download mode? Looks like more of a  board decision.

I am still trying to understand the need for 560a81e ("MLK-21854-2
imx: uuu: Update autoboot to run uuu or fastboot") inside u-boot-fslc.

+#if defined(is_boot_from_usb)
+       if (is_boot_from_usb() && env_get("bootcmd_mfg")) {

In u-boot-fslc the only imx6/imx7 board that defines bootcmd_mfg is
imx7d-sabresd.

So as of now in u-boot-fslc the common behavior inside
common/autoboot.c only applies to a single board.

I am not convinced that 560a81e ("MLK-21854-2 imx: uuu: Update
autoboot to run uuu or fastboot") should be kept only in u-boot-fslc.

We could come up with a solution in mainline U-Boot for it too. I am
willing to help. I just need a use concrete use-case
that I can test.

otavio:
@fabioestevam kind of. The only board with bootcmd_mfg is indeed sabresd but any board can define it and it'd just work. 

In a more generic solution, I think, i.MX would need to provide a `boot-source` command and then we'd use it inside scripts to allow any behavior need.

zandrey:
> @fabioestevam kind of. The only board with bootcmd_mfg is indeed sabresd but any board can define it and it'd just work.
> 
> In a more generic solution, I think, i.MX would need to provide a `boot-source` command and then we'd use it inside scripts to allow any behavior need.

@otavio Latest U-Boot source base have a _distro boot_ enabled per default, which iterates between extlinux, boot script, and NFS boot IIRC. The behavior that is present in `560a81e ("MLK-21854-2 imx: uuu: Update autoboot to run uuu or fastboot")` commit can be moved to the script of the corresponding board, thus making `autoboot` change unnecessary.

Every board can define it's boot flow in the _distro boot_ chain, so why do we still need to keep this piggy-back in `autoboot` command?

In such case, the only thing that would be required is to provide a `usb` sub-command, which should cover `disconnect_from_pc()` call.

Would that be an option to satisfy every derivative?

-- andrey

otavio:
Yes @zandrey, this would be a good option. We'd need to get the patches on the layer and get them applied.

zandrey:
@fabioestevam I've also seen your [imx8mm: Add serial download support](https://lore.kernel.org/u-boot/20210703195837.28153-1-festevam@denx.de/) patchset on U-Boot ML. Would that be something to cover the cases which those remaining patches are trying to address?

As I understand, those 2 remaining patches brings a possibility to switch to `uuu` or `fastboot`. `uuu` should be covered by your patchset, while `fastboot` is already supported in upstream. The only thing that would be left is to select the boot mode, and this can be done via boot script which is a part of _distro boot_ chain.

Do I understand it correct?

-- andrey

fabioestevam:
Hi Andrey,

On Sat, Dec 4, 2021 at 7:19 PM Andrey Zhizhikin
***@***.***> wrote:
>
> @fabioestevam I've also seen your imx8mm: Add serial download support patchset on U-Boot ML. Would that be something to cover the cases which those remaining patches are trying to address?
>
> As I understand, those 2 remaining patches brings a possibility to switch to uuu or fastboot. uuu should be covered by your patchset, while fastboot is already supported in upstream. The only thing that would be left is to select the boot mode, and this can be done via boot script which is a part of distro boot chain.
>
> Do I understand it correct?

The following two patches that are still present in u-boot-fslc:

2d107f6 ("imx: iMX6: Define disconnect_from_pc function")
560a81e ("MLK-21854-2 imx: uuu: Update autoboot to run uuu or fastboot")

only affects imx6/imx7/imx7ulp, not imx8m.

Please note that is_boot_from_usb is only defined for imx6/imx7/imx7ulp.

The patch I sent for adding serial download support to imx8m still
needs to be reworked, but this is a separate problem.

Regards,

Fabio Estevam

zandrey:
Hello Fabio,

On Tue, Dec 7, 2021 at 12:07 PM fabioestevam ***@***.***> wrote:
>
> Hi Andrey,
>
> On Sat, Dec 4, 2021 at 7:19 PM Andrey Zhizhikin
> ***@***.***> wrote:
> >
> > @fabioestevam I've also seen your imx8mm: Add serial download support patchset on U-Boot ML. Would that be something to cover the cases which those remaining patches are trying to address?
> >
> > As I understand, those 2 remaining patches brings a possibility to switch to uuu or fastboot. uuu should be covered by your patchset, while fastboot is already supported in upstream. The only thing that would be left is to select the boot mode, and this can be done via boot script which is a part of distro boot chain.
> >
> > Do I understand it correct?
>
> The following two patches that are still present in u-boot-fslc:
>
> 2d107f6 ("imx: iMX6: Define disconnect_from_pc function")
> 560a81e ("MLK-21854-2 imx: uuu: Update autoboot to run uuu or fastboot")
>
> only affects imx6/imx7/imx7ulp, not imx8m.

Understood, thanks for clarification.

>
> Please note that is_boot_from_usb is only defined for imx6/imx7/imx7ulp.

I was tripped a bit by commit "usb: ehci-mx6: move mode set/detect to
probe" [1], but now I see it is not really related to those 2 original
commits.

I guess that 2d107f6 ("imx: iMX6: Define disconnect_from_pc function")
can be submitted as-is to upstream, since similar function is already
present in arch/arm/include/asm/arch-mx7/imx-regs.h
and arch/arm/include/asm/arch-mx7ulp/imx-regs.h.

Patch 560a81e ("MLK-21854-2 imx: uuu: Update autoboot to run uuu or
fastboot") however need to be re-worked, but I did not figure out the
proper way yet. Need to think about this one further on.

>
> The patch I sent for adding serial download support to imx8m still
> needs to be reworked, but this is a separate problem.

Yes, I see it now.

>
> Regards,
>
> Fabio Estevam
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly, view it on GitHub, or unsubscribe.


Link: [1]: ***@***.***/
-- 
Regards,
Andrey.

zandrey:
@otavio 
As there is a U-Boot update in #950, I will close this PR for now. I'll keep the branch in my fork as a PoC for future references.

-- andrey

PR #952, opened 2022-01-19
[imx8mq-evk]: Add Mainline BSP support

zandrey:
Recent versions of upstream U-Boot and Kernel do provide support for i.MX8MQ derivative, and it can be built using Mainline BSP.

Convert machine definition to include all parts, that are required to build Mainline BSP:
- Drop hardcoded BSP flavor setting in machine configuration
- Add boot container machine override
- Split machine features
- Define ATF load address
- Add dependency to firmware package
- Provide `extlinux` configuration to utilize _distro boot_ feature

Tested on `imx8mq-evk` machine for both `nxp` and `mainline` setting in `IMX_DEFAULT_BSP`.

-- andrey

### DIFF
--- a/conf/machine/imx8mq-evk.conf
+++ b/conf/machine/imx8mq-evk.conf
@@ -4,15 +4,18 @@
 #@DESCRIPTION: Machine configuration for NXP i.MX 8M Quad Evaluation Kit
 #@MAINTAINER: Jun Zhu <junzhu@nxp.com>
 
-MACHINEOVERRIDES =. "mx8:mx8m:mx8mq:"
+MACHINEOVERRIDES =. "imx-boot-container:mx8:mx8m:mx8mq:"
 
 require conf/machine/include/imx-base.inc
 require conf/machine/include/arm/armv8a/tune-cortexa53.inc
 
-IMX_DEFAULT_BSP = "nxp"
+# Mainline BSP defaults to "generic" cortexa53 configuration,
+# adjust it here to include crypto extension which enables
+# inline NEON and FPU code generation
+DEFAULTTUNE:use-mainline-bsp = "cortexa53-crypto"
 
-MACHINE_FEATURES += "pci wifi bluetooth optee bcm43455 bcm4356"
-MACHINE_FEATURES:append:use-nxp-bsp = " bcm4359"
+MACHINE_FEATURES += "pci wifi bluetooth bcm43455 bcm4356"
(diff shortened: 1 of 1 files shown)


conf/machine/imx8mq-evk.conf
```diff
@@ -4,40 +4,47 @@
...
+# adjust it here to include crypto extension which enables
+# inline NEON and FPU code generation
+DEFAULTTUNE:use-mainline-bsp = "cortexa53-crypto"
+
+MACHINE_FEATURES += "pci wifi bluetooth bcm43455 bcm4356"
+MACHINE_FEATURES:append:use-nxp-bsp = " optee bcm4359"
 
 MACHINE_SOCARCH_FILTER:append:mx8mq = " virtual/libopenvg virtual/libgles1 virtual/libgles2 virtual/egl virtual/mesa virtual/libgl virtual/libg2d"
 
+KERNEL_DEVICETREE_BASENAME = "imx8mq-evk"
```
otavio, line 24: What is the goal of this variable? it doesn't vary across nxp and mainline so it adds a indirection.

conf/machine/imx8mq-evk.conf
```diff
@@ -4,40 +4,47 @@
...
 #@MAINTAINER: Jun Zhu <junzhu@nxp.com>
 
-MACHINEOVERRIDES =. "mx8:mx8m:mx8mq:"
+MACHINEOVERRIDES =. "imx-boot-container:mx8:mx8m:mx8mq:"
 
 require conf/machine/include/imx-base.inc
 require conf/machine/include/arm/armv8a/tune-cortexa53.inc
 
-IMX_DEFAULT_BSP = "nxp"
+IMX_DEFAULT_BSP = "mainline"
```
otavio, line 12: We need to remove this setting here and change imx-base.inc.

conf/machine/imx8mq-evk.conf
zandrey: Correct, will drop it.

conf/machine/imx8mq-evk.conf
zandrey: Make sense, will re-work it in _V2_.

zandrey:
Changes in _V2_:
- Dropped hardcoded BSP flavor, it is already using default `"mainline"` value from _imx-base.inc_
- Drop DTB list split, it does not differ between `mainline` and `nxp` flavors

otavio (approved):

PR #956, opened 2022-01-21
imx-gpu-viv: Limit dependency on kernel-module-imx-gpu-viv

thochstein:
Building core-image-weston with multilib support and builtin graphics
fails:

```
ERROR: Trying to resolve runtime dependency kernel-module-imx-gpu-viv resulted in conflicting PREFERRED_PROVIDER entries being found.
The providers found were: ['virtual:multilib:lib32:/opt/work/upstream/sources/meta-freescale/recipes-kernel/linux/linux-imx-mfgtool_5.10.bb', '/opt/work/upstream/sources/meta-freescale/recipes-kernel/linux/linux-imx-mfgtool_5.10.bb']
The PREFERRED_PROVIDER entries resulting in this conflict were: ['PREFERRED_PROVIDER_lib32-linux-imx-mfgtool = lib32-linux-imx-mfgtool', 'PREFERRED_PROVIDER_linux-imx-mfgtool = linux-imx-mfgtool']. You could set PREFERRED_RPROVIDER_kernel-module-imx-gpu-viv
ERROR: Required build target 'core-image-weston' has no buildable providers.
Missing or unbuildable dependency chain was: ['core-image-weston', 'matchbox-terminal', 'gtk+3', 'virtual/egl', 'kernel-module-imx-gpu-viv']
```

This is the configuration added in local.conf:

```
require conf/multilib.conf
MULTILIBS = "multilib:lib32"
DEFAULTTUNE:virtclass-multilib-lib32 = "armv7athf-neon"
IMAGE_INSTALL:append = " lib32-glibc lib32-libgcc lib32-libstdc++"
MACHINE_USES_VIVANTE_KERNEL_DRIVER_MODULE = "0"
```

The error is confusing since it refers to _module_ graphics and to
linux-imx-mfgtool, neither of which are relevant to the build.

Rework the implementation so that the module graphics are recommended
only when configured and the logic marking the kernel as a provider of
module graphics can be dropped. Also, move the dependency to the
machine level to avoid making imx-gpu-viv machine-specific.

Signed-off-by: Tom Hochstein <tom.hochstein@nxp.com>

PR #962, opened 2022-01-25
fsl-vivante-kernel-driver-handler.bbclass: Fix multilib support

otavio:
Building core-image-weston with multilib support and builtin graphics
fails:

```
ERROR: Nothing PROVIDES 'virtual/arm-oelmllib32-linux-gnueabi-binutils'. Close matches:
  virtual/lib32-arm-oelmllib32-linux-gnueabi-binutils
  virtual/lib32-arm-oelmllib32-linux-gnueabi-go-runtime
  virtual/lib32-arm-oelmllib32-linux-gnueabi-rust
ERROR: Nothing PROVIDES 'virtual/arm-oelmllib32-linux-gnueabi-gcc'. Close matches:
  virtual/lib32-arm-oelmllib32-linux-gnueabi-g++
  virtual/lib32-arm-oelmllib32-linux-gnueabi-gcc
  virtual/lib32-arm-oelmllib32-linux-gnueabi-rust
```

This is the configuration added in local.conf:

```
require conf/multilib.conf
MULTILIBS = "multilib:lib32"
DEFAULTTUNE:virtclass-multilib-lib32 = "armv7athf-neon"
IMAGE_INSTALL:append = " lib32-glibc lib32-libgcc lib32-libstdc++"
MACHINE_USES_VIVANTE_KERNEL_DRIVER_MODULE = "0"
```

This commit fixes the provides so BitBake can properly handle the
dependencies.

### DIFF
--- a/classes/fsl-vivante-kernel-driver-handler.bbclass
+++ b/classes/fsl-vivante-kernel-driver-handler.bbclass
@@ -41,9 +41,9 @@ python fsl_vivante_kernel_driver_handler () {
         return
 
     if use_vivante_kernel_driver_module != "1":
-        e.data.appendVar('RPROVIDES:${KERNEL_PACKAGE_NAME}-base', ' kernel-module-imx-gpu-viv')
-        e.data.appendVar('RREPLACES:${KERNEL_PACKAGE_NAME}-base', ' kernel-module-imx-gpu-viv')
-        e.data.appendVar('RCONFLICTS:${KERNEL_PACKAGE_NAME}-base', ' kernel-module-imx-gpu-viv')
+        e.data.appendVar('RPROVIDES:${KERNEL_PACKAGE_NAME}-base', ' ${MLPREFIX}kernel-module-imx-gpu-viv')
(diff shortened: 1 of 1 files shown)


thochstein:
This solution doesn't address the root cause, a runtime recommendation for a kernel module that doesn't exist. Removing the incorrect runtime recommendation allows us to remove the 'trick' of having the kernel provide a kernel module, resulting in a much simpler design.

otavio:
The runtime recommendation isn't wrong. It is needed to have the dependencies sort out automatically. Is there a test case for the use case you still see the error?

thochstein:
This solution works.

The runtime recommendation for the builtin case _is_ wrong if it solves no problem and only complicates the recipes. What problem do you see with #956 that is addressed by this solution?

otavio:
The dependencies here are important so galcore is automatically added to the image; the removal would require manual work from user.

thochstein:
Thanks, Otavio. As we discussed, the problem with #956 is that it pulls in graphics automatically even for headless boards. I guess we justify the complicated solution here for this optimization.

Can you backport please to `honister`?

Issue #791, opened 2021-05-21
UBOOT_ENTRYPOINT not set for mx8m machines built with mainline bsp


pberginkonsult:
When building mainline bsp for, as example, imx8mm_lpddr4_evk the variable UBOOT_ENRTYPOINT is not set. This causes the fitImage generation to be corrupt as 'Load Address' in the uImage header gets wrong value.

The MACHINEOVERRIDES mx8m was removed from mainline bsp in the commit 069482f24b3e07c8118f4cf9af7564220ac5e82b. This caused that this change ac25d3a3d09a2707d9ee614d3ed506fd2ebe2f83, setting DEFAULTTUNE and UBOOT_ENTRYPOINT for mx8m machines will not take effect for mainline bsp.

zandrey:
@pberginkonsult

`master` branch contains U-Boot `2021.07-rc1`, which has the _distro boot_ enabled by default. In this case, I'm not sure if the kernel FIT load support is still enabled and operable.

In fact, even before ac25d3a3d09a2707d9ee614d3ed506fd2ebe2f83 ("imx-base.inc: Consolidate DEFAULTTUNE and UBOOT_ENTRYPOINT for 8M") was merged - original `MACHINEOVERRIDES` were anyway filtered out in the _mainline_ BSP, which suggests that it was not working even before.

Is it necessary that you use FIT image load, or maybe you can switch to use _extlinux_ instead to align with upstream U-Boot? You can see the 122bfa162197075949adf826a39119cc21c6955f ("conf: machine: imx8m[m,n,p]: enable extlinux support") for the reference and reasoning.

-- andrey

pberginkonsult:
> In fact, even before [ac25d3a](https://github.com/Freescale/meta-freescale/commit/ac25d3a3d09a2707d9ee614d3ed506fd2ebe2f83) ("imx-base.inc: Consolidate DEFAULTTUNE and UBOOT_ENTRYPOINT for 8M") was merged - original `MACHINEOVERRIDES` were anyway filtered out in the _mainline_ BSP, which suggests that it was not working even before.
>
Could be tha case, haven't checked. 
[...]

PR #980, opened 2022-02-19
Generalize overrides subsystem for NXP and Mainline support

otavio:
Essentially, we extend the overrides to a generic-bsp, nxp-bsp, and mainline-bsp.

So, for example, the mx8mq override is split into:

- imx-generic-bsp: compatible with every i.MX SoC and both BSP variants
- imx-nxp-bsp: compatible with every i.MX SoC but specific to NXP BSP
- imx-mainline-bsp: compatible with every i.MX SoC but specific to Mainline BSP

- mx8-generic-bsp: compatible with every i.MX8 SoC and both BSP variants
- mx8-nxp-bsp: compatible with every i.MX8 SoC but specific to NXP BSP
- mx8-mainline-bsp: compatible with every i.MX8 SoC but specific to Mainline BSP

- mx8m-generic-bsp: compatible with every i.MX8M SoC and both BSP variants
- mx8m-nxp-bsp: compatible with every i.MX8M SoC but specific to NXP BSP
- mx8m-mainline-bsp: compatible with every i.MX8M SoC but specific to Mainline BSP

- mx8mq-generic-bsp: compatible with every i.MX8MQ SoC and both BSP variants
- mx8mq-nxp-bsp: compatible with every i.MX8MQ SoC but specific to NXP BSP
- mx8mq-mainline-bsp: compatible with every i.MX8MQ SoC but specific to Mainline BSP

The extender mechanism is responsible for extending the override list to include the generic overrides. We can then use the three different variants to handle the metadata correctly.

Generically speaking, the conversion mainly was automated (with a lot of back and forth until getting it right).

To convert an existing layer, the following script can be used:
```sh
  git ls-files classes recipes-* \
      | xargs sed -i \
              -e 's,:\(mx[6-8]\w*\),:\1-nxp-bsp,g' \
              -e 's,(\(mx[6-8]\w*\)),(\1-nxp-bsp),g' \
              -e 's,\(mx[6-8]\w*\)|,\1-nxp-bsp|,g' \
              -e 's,|\(mx[6-8]\w*\)),|\1-nxp-bsp),g' \
              \
              -e 's,:\(mx[5s]\w*\),:\1-generic-bsp,g' \
              -e 's,(\(mx[5s]\w*\)),(\1-generic-bsp),g' \
              -e 's,\(mx[5s]\w*\)|,\1-generic-bsp|,g' \
              -e 's,|\(mx[5s]\w*\)),|\1-generic-bsp),g' \
              \
              -e 's,:\(vf\w*\),:\1-generic-bsp,g' \
              -e 's,:\(vf[56]0\w*\),:\1-generic-bsp,g' \
              -e 's,\(vf\w*\)|,\1-generic-bsp|,g' \
              -e 's,|\(vf\w*\)),|\1-generic-bsp),g' \
              -e 's,\(vf[56]0\w*\)|,\1-generic-bsp|,g' \
              -e 's,|\(vf[56]0\w*\)),|\1-generic-bsp),g' \
              \
              -e 's,:\(imx\) ,:\1-nxp-bsp ,g' \
              -e 's,(\(imx\)),(\1-nxp-bsp),g' \
              -e 's,\(imx\)|,\1-nxp-bsp|,g' \
              -e 's,|\(imx\)),|\1-nxp-bsp),g'

  for d in $(find -type d | egrep '/mx[6-8]w*'); do
      git mv $d $d-nxp-bsp
  done

  for d in $(find -type d | egrep '/imx$'); do
      git mv $d $d-nxp-bsp
  done

  for d in $(find -type d | egrep '/mx[5s]w*'); do
      git mv $d $d-generic-bsp
  done
```

Fixes: #791.


otavio:
@zandrey @thochstein @angolini 

Please review this PR very carefully; it changes a lot of internal logic and certainly we can later rework code to reduce duplication but I tried to reduce logical changes.

MaxKrummenacher:
Thanks for this. This will make live much easier.

For me weston-init for use-nxp-bsp failed as `imx` is used for a FILESOVERRIDES. Renaming the override directory fixed it for me, but there are more recipes affected.
```
git mv recipes-graphics/wayland/weston-init/imx recipes-graphics/wayland/weston-init/use-nxp-bsp

find . -name imx
./dynamic-layers/openembedded-layer/recipes-graphics/xserver-common/xserver-common/imx
./recipes-kernel/linux/linux-imx/imx
./recipes-kernel/linux/linux-fslc-imx/imx
./recipes-bsp/alsa-state/alsa-state/imx
./recipes-multimedia/pulseaudio/pulseaudio/imx
```
Can you work that into this patchset?

otavio:
@MaxKrummenacher 
> Thanks for this. This will make live much easier.

I believe so. It's also easier to express what we want using the new overrides and reduce risk of misunderstandings.
  
> For me weston-init for use-nxp-bsp failed as `imx` is used for a FILESOVERRIDES. Renaming the override directory fixed it for me, but there are more recipes affected.
...
> Can you work that into this patchset?

I did this; could you review and test it one more time?

MaxKrummenacher:
@otavio 
I started a build. Looks like the ./recipes-bsp/alsa-state/alsa-state/imx and ./recipes-multimedia/pulseaudio/pulseaudio/imx directories were false positives. This is what gets added to FILESEXTRAPATHS and are not FILESOVERRIDES. Why we need that additional directory I do not know. At least the pulseaudio recipe now couldn't find the needed local files.
Apart from that, the build is still running. Will report tomorrow.

thochstein:
The design is clearer. Thanks @otavio.

I wonder though if the naming can be simplified. Do we need `-bsp`? Do we need `-generic`, or can we just say that an override with no qualifier is implicitly generic?

```
imx-generic-bsp  -> imx
imx-nxp-bsp      -> imx-nxp
imx-mainline-bsp -> imx-mainline
```

recipes-bsp/u-boot/u-boot-imx_2021.04.bb
```diff
@@ -38,4 +38,4 @@ do_deploy:append:mx8m() {
 }
 
 PACKAGE_ARCH = "${MACHINE_ARCH}"
-COMPATIBLE_MACHINE = "(mx6|mx7|mx8|use-mainline-bsp)"
+COMPATIBLE_MACHINE = "(mx6-generic-bsp|mx7-generic-bsp|mx8-generic-bsp)"
```
zandrey, line 41: Not sure if the `*-generic-bsp` is required here, perhaps `*-nxp-bsp` would be sufficient - _Mainline BSP_ does use `u-boot-fslc` recipe instead.

recipes-bsp/imx-atf/imx-atf_2.4.bb
```diff
@@ -67,4 +67,4 @@ do_deploy() {
 addtask deploy after do_compile
 
 PACKAGE_ARCH = "${MACHINE_SOCARCH}"
-COMPATIBLE_MACHINE = "(mx8|imx-boot-container)"
+COMPATIBLE_MACHINE = "(mx8-generic-bsp)"
```
zandrey, line 70: As a "wishful thinking": can we pin it to something which is not _generic_? This would allow to build upstream TF-A from _meta-arm_ for `mx8m`, which is sometimes beneficial for layer's users...

I might look into how this can be done.

recipes-security/optee-imx/optee-os_3.10.0.imx.bb
```diff
@@ -93,4 +93,4 @@ FILES:${PN}-staticdev = "/usr/include/optee/"
 RDEPENDS:${PN}-dev += "${PN}-staticdev"
 
 PACKAGE_ARCH = "${MACHINE_ARCH}"
-COMPATIBLE_MACHINE = "(imx|imx-boot-container)"
+COMPATIBLE_MACHINE = "(imx-nxp-bsp|imx-boot-container)"
```
zandrey, line 96: Since I see that `imx-boot-container` is dropped from all other `COMPATIBLE_MACHINE` - perhaps it can be dropped from here as well.

I never got the NXP OP-TEE working with Boot Container, and it is not packed at the moment. I'd check if that can safely be removed, and if needed - re-introduced later.

zandrey (commented):
@otavio This one looks good, there is a really clear separation done in this PR.

Besides few small things below, which I mainly would like to test myself - it's fine for me.

I'll start few `mx8m` builds for both BSP flavors to see if I get some errors along the way.

Thanks a lot!

recipes-bsp/u-boot/u-boot-imx_2021.04.bb
otavio: Generally yes, but I used `generic` here because we eventually use `imx` one while converting a machine mainline. Also, other BSP might be using this as the `mainline` was accepted.

recipes-bsp/imx-atf/imx-atf_2.4.bb
otavio: We can, but I believe it would need to be a separated patch, right?

otavio:
> The design is more apparent. Thanks @otavio.
> 
> I wonder though if the naming can be simplified. Do we need `-bsp`? Do we need `-generic`, or can we just say that an override with no qualifier is implicitly generic?
> 
> ```
> imx-generic-bsp  -> imx
> imx-nxp-bsp      -> imx-nxp
> imx-mainline-bsp -> imx-mainline
> ```

`-bsp` suffix has been added on purpose as we'd like to use a clear indication of what it is used for.

otavio:
> @otavio I started a build. Looks like the ./recipes-bsp/alsa-state/alsa-state/imx and ./recipes-At least the pulseaudio recipe now couldn't find the needed local files. Apart from that, the build is still running. Will report tomorrow.

Fixed. I changed the `.bbappend` @MaxKrummenacher

recipes-security/optee-imx/optee-os_3.10.0.imx.bb
otavio: I removed it.

thochstein (commented):
There are typos in the commit message for the `-nxp-bsp` examples, like `i.MX SoC8M`.

recipes-bsp/imx-atf/imx-atf_2.4.bb
zandrey: Yes, this can be done separate after this PR is merged.

thochstein:
> @otavio thanks a lot for taking time to think about this issue, and to enhance the class. I'm reviewing and I'm thinking about the changes Could you, please, clarify one point? How many BSPs do we support?

Perhaps the main commit message could get a little more background?
- What BSPs are supported, with a definition
- How the BSPs are supported with the existing design (I'm think about SOC overrides like mx8m limited to NXP BSP, right?)
- What the problems are with the existing design and how the new design fixes it

thochstein:
Besides these choices, there are also the choices provided by the distros. It would be good if there were a README laying out these configuration choices for the user.

thochstein:
There are corresponding changes to be made in meta-freescale-distro: https://github.com/Freescale/meta-freescale-distro/pull/83

otavio:
@thochstein
> There are typos in the commit message for the `-nxp-bsp` examples, like `i.MX SoC8M`.

Thanks; fixed.

> @otavio thanks a lot for taking time to think about this issue, and to enhance the class. I'm reviewing and I'm thinking about the changes Could you, please, clarify one point? How many BSPs do we support?

1. `mainline`: which uses full mainline support
2. `nxp`: which uses u-boot-fslc (when possible) and NXP components (with FSLC fixes)

We also have support for NXP "pristine" release forcing the `-imx`  (without FSLC fixes) - the `fsl` distro.

angolini:
>     1. `mainline`: which uses full mainline support 
>     2. `nxp`: which uses u-boot-fslc (when possible) and NXP components (with FSLC fixes)
> We also have support for NXP "pristine" release forcing the `-imx` (without FSLC fixes) - the `fsl` distro.

It makes me think we support 2 BSPs and some distros.

I understand what the *nxp* BSP is composed by packages from `imx-generic-bsp` plus packages from `imx-nxp-bsp`.

When you say "generic", do you mean common? (common for both supported BSPs)

otavio:
Yes. The generic is intended being used by any i.MX machine (without being BSP specific).

angolini:
Maybe, somehow, we should document this stuff

recipes-kernel/linux/linux-fslc-imx_5.10.bb
```diff
@@ -87,4 +87,4 @@ LOCALVERSION = "-5.10.52-2.1.0"
 
 DEFAULT_PREFERENCE = "1"
 
-COMPATIBLE_MACHINE = "(mx6|mx7|mx8)"
```
angolini, line 90: Why `linux-fslc-imx` is compatible only with `*-nxp-bsp`?

otavio:
I'd love some help on this (@angolini I am looking at you <3)

angolini:
> I'd love some help on this (@angolini I am looking at you <3)

When I first said, I was volunteering <3 @otavio 

But at first I need to understand how to describe all of this to one not familiar with this. I'll be thinking about this

recipes-security/optee-imx/optee-test_3.10.0.imx.bb
```diff
@@ -60,4 +60,4 @@ do_install () {
 
 FILES:${PN} = "/usr/bin/ /lib*/optee_armtz/"
 
-COMPATIBLE_MACHINE = "(mx6-nxp-bsp|mx7-nxp-bsp|mx8-nxp-bsp)"
+COMPATIBLE_MACHINE = "(imx-nxp-bsp)"
```
angolini, line 63: We have `uboot-imx` on generic-bsp, but we have `optee` on nxp-bsp. Is there a reason?

recipes-kernel/linux/linux-fslc-imx_5.10.bb
otavio: Do you see a reason to use linux-fslc-imx with mainline BSP?

recipes-security/optee-imx/optee-test_3.10.0.imx.bb
otavio: Can it be used with mainline nowadays?

otavio: It was restricted; so I kept it as well.

recipes-kernel/linux/linux-fslc-imx_5.10.bb
angolini: I don't, but if I understand it right, it was possible to use it before and now it is not possible any more.
Anyway, nothing that a future PR would not add back

otavio: It wasn't as `use-mainline-bsp` was not accepted. So I just kept as before.

angolini: ok

otavio:
Hello everyone,

Could you guys do a run check on this so we can consider merging it?

zandrey:
@otavio 

For `use-mainline-bsp` and command-line only image:

```
Tested-by: Andrey Zhizhikin <andrey.z@gmail.com> # imx8mm-lpddr4-evk, imx8mn-ddr4-evk, imx8mp-lpddr4-evk, imx8mq-evk
```

Boot-tested all machines above, no issues found.

Layer configuration used in the test build is this PR branch + local modifications to set layer compatibility with "_kirkstone_" (did a build before PR #981 was merged).

-- andrey

thochstein:
@otavio, sorry I'm still working through other kirkstone build breaks, hope to get to it soon today.

thochstein:
I did a community build (no `meta-imx`) of `core-image-weston` using `fsl-xwayland/imx8mq-evk`. It booted successfully to `weston`.

Now I'm doing an 'NXP' build (with `meta-imx`).

MaxKrummenacher:
I did successfully build test the Toradex machines.
I did successfully install and run the images on:
Colibri iMX6, i.MX 6DL with use-mainline-bsp
Apalis iMX6, i.MX 6Q with use-nxp-bsp
Apalis  iMX8, i.MX 8QM with use-nxp-bsp
weston and glmark2-es2-wayland work fine.
(weston will need some further tweaks but those are unrelated to the changes in this patchset)

PR #982, opened 2022-02-23
imx-base.inc: Add support to include extra data for NIP devices

otavio:


thochstein (approved):
Thanks!

PR #991, opened 2022-02-28
imx-base.inc: add imx-generic-bsp override before extender

sapiippo:
imx-generic-bsp is common to all SoCs, add it to overrides without
the extender. This allows the override to be used before the extender
has run e.g., when including common inc file:

SOC_INCLUDE:imx-generic-bsp = "imx.inc"
include "${SOC_INCLUDE}"


otavio:
@sapiippo this seems to not be something we should support in the BSP support code.

I understand why you'd like to do this but as for the other BSP you can tie it to machine override (machine name) or use the new i.MX override (https://github.com/Freescale/meta-freescale/blob/9a926da11383733c50073f7e3e7fc08c9ed064fe/conf/machine/include/imx-base.inc#L521-L522).

PR #992, opened 2022-03-02
dynamic-layers: Update SOC overrides

thochstein:
Fix dynamic layer overrides missed by the script.


otavio (approved):
Thanks for the fix. I'll fix this in the upcoming helper script as well.

Issue #990, opened 2022-02-28
The compatible issue for -generic-bsp overriding


junzhuimx:
The issue happens on the commit -- https://github.com/Freescale/meta-freescale/commit/30f5ce1d2168389295b4813e01898c2fb0f9bab9 
It make a lot of generic SOC overriding to be NXP BSP specific.

Recently, the a lot of recipes with "imx/mx6/mx7/mx8/mx8m" SOC overides are generic. But use "imx-generic-bsp/mx6-generix-bsp/....." to override these setting
- it is not necessary: make the recipe line longer and easy to typo
- 3rd-party/customer's meta layer can not compatible with latest changes and has no error raised by bitbake engine

otavio:
The provided PR assist in the convertion and also provide a script to do most of it.

PR #994, opened 2022-03-02
Ensure we fail if old SoC overrides are in use

otavio:
To assist existing layers to convert to the new BSP-specific
overrides. Besides failing the parsing of the recipes where it is in
use, we provide a script to automate most of it.

Fixes: #990.

### DIFF
--- a/classes/machine-overrides-extender.bbclass
+++ b/classes/machine-overrides-extender.bbclass
@@ -55,3 +55,22 @@ python machine_overrides_extender_handler() {
 
 machine_overrides_extender_handler[eventmask] = "bb.event.ConfigParsed"
 addhandler machine_overrides_extender_handler
+
+python machineoverrides_filtered_out_qa_handler() {
+    filtered_out = (d.getVar('MACHINEOVERRIDES_EXTENDER_FILTER_OUT') or "").split()
+    qa_error = d.getVar('MACHINEOVERRIDES_FILTERED_OUT_QA_ERROR')
+
+    for var in d.overridedata:
+        # We need to allow the overrides being used in the extender
+        # so avoid processing it.
+        if 'MACHINEOVERRIDES_EXTENDER' in var:
+            continue
+
+        for (r, o) in d.overridedata[var]:
+            common = list(set(o.split(":")).intersection(filtered_out))
+            if len(common) > 0:
+                raise bb.parse.SkipRecipe(qa_error % common)
+}
+
+machineoverrides_filtered_out_qa_handler[eventmask] = "bb.event.RecipeParsed"
+addhandler machineoverrides_filtered_out_qa_handler
--- a/conf/machine/include/imx-base.inc
+++ b/conf/machine/include/imx-base.inc
@@ -224,6 +224,8 @@ MACHINEOVERRIDES_EXTENDER_FILTER_OUT = " \
     mx8dxl \
 "
 
+MACHINEOVERRIDES_FILTERED_OUT_QA_ERROR = "%s overrides cannot be used and need conversion to use the new BSP-specific overrides. Check 'meta-freescale/scripts/convert-bsp-specific-overrides'."
+
 # Sub-architecture support
 MACHINE_SOCARCH_SUFFIX ?= ""
 MACHINE_SOCARCH_SUFFIX:mx6q-nxp-bsp = "-mx6qdl"
--- a/scripts/convert-bsp-specific-overrides
+++ b/scripts/convert-bsp-specific-overrides
@@ -1,0 +1,87 @@
+#!/bin/sh
+# Convert old-style NXP overrides to new style BSP overides
+#
+# Essentially, we extend the overrides to a generic-bsp, nxp-bsp, and mainline-bsp.
+#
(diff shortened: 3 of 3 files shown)


conf/machine/include/imx-base.inc
```diff
@@ -224,6 +224,8 @@ MACHINEOVERRIDES_EXTENDER_FILTER_OUT = " \
     mx8dxl \
 "
 
+MACHINEOVERRIDES_FILTERED_OUT_QA_ERROR = "%s overrides cannot be used and need convertion to use the new BSP-specific overrides. Check 'meta-freescale/scripts/convert-bsp-specific-overrides'."
```
thochstein, line 227: convertion -> conversion

thochstein: One problem with the original script from the commit is that each time you run it it appends again. Is that fixed?

otavio: Honestly, I don't know how to fix this issue. Ideas?

thochstein: This looks like something to play with: https://stackoverflow.com/questions/4953738/how-to-globally-replace-strings-in-lines-not-starting-with-a-certain-pattern#answer-4953824

classes/machine-overrides-extender.bbclass
```diff
@@ -55,3 +55,22 @@ python machine_overrides_extender_handler() {
...
+    for var in d.overridedata:
+        # We need to allow the overrides being used in the extender
+        # so avoid processing it.
+        if 'MACHINEOVERRIDES_EXTENDER' in var:
+            continue
+
+        for (r, o) in d.overridedata[var]:
+            common = list(set(o.split(":")).intersection(filtered_out))
+            if len(common) > 0:
+                raise bb.parse.SkipRecipe(qa_error % common)
```
thochstein, line 72: I'm not familiar with how this will appear to the user. Can you give an example of what the user will see if a variable has an override like `:mx8`?

scripts/convert-bsp-specific-overrides
```diff
@@ -0,0 +1,76 @@
+#!/bin/sh
+# Generalize overrides subsystem for NXP and Mainline support
```
thochstein, line 2: I guess this should say what the script does now, something like `# Convert old-style NXP overrides like :mx8 to new style BSP overides like :mx8-nxp-bsp`

scripts/convert-bsp-specific-overrides
```diff
@@ -0,0 +1,76 @@
...
+            -e 's,|\(vf\w*\)),|\1-generic-bsp),g' \
+            -e 's,\(vf[56]0\w*\)|,\1-generic-bsp|,g' \
+            -e 's,|\(vf[56]0\w*\)),|\1-generic-bsp),g' \
+            \
+            -e 's,:\(imx\) ,:\1-nxp-bsp ,g' \
+            -e 's,(\(imx\)),(\1-nxp-bsp),g' \
+            -e 's,\(imx\)|,\1-nxp-bsp|,g' \
+            -e 's,|\(imx\)),|\1-nxp-bsp),g'
+
+for d in $(find -type d | egrep '/mx[6-8]w*'); do
```
thochstein, line 55: This section could have a comment that it is about converting override folder names to the new format.

classes/machine-overrides-extender.bbclass
otavio: It looks like:
```console
$ bitbake glmark2
Loading cache: 100% |                                                         | ETA:  --:--:--
Loaded 0 entries from dependency cache.
Parsing recipes: 100% |########################################################| Time: 0:00:14
Parsing of 1916 .bb files complete (0 cached, 1916 parsed). 2900 targets, 413 skipped, 0 masked, 0 errors.
Removing 1 recipes from the imx8mq_evk sysroot: 100% |#########################| Time: 0:00:00
ERROR: Nothing PROVIDES 'glmark2'
glmark2 was skipped: ['mx8'] overrides cannot be used and need convertion to use the new BSP-specific overrides. Check 'meta-freescale/scripts/contrib/convert-bsp-specific-overrides'.

Summary: There was 1 ERROR message, returning a non-zero exit code.
```

conf/machine/include/imx-base.inc
otavio: I am unsure about increasing the complexity to allow reuse. I foresee this script being used once in a layer to do the complete conversion and never again. When someone upstream a recipe, it is their responsibility to do it during the process.

thochstein: The error message does suggest the script to fix the problem. Please update the error message and in the comment at the top of the script that the script is intended for migrating existing layers to kirkstone and is not intended to be used more than once.

thochstein: Oh, perhaps you can add a simple grep to the script and fail if it finds new overrides?

angolini (approved):

conf/machine/include/imx-base.inc
otavio: I implemented this idea. Please review.

angolini (approved):

PR #999, opened 2022-03-07
imx-base.inc: add back mxs/mx23 support

twoerner:
When include/imx-base.inc was simplified to use the imx-{mainline,nxp}-bsp
overrides, the "mxs" and "mx23" categories of machines were missed.

Build tested with imx233-olinuxino-* and imx23evk.
Run tested with imx233-olinuxino-maxi.

### DIFF
--- a/conf/machine/include/imx-base.inc
+++ b/conf/machine/include/imx-base.inc
@@ -165,9 +165,10 @@ MACHINEOVERRIDES_EXTENDER:mx8dxl:use-nxp-bsp = "imx-generic-bsp:imx-nxp-bsp:mx8-
 ### Mainline BSP specific overrides
 #######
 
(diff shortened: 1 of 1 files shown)


conf/machine/include/imx-base.inc
```diff
@@ -165,6 +165,10 @@ MACHINEOVERRIDES_EXTENDER:mx8dxl:use-nxp-bsp = "imx-generic-bsp:imx-nxp-bsp:mx8-
 ### Mainline BSP specific overrides
 #######
 
+MACHINEOVERRIDES_EXTENDER:mxs:use-mainline-bsp   = "imx-generic-bsp:imx-mainline-bsp:mxs-generic-bsp:mxs-mainline-bsp"
+
```
otavio, line 169: The `mxs` can be dropped. It is equivalent to `mx5` or `mx6` which intends to group a set of machines. In this case, it'd group `mx23` and `mx28`.

otavio (changes requested):
Good catch, thanks for looking at it. I mentioned a small rework below...

otavio (approved):

PR #1007, opened 2022-03-13
weston-init: fix some invalid overrides

liuming50:
The mx8qm-nxp-bsp/mx8qxp-nxp-bsp overrides would not work because
imxgpu2d would take effect in the end since it has higher priority
in MACHINEOVERRIDES.

Append mx8qm-nxp-bsp/mx8qxp-nxp-bsp after imxgpu2d can fix the problem.

### DIFF
--- a/recipes-graphics/wayland/weston-init.bbappend
+++ b/recipes-graphics/wayland/weston-init.bbappend
@@ -28,8 +28,8 @@ INI_UNCOMMENT_ASSIGNMENTS:append:mx8mq-nxp-bsp = " \
 # Ideally, this should be seamless and Vivante ought to handle it internally and take the fastest
 # rendering code.
 INI_UNCOMMENT_USE_G2D:imxgpu2d ?= "use-g2d=1"
-INI_UNCOMMENT_USE_G2D:mx8qm-nxp-bsp = ""
-INI_UNCOMMENT_USE_G2D:mx8qxp-nxp-bsp = ""
+INI_UNCOMMENT_USE_G2D:imxgpu2d:mx8qm-nxp-bsp = ""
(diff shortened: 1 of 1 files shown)


otavio (approved):

thochstein:
Hi @otavio. This highlights a change in behavior introduced by the new SOC design. In the old design the SOC override was always higher priority:

```
MACHINEOVERRIDES="aarch64:armv8a:imx:use-nxp-bsp:mx8:mx8x:imxdrm:imxdpu:imxgpu:imxgpu2d:imxgpu3d:mx8qxp:imx8qxp-mek"
```

Assuming this is an oversight, we should fix this in `imx-base.inc`, right?

```
MACHINEOVERRIDES_EXTENDER:mx8qxp:use-nxp-bsp = "imx-generic-bsp:imx-nxp-bsp:mx8-generic-bsp:mx8-nxp-bsp:mx8x-generic-bsp:mx8x-nxp-bsp:imxdrm:imxdpu:imxgpu:imxgpu2d:imxgpu3d:mx8qxp-generic-bsp:mx8qxp-nxp-bsp"
```

otavio:
I'll think about it; not sure.

Did you find other cases where it caused changes?

thochstein:
No, this is the first case I am aware of.

The previous design seems to be intuitive, that a feature override is meant for multiple SoCs and is easily overridden by the SoC override. Making the feature override higher priority than the SoC override just seems wrong. I guess you should have a good reason to change it since it does break existing logic.

otavio:
Yes, it does make sense. Can you prepare a PR for this? Doing so we can revert this PR as well.

thochstein:
See #1010.

PR #1009, opened 2022-03-15
RFC: mesa: provide a downgraded recipe

MaxKrummenacher:
@thochstein
Is an upgrade to the i.MX forks of libdrm/wayland-protocols shortly planned so that we don't need such a hack?


otavio (changes requested):
@thochstein can you do the upgrade so we don't need this downgrade?

thochstein:
Working on it. I found we need `wayland-protocols` upgrade as well.

thochstein:
See #1015

otavio:
I merged it as this fixed a build failure. I am closing this now.

PR #1013, opened 2022-03-22
fsl-eula-unpack: Add handling for multilib install

thochstein:
After installing from the archive, adjust the lib folder(s) to handle
the multilib case.

### DIFF
--- a/classes/fsl-eula-unpack.bbclass
+++ b/classes/fsl-eula-unpack.bbclass
@@ -172,3 +172,12 @@ python do_unpack() {
 }
 
 do_unpack[vardepsexclude] += "FSL_EULA_FILE"
+
+do_install:append() {
+    if [ "${base_libdir}" != "/lib" ] && [ -d "${D}/lib" ]; then
+        mv ${D}/lib ${D}${base_libdir}
+    fi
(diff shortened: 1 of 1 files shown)


thochstein:
@panfir, please have a look as well.

classes/fsl-eula-unpack.bbclass
```diff
@@ -172,3 +172,12 @@ python do_unpack() {
...
 
 do_unpack[vardepsexclude] += "FSL_EULA_FILE"
+
+do_install:append() {
+    if [ "${base_libdir}" != "/lib" ] && [ -d "${D}/lib" ]; then
+        mv ${D}/lib ${D}${base_libdir}
+    fi
+    if [ "${libdir}" != "/usr/lib" ] && [ -d "${D}/usr/lib" ]; then
+        mv ${D}/usr/lib ${D}${libdir}
+    fi
```
otavio, line 182: Why this hack here? This seems like something should be fixed in respective recipes, not hidden behind the class.

thochstein: Okay, that's fair. We have a newer version of unpack that also does a generic install of the archive, hence my thinking here, but you're right it doesn't belong here with the older design.

PR #1016, opened 2022-03-25
libdrm: remove unused patch

MaxKrummenacher:
With the update to 2.4.109.imx the patch is no longer needed.
Also remove it from the layer.

Signed-off-by: Max Krummenacher <max.krummenacher@toradex.com>


thochstein:
Oops, thanks!

PR #1030, opened 2022-04-04
weston: Add vsync to cloned displays.

pwojtaszczyk-ts:
When using g2d clone mode allocates additional surfaces according to
FB_MULTI_BUFFER for each cloned display. The g2d blit from the main screen
to the cloned displays is done using inactive surface. The FBIOPAN_DISPLAY
ioctl to the surface is done afterwards.


otavio (approved):
@thochstein, could you ask for an internal review and, if accepted, be applied for the next release?

thochstein:
Will do, thanks!

thochstein:
Okay, the patch is internal now, will be in our next release scheduled end of June.

PR #1066, opened 2022-05-03
linux-imx 5.15.5 plus u-boot-imx

thochstein:


recipes-kernel/linux/linux-imx-README
```diff
@@ -0,0 +1,6 @@
+The linux-imx kernel build uses a defconfig maintained in the layer. That
+defconfig is a copy of a defconfig maintained by NXP in the source.
+
+| defconfig in layer              | copied from NXP kernel source     |
+| linux-imx/imx-nxp-bsp/defconfig | arch/arm/configs/imx_v7_defconfig   |
+| linux-imx/mx8-nxp-bsp/defconfig | arch/arm64/configs/imx_v8_defconfig |
```
otavio, line 6: I like the idea but it ought to be inside the `linux-imx_xx.bb` file and not a new one. It belongs to the recipe.

recipes-bsp/u-boot/u-boot-imx_2021.04.bb
```diff
@@ -38,4 +38,4 @@ do_deploy:append:mx8m-nxp-bsp() {
 }
 
 PACKAGE_ARCH = "${MACHINE_ARCH}"
-COMPATIBLE_MACHINE = "(mx6-generic-bsp|mx7-generic-bsp|mx8-generic-bsp)"
+COMPATIBLE_MACHINE = "(imx-generic-bsp)"
```
otavio, line 41: Not the same as it isn't expected to work with i.MX28 for example.

otavio (changes requested):

recipes-bsp/u-boot/u-boot-imx_2021.04.bb
thochstein: Oh, I thought the `imx` override was only 6, 7, and 8. Was it? Is this a change with the new SOC overrides?

otavio: No; also NXP produces the other i.MX so we ought to be consistent.

otavio (approved):

otavio:
@zandrey can you review this, please?

recipes-kernel/linux/linux-imx-headers_5.15.bb
```diff
@@ -39,17 +39,17 @@ IMX_UAPI_HEADERS = " \
...
     # installed by mistake.
     oe_runmake headers_install INSTALL_HDR_PATH=${B}${exec_prefix}
 
     # Kernel should not be exporting this header
-    rm -f ${D}${exec_prefix}/include/scsi/scsi.h
+    rm -f ${B}${exec_prefix}/include/scsi/scsi.h
 
     # The ..install.cmd conflicts between various configure runs
-    find ${D}${includedir} -name ..install.cmd | xargs rm -f
+    find ${B}${includedir} -name ..install.cmd | xargs rm -f
```
thochstein, line 50: @zandrey, these lines for `scsi.h` and `install.cmd` puzzled me, and I thought to remove them, but I saw that they were in `linux-libc-headers` and thought perhaps your intent was to copy what was there just in case? Seems a maintenance problem though keeping these in sync. Is it worth it? At the least I guess we need a comment?

otavio: I think the closer it is, easier the maintenance. So it makes sense to let it aligned.

PR #1069, opened 2022-05-07
libimxdmabuffer: Upgrade to version 1.1.2

dv1:
* waf: update to 2.0.23
* dma-heap: Add support for uncached dma-heap memory
  imx_dma_buffer_dma_heap_allocator_new_from_fd() is
  a new function that partially deprecates the dma_heap_fd
  argument of imx_dma_buffer_dma_heap_allocator_new().
* Add sync access functions to ensure cache
  coherency when allocating cached DMA memory;
  only done by dma-heap allocator at this time
* Relax buffer mapping flags checks
* Add API functions to retrieve dma-heap / ION FDs

This requires linux-imx 5.15.5 since linux-imx-headers - a libimxdmabuffer dependency - has been updated to 5.15.5 and because this is the linux-imx version that introduces uncached dma-heap memory allocation. I propose to merge this recipe upgrade once linux-fslc-imx also has been upgraded to 5.15.5.


recipes-bsp/libimxdmabuffer/libimxdmabuffer_1.1.2.bb
```diff
@@ -18,13 +18,17 @@ S = "${WORKDIR}/git"
...
-# This must be one that allocates physically contiguous memory, otherwise
-# it cannot allocate usable DMA buffers.
-DMA_HEAP_DEVICE_NODE_PATH = "/dev/dma_heap/linux,cma"
+# Device node path to the DMA-BUF heap to use in the dma-heap allocator. This
+# must be one that allocates physically contiguous memory, otherwise it cannot
+# allocate usable DMA buffers. An uncached DMA heap is selected to avoid issues
+# with cache coherence. Note that the uncached heap is not available prior to
+# imx-kernel 5.15.5; older kernels should use "/dev/dma_heap/linux,cma" instead
+# and remove the "--dma-heap-uncached-memory" switch below from EXTRA_OECONF.
+DMA_HEAP_DEVICE_NODE_PATH = "/dev/dma_heap/linux,cma-uncached"
```
otavio, line 27: The node path ought to be handled at compile time or runtime. If possible to do it at runtime, it'd be compatible to new and old kernels.

If compile time is the only option here tie it to a PACKAGECONFIG so it is always used right.

otavio (changes requested):

dv1:
Updated. dma-heap flags are now in their own variables and are added to the PACKAGECONFIG.

And these are meant as default settings, for cases when a device node can't be explicitly selected at runtime, so I keep them at compile time.

recipes-bsp/libimxdmabuffer/libimxdmabuffer_1.1.2.bb
```diff
@@ -48,12 +53,12 @@ PACKAGECONFIG:append:mx8m-nxp-bsp     = " dwl"
...
-PACKAGECONFIG[ipu]      = "--with-ipu-allocator=yes,               --with-ipu-allocator=no,"
-PACKAGECONFIG[g2d]      = "--with-g2d-allocator=yes,               --with-g2d-allocator=no,virtual/libg2d"
-PACKAGECONFIG[pxp]      = "--with-pxp-allocator=yes,               --with-pxp-allocator=no,"
-PACKAGECONFIG[dma-heap] = "--with-dma-heap-allocator=yes,          --with-dma-heap-allocator=no,"
+PACKAGECONFIG[dwl]      = "--with-dwl-allocator=yes ${HANTRO_CONF},        --with-dwl-allocator=no,imx-vpu-hantro"
+PACKAGECONFIG[ion]      = "--with-ion-allocator=yes,                       --with-ion-allocator=no,"
+PACKAGECONFIG[ipu]      = "--with-ipu-allocator=yes,                       --with-ipu-allocator=no,"
+PACKAGECONFIG[g2d]      = "--with-g2d-allocator=yes,                       --with-g2d-allocator=no,virtual/libg2d"
+PACKAGECONFIG[pxp]      = "--with-pxp-allocator=yes,                       --with-pxp-allocator=no,"
+PACKAGECONFIG[dma-heap] = "--with-dma-heap-allocator=yes ${DMA_HEAP_CONF}, --with-dma-heap-allocator=no,"
```
otavio, line 61: Why you add a single instead one as: `dma-heap-cached` and `dma-heap-uncached`?

otavio (changes requested):

recipes-bsp/libimxdmabuffer/libimxdmabuffer_1.1.2.bb
dv1: Because there is one single dma-heap allocator. The values from `DMA_HEAP_CONF` are compile-time defaults. If users want to explicitly open a certain dma-heap, they specify an open dma-heap FD and whether that dma-heap is cached or not to `imx_dma_buffer_dma_heap_allocator_new_from_fd()`, otherwise they call `imx_dma_buffer_dma_heap_allocator_new()` and let libimxdmabuffer use the configured default.

recipes-bsp/libimxdmabuffer/libimxdmabuffer_1.1.2.bb
```diff
@@ -34,6 +28,17 @@ EXTRA_OECONF = "--imx-linux-headers-path=${STAGING_INCDIR_IMX} \
...
+# must be one that allocates physically contiguous memory, otherwise it cannot
+# allocate usable DMA buffers. An uncached DMA heap is selected to avoid issues
+# with cache coherence. Note that the uncached heap is not available prior to
+# imx-kernel 5.15.5; older kernels should use "/dev/dma_heap/linux,cma" instead
+# and remove the "--dma-heap-uncached-memory" switch below from DMA_HEAP_CONF.
+DMA_HEAP_DEVICE_NODE_PATH = "/dev/dma_heap/linux,cma-uncached"
+
+DMA_HEAP_CONF = "--dma-heap-device-node-path=${DMA_HEAP_DEVICE_NODE_PATH} \
+                 --dma-heap-uncached-memory"
+
```
thochstein, line 41: I think what Otavio means below is to define two package configs `dma-heap-cached` and `dma-heap-uncached`. Then the user can more easily select the cached heap.

dv1: Ah, true, right now it cannot be selected just by setting one value. Although, is `PACKAGECONFIG` really the best option for this? What about a separate `USE_UNCACHED_DMA_HEAP` variable that can be set to 1 or 0?

thochstein: I prefer the `PACKAGECONFIG` approach here because it elevates that choice into the `PACKAGECONFIG` list itself.

> I mean, you wouldn't be able to add both dma-heap-cached and dma-heap-uncached to PACKAGECONFIG.

You can set the sixth field of the `PACKAGECONFIG` variable to capture this mutual exclusion.

dv1: Alright, I will use `PACKAGECONFIG` then.

dv1:
Updated. This adds the `dma-heap-cached` and `dma-heap-uncached` flags now. Also, this adds a backported patch to fix a build problem when the G2D allocator is used.

otavio (approved):

Issue #1084, opened 2022-05-12
Build failure for `imx-atf` on master and kirkstone


otavio:
Unfortunately when I went to build `core-image-weston` I ran into a build failure for `imx-atf`:

```
| plat/imx/imx8m/hab.c: In function 'imx_hab_handler':
| plat/imx/imx8m/hab.c:64:57: error: array subscript 0 is outside array bounds of 'uint32_t[0]' {aka 'unsigned int[]'} [-Werror=array-bounds]
|    64 | #define HAB_RVT_CHECK_TARGET_ARM64      ((unsigned long)*(uint32_t *)(HAB_RVT_BASE + 0x18))
|       |                                                         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

_Originally posted by @thochstein in https://github.com/Freescale/meta-freescale/issues/1080#issuecomment-1124471682_

otavio:
@thochstein, I opened this issue so that it isn't forgotten.  It might work arounded using `-Warray-bounds`.

MaxKrummenacher:
I assume this discussion (and eventually the patch) will fix it from within GCC:

https://gcc.gnu.org/bugzilla/show_bug.cgi?id=99578#c39

thochstein:
Thanks, @MaxKrummenacher !

MaxKrummenacher:
@thochstein 
I may have been to optimistic. The workaround described in the gcc bugzilla entry makes the error go away if the constant value casted to pointer and then dereferenced is > 4kB, i.e. > 0x1000.
Looks like the HAB_RVT_BASE is at addresses 0x0880 or 0x0900 so those false positves may still be reported.

PR #1089, opened 2022-05-16
imx-atf: Suppress array-bounds error

thochstein:
The array-bounds error is triggered now in cases where it was silent before,
causing errors like:

```
plat/imx/imx8m/hab.c: In function 'imx_hab_handler':
plat/imx/imx8m/hab.c:64:57: error: array subscript 0 is outside array bounds of 'uint32_t[0]' {aka 'unsigned int[]'} [-Werror=array-bounds]
   64 | #define HAB_RVT_CHECK_TARGET_ARM64      ((unsigned long)*(uint32_t *)(HAB_RVT_BASE + 0x18))
      |                                                         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

It appears that these new cases will be silenced soon in gcc [1], so for
now just silence the warning locally as workaround.

[1] https://gcc.gnu.org/bugzilla/show_bug.cgi?id=99578#c39

Fixes: #1084

### DIFF
--- a/recipes-bsp/imx-atf/imx-atf/0001-Makefile-Suppress-array-bounds-error.patch
+++ b/recipes-bsp/imx-atf/imx-atf/0001-Makefile-Suppress-array-bounds-error.patch
@@ -1,0 +1,42 @@
+From 85f576054a4d5496706bbd37a520942f51cb74b9 Mon Sep 17 00:00:00 2001
+From: Tom Hochstein <tom.hochstein@nxp.com>
+Date: Mon, 16 May 2022 13:45:16 -0500
+Subject: [PATCH] Makefile: Suppress array-bounds error
+
+The array-bounds error is triggered now in cases where it was silent before,
+causing errors like:
+
+```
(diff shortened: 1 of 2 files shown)


tprrt:
Hi,

This change introduces a regression:
`ERROR: imx-atf-2.4+gitAUTOINC+05f788b9bb-r0 do_fetch: Bitbake Fetcher Error: FetchError('Unable to fetch URL from any source.', 'file://0001-alsa-state-Remove-PCM-devices-mapped-by-5.15-kernel.patch')
`

The SRC_URI isn't append with the right file patch, `file://0001-alsa-state-Remove-PCM-devices-mapped-by-5.15-kernel.patch` instead `file://0001-Makefile-Suppress-array-bounds-error.patch`.

In addition, I think it should be necessary to add the following line in the recipe:
`FILESEXTRAPATHS:prepend := "${THISDIR}/${PN}:"`

thochstein:
Sorry for the mistake. Note that `FILESEXTRAPATHS` is not needed here.

PR #1091, opened 2022-05-17
kernel-module-isp-vvcam: fix error found by gcc 12

MaxKrummenacher:
@thochstein 
Another recipe which doesn't build on current master. (i.MX 8M Plus machine specific recipe)
I believe the warning treated as error is legitimate.

Can you check internally what's the way forward?
Less intrusive to my patch is to simple delete the whole 'if' clause which will never ever be executed.


thochstein:
@MaxKrummenacher, your fix is used as-is. Thanks!

PR #1095, opened 2022-05-26
kernel-module-nxp89xx: fix compatible machine override

ricardosalveti:
Module should be compatible with imx-nxp-bsp, as it is specific to the
nxp-bsp.


ricardosalveti:
Please backport to kirkstone if merged in master.

PR #1096, opened 2022-05-31
Remove lzop dependency dropped upstream

zandrey:
Upstream commit [`dea5e88637 ("lzop: remove recipe from oe-core")`](https://github.com/openembedded/openembedded-core/commit/dea5e8863792dc7bb3324b543e04da4c94a060aa) dropped support of _lzop_ and removed recipe from OE-Core.

Drop _lzop_ dependencies in recipes across the layer, it is not available anymore.

Cc: @Ossanes

### DIFF
--- a/recipes-bsp/u-boot/u-boot-fslc_2022.04.bb
+++ b/recipes-bsp/u-boot/u-boot-fslc_2022.04.bb
@@ -8,7 +8,7 @@ version, or because it is not applicable for upstreaming."
 
 inherit ${@oe.utils.ifelse(d.getVar('UBOOT_PROVIDES_BOOT_CONTAINER') == '1', 'imx-boot-container', '')}
 
-DEPENDS += "bc-native dtc-native python3-setuptools-native lzop-native"
+DEPENDS += "bc-native dtc-native python3-setuptools-native"
 
 # Location known to imx-boot component, where U-Boot artifacts
 # should be additionally deployed.
--- a/recipes-kernel/linux/linux-fslc.inc
+++ b/recipes-kernel/linux/linux-fslc.inc
@@ -3,7 +3,7 @@
 
 require recipes-kernel/linux/linux-imx.inc
 
-DEPENDS += "lzop-native bc-native"
+DEPENDS += "bc-native"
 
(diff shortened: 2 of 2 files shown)


otavio (approved):

S4mw1s3:
I'm currently getting the error "/bin/sh: lzop: command not found" when compiling linux-fslc v5.18. Although I'm not using the linux-fslc defconfig from meta-freescale, I see it currently also still contains `CONFIG_KERNEL_LZO=y`. Shouldn't that be changed after this merge request to avoid the error "/bin/sh: lzop: command not found"?

zandrey:
Hello @S4mw1s3 

Thanks for reporting this!

You're right, kernel config still contains the `LZO` compression...

I've followed-up a bit the discussion for that removal, and it look like it has been migrated from _OE-Core_ to _meta-openembedded_ with commit [`90ff53b8df ("lzop: add (from oe-core)")`](https://github.com/openembedded/meta-openembedded/commit/90ff53b8df1e3259cbc201c658a4f3f4dddf3aa8)

I see 2 possible solutions here:
1. Remove the `CONFIG_KERNEL_LZO` config option from Kernel config. `LZO` is quite old and unmaintained (last update dated 2017), and if there are not a lot or users here that utilize this compression option - then it does not make sense to keep it in the `linux-fslc` fork.
2. Revert this PR and introduce layer dependency to _meta-openembedded_ since it would be required to build Kernel with `LZO` enabled.

TBH, I would really appreciate option 1 from above, since this would not require this layer to have any dependencies as it is now...

@otavio Since this is a conceptual question, I would like to defer it to you to decide how to proceed with this issue and which solution to implement from those above. Perhaps there is a third one exists, which I overlooked.

Thanks a lot!

-- andrey

otavio:
I prefer to avoid the dependency on extra layer. Can you remove it from defconfig?

zandrey:
Hello @otavio 

> I prefer to avoid the dependency on extra layer. Can you remove it from defconfig?

Yes, sure! I'm working on the Kernel update now anyway, so I can drop the `CONFIG_KERNEL_LZO` as well (both from Kernel tree and layer).

-- andrey

otavio:
Please commit it in the linux kernel repo as well so we keep it in there for future rebase.

PR #1109, opened 2022-06-14
Support upstream TF-A usage for i.MX8M machines

zandrey:
This PR adds an extension to be able to build and use [_TF-A  from upstream_](https://git.trustedfirmware.org/TF-A/trusted-firmware-a.git/).

In order to use the upstream _TF-A_ in the BSP, one need to pull `meta-arm` layer into the setup to get the `trusted-firmware-a` recipe and package.

Additional variable `DEFAULT_ATF_PROVIDER` shall be set to point to `trusted-firmware-a` recipe name, so that it would be picked up instead of NXP one. As a continuation of this work, TF-A provider can be made distro-specific, for the moment it is left open in this PR.

Above provider choice (when upstream TF-A is set) requires `COMPATIBLE_MACHINE` definition, which is added via _dynamic-layers_ mechanism.

Verified booting for following machines (upstream TF-A version `v2.7`):
- _imx8mm-lpddr4-evk_: **PASS**
- _imx8mn-ddr4-evk_: **PASS**
- _imx8mp-lpddr4-evk_: **PASS**

**Cc:** @thochstein @MaxKrummenacher @gibsson @grembeter @htho-lgs @Ossanes 

-- andrey

### DIFF
--- a/classes/imx-boot-container.bbclass
+++ b/classes/imx-boot-container.bbclass
@@ -47,7 +47,7 @@ do_resolve_and_populate_binaries() {
                         cp ${DEPLOY_DIR_IMAGE}/${ddr_firmware} ${B}/${config}/
                     done
                     if [ -n "${ATF_MACHINE_NAME}" ]; then
-                        cp ${DEPLOY_DIR_IMAGE}/${BOOT_TOOLS}/${ATF_MACHINE_NAME} ${B}/${config}/bl31.bin
+                        cp ${DEPLOY_DIR_IMAGE}/${ATF_MACHINE_NAME} ${B}/${config}/bl31.bin
                     else
                         bberror "ATF binary is undefined, result binary would be unusable!"
                     fi
--- a/classes/imx-boot-container.bbclass
+++ b/classes/imx-boot-container.bbclass
@@ -29,7 +29,7 @@ ATF_MACHINE_NAME:append = "${@bb.utils.contains('MACHINE_FEATURES', 'optee', '-o
 # appropriate dependencies for populate binaries task
 do_resolve_and_populate_binaries[depends] += " \
     ${@' '.join('%s:do_deploy' % r for r in '${IMX_EXTRA_FIRMWARE}'.split() )} \
-    imx-atf:do_deploy \
+    ${IMX_DEFAULT_ATF_PROVIDER}:do_deploy \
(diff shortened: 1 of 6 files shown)


classes/imx-boot-container.bbclass
```diff
@@ -29,7 +29,7 @@ ATF_MACHINE_NAME:append = "${@bb.utils.contains('MACHINE_FEATURES', 'optee', '-o
 # appropriate dependencies for populate binaries task
 do_resolve_and_populate_binaries[depends] += " \
     ${@' '.join('%s:do_deploy' % r for r in '${IMX_EXTRA_FIRMWARE}'.split() )} \
-    imx-atf:do_deploy \
+    ${DEFAULT_ATF_PROVIDER}:do_deploy \
```
otavio, line 32: I'd prefer if this variable had the IMX_ prefix. It shows to user it is something i.MX specific.

otavio (changes requested):

classes/imx-boot-container.bbclass
zandrey: Make sense, would change this in _V2_.

zandrey:
Hello @fabioestevam 

Thanks for the review and your comments here!

> Hi @zandrey
> 
> The series looks good. I would suggest including a short notice explaining that mainline TF-A currently does not support secure boot. This way the users become aware of this limitation and can select NXP TF-A if they need HAB support.

Good point, I will add a note about the fact that at the time of writing - upstream TF-A does not have HAB support included, but this can change with the time.

@otavio Should this PR and implementation be accepted, is it possible that this fact be reflected in _Release Notes_ as well? Guess that would be beneficial to included it there.

> 
> There was a thread about this topic in the TF-A list:
> 
> https://lists.trustedfirmware.org/archives/list/tf-a@lists.trustedfirmware.org/message/5UYYI4CNLQ5OWWHU466JBK6I5QIQ2VRS/

Thanks, I'll put this link into commit message as well.

> 
> Thanks

-- andrey

zandrey:
> LGTM

Thanks @thochstein !

zandrey:
Changes in _V2_:
- Renamed variable `DEFAULT_ATF_PROVIDER` to `IMX_DEFAULT_ATF_PROVIDER` as suggested by @otavio 
- Added a comment and expanded commit message declaring absence of HAB feature in upstream TF-A as suggested by @fabioestevam

otavio (approved):

PR #1119, opened 2022-06-23
isp-imx: fix source and build directories

lucaceresoli:
The source and build directories are usually ${WORKDIR}/build and
${WORKDIR}/${BP} but not always. For example when using devtool they are
different, leading to install failures such as:

  cp: cannot stat '.../tmp/work/.../isp-imx/4.2.2.16.0-r0/build/generated/release/bin/*_test': No such file or directory

Fix by using the S and B variables which always point to the correct
directory.

Signed-off-by: Luca Ceresoli <luca.ceresoli@bootlin.com>

### DIFF
--- a/recipes-bsp/isp-imx/isp-imx_4.2.2.16.0.bb
+++ b/recipes-bsp/isp-imx/isp-imx_4.2.2.16.0.bb
@@ -58,22 +58,22 @@ do_install() {
     install -d ${D}/${includedir}
     install -d ${D}/opt/imx8-isp/bin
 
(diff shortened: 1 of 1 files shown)


otavio:
Pleare prepare a PR but it is EOL so we won't keep doing the backports for EOL releases.

lucaceresoli:
> Pleare prepare a PR but it is EOL so we won't keep doing the backports for EOL releases.

Done: #1121.

It would be great if honister could receive at least simple fixes until Boundary Devices releases a kirkstone version!

PR #1124, opened 2022-06-27
firmware-imx: create symbolic links for CODA VPU firmwares

liuming50:
Create symbolic links in ${nonarch_base_libdir}/firmware for CODA VPU
firmwares, otherwise CODA driver could not load them directly in probe
function and hence fallback firmware loading being triggered, which is
usually 40-60 seconds later after system boots up.

### DIFF
--- a/recipes-bsp/firmware-imx/firmware-imx_8.15.bb
+++ b/recipes-bsp/firmware-imx/firmware-imx_8.15.bb
@@ -65,6 +65,29 @@ do_install() {
 # which contains only files that matches the pattern.
 #
 python populate_packages:prepend() {
+    # CODA driver tries to locate VPU firmwares directly in ${nonarch_base_libdir}/firmware, to
+    # avoid fallback loading which is usually 40-60 seconds later after system boots up, let's
+    # create symbolic links in ${nonarch_base_libdir}/firmware for VPU firmwares.
+    def coda_vpu_links(file, pkg, pattern, format, basename):
+        # Only CODA VPU firmwares need this procedure
+        if 'imx8' in basename:
+            return
+
+        dir = os.path.dirname(file)
+        dir = os.path.abspath(os.path.join(dir, os.pardir))
+        cwd = os.getcwd()
+
+        os.chdir(dir)
+
+        name = os.path.basename(file)
+        os.symlink(os.path.join("vpu", name), name)
+
+        oldfiles = d.getVar('FILES:' + pkg)
+        newfile = os.path.join(d.getVar('nonarch_base_libdir'), "firmware", name)
+        d.setVar('FILES:' + pkg, oldfiles + " " + newfile)
+
+        os.chdir(cwd)
+
     easrcdir = bb.data.expand('${nonarch_base_libdir}/firmware/imx/easrc', d)
     do_split_packages(d, easrcdir, '^easrc-([^_]*).*\.bin',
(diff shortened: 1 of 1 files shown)


recipes-bsp/firmware-imx/firmware-imx_8.15.bb
```diff
@@ -53,6 +53,11 @@ do_install() {
     # as some of other files are provided by packages from other recipes.
     install -d ${D}${nonarch_base_libdir}/firmware/vpu
     install -m 0644 ${S}/firmware/vpu/vpu_fw_imx*.bin ${D}${nonarch_base_libdir}/firmware/vpu
+    ln -srf ${D}${nonarch_base_libdir}/firmware/vpu/vpu_fw_imx27_TO2.bin ${D}${nonarch_base_libdir}/firmware/vpu_fw_imx27_TO2.bin
```
otavio, line 56: It would be better to do this as a `for` loop.

recipes-bsp/firmware-imx/firmware-imx_8.15.bb
```diff
@@ -128,5 +133,10 @@ FILES:${PN}-hdmi = " \
     ${nonarch_base_libdir}/firmware/hdmirxfw.bin \
     ${nonarch_base_libdir}/firmware/dpfw.bin \
 "
+FILES:${PN}-vpu-imx27 += "${nonarch_base_libdir}/firmware/vpu_fw_imx27_TO2.bin"
```
otavio, line 136: The package splitting could be reworked to do this automatically.

otavio (changes requested):

recipes-bsp/firmware-imx/firmware-imx_8.15.bb
```diff
@@ -65,6 +65,26 @@ do_install() {
...
+        dir = os.path.abspath(os.path.join(dir, os.pardir))
+        cwd = os.getcwd()
+
+        os.chdir(dir)
+        os.symlink(os.path.join("vpu", name), name)
+        oldfiles = d.getVar('FILES:' + pkg)
+        newfile = os.path.join(d.getVar('nonarch_base_libdir'), "firmware", name)
+        d.setVar('FILES:' + pkg, oldfiles + " " + newfile)
+        os.chdir(cwd)
+
```
otavio, line 87: I propose following change so it is easier to read:

```suggestion
    # CODA driver tries to locate VPU firmwares directly in ${nonarch_base_libdir}/firmware, to
    # avoid fallback loading which is usually 40-60 seconds later after system boots up, let's
    # create symbolic links in ${nonarch_base_libdir}/firmware for VPU firmwares.
    def coda_vpu_links(file, pkg, pattern, format, basename):
        # Only CODA VPU firmwares need this procedure
        if 'imx8' in basename:
            return

        dir = os.path.dirname(file)
        dir = os.path.abspath(os.path.join(dir, os.pardir))
        cwd = os.getcwd()
        
        os.chdir(dir)
        
        name = os.path.basename(file)
        os.symlink(os.path.join("vpu", name), name)
        
        oldfiles = d.getVar('FILES:' + pkg)
        newfile = os.path.join(d.getVar('nonarch_base_libdir'), "firmware", name)
        d.setVar('FILES:' + pkg, oldfiles + " " + newfile)
        
        os.chdir(cwd)

```

otavio (changes requested):

PR #1133, opened 2022-07-12
weston: 9.0.0.imx: fix build with latest master

MaxKrummenacher:
This seems to be master only, i.e. on Kirkstone the wayland update is not present.


otavio:
Can you check if it hurts to include in kirkstone? I've keeping them in sync.

MaxKrummenacher:
Will do.
Unfortunately this has to wait until tomorrow.

MaxKrummenacher:
@otavio 
I built weston_9.0.0-imx kirkstone without and with the patch applied. Both builts succeed.
I think we are save to backport the fix.

PR #1134, opened 2022-07-12
remove random file

MaxKrummenacher:
@zandrey 
I guess my assumption is correct that the file sipped in unintended.


zandrey:
@MaxKrummenacher 
True, good catch! This was the Kernel upgrade log, which I generate separately and use in commit message. Somehow it managed to slipped into the commit and I obviously overlooked it... 😞 

Thanks a lot!

-- andrey

zandrey (approved):

PR #1141, opened 2022-07-23
Secure boot rework

otavio:


recipes-bsp/imx-atf/imx-atf_2.6.bb
```diff
@@ -9,10 +9,10 @@ FILESEXTRAPATHS:prepend := "${THISDIR}/${PN}:"
 
 PV .= "+git${SRCPV}"
 
-SRCBRANCH = "lf_v2.4"
+SRCBRANCH = "lf_v2.6"
 SRC_URI = "git://source.codeaurora.org/external/imx/imx-atf.git;protocol=https;branch=${SRCBRANCH} \
            file://0001-Makefile-Suppress-array-bounds-error.patch"
-SRCREV = "05f788b9bbb13d002997e35008a4b945f7e2957b"
+SRCREV = "c6a19b1a351308cc73443283f6aa56b2eff791b8"
```
otavio, line 15: @thochstein, please send the `0001-Makefile-Suppress-array-bounds-error.patch` for internal review. Ideally, those warnings ought to be fixed but merging this also helps.

thochstein: As the commit message says, the patch was only meant to be temporary. It looks like we never needed it. Please try just dropping it.

otavio: I did and it fails build using new GCC, so yes, it is required.

thochstein: Oh, strange, I wonder why we do not have the problem.

otavio: It fails if using master. Please check.

thochstein: The warnings are false-positives. The fix commit is shown here:

https://gcc.gnu.org/bugzilla/show_bug.cgi?id=99578#c39

As the commit says, the fix for GCC 12 is not a complete fix, and code that uses a constant less than 4kB as an address --which is unfortunately the case here -- will still trigger the warning. For GCC 13 it appears that the false-positives are completely eliminated by looking more specifically for a NULL value.

So I don't think there's anything to do but to use this workaround for now and wait for the fix in GCC 13.

otavio: Would you mind sending a PR adding this information to the patch file?

conf/machine/include/imx-base.inc
```diff
@@ -503,22 +503,18 @@ WKS_FILE_DEPENDS ?= " \
...
 # 'imx-boot-container' override, and test if the U-Boot built 'flash.bin'
 # binary is used a replacement.
-# Note, that the results binary name of the boot container is set to 'imx-boot'
+#
+# NOTE: the results binary name of the boot container is set to 'imx-boot'
 # for both NXP and Mainline BSP.
-# For Mainline BSP: the 'flash.bin' boot container is renamed during the
-# deployment task extesion execution defined in imx-boot-container class.
-# For NXP BSP: rename is done in 'imx-boot' recipe at the execution of compile
-# task.
```
angolini, line 520: I'm unsure of removing this comment. It seems to me that it might still be helpfull

angolini (approved):

conf/machine/include/imx-base.inc
otavio: We could have it inside the class as it doesn't belongs to `imx-base.inc` for sure.

angolini: ok

recipes-bsp/imx-atf/imx-atf_2.6.bb
otavio: I reviewed the comment and it looks fine. @kraj have this issue been fixed in GCC13? Could this be backported?

classes/imx-boot-container.bbclass
```diff
@@ -20,6 +20,9 @@
 # by variable UBOOT_PROVIDES_BOOT_CONTAINER, which is defined in the
 # base machine include file (imx-base.inc), and is set to "1" when the
 # 'imx-boot-container' is present in MACHINEOVERRIDES.
+#
+# NOTE: A backwards-compatible symlink is added for 'flash.bin', named
+# 'imx-boot', during the deployment task.
```
otavio, line 25: @angolini please check.

recipes-bsp/imx-atf/imx-atf_2.6.bb
kraj: yes it is ok in gcc13 and also in clang14

otavio: Could it be backported to kirkstone? or not?

thochstein: There's one thing still bothering me though. The bug mentions fixing GCC 11 as well. We are building imx-atf from our meta-imx layer with GCC 11 without any problem with array-bounds. Does it make sense?

thochstein: > Could it be backported to kirkstone? or not?

You'd think if they could have done the complete solution in GCC 12 they would have?

kraj: its not an easy backport since it need infra which is only added in gcc13

kraj: > 

its fixed/backported partially into gcc 11.3 and if you are using latest kirkstone then we are on 11.3

thochstein: So in the case of imx-atf, the false positives we see are not fixed by the partial fix since the constants are less than 4kB. So what's strange is we do not see the array-bounds error.

thochstein: > Would you mind sending a PR adding this information to the patch file?

See #1142.

classes/imx-boot-container.bbclass
angolini: Nice @otavio

recipes-bsp/imx-atf/imx-atf_2.4.bb
```diff
@@ -48,6 +49,9 @@ EXTRA_OEMAKE += 'LD="${@remove_options_tail(d.getVar('LD'))}"'
 
 EXTRA_OEMAKE += 'CC="${@remove_options_tail(d.getVar('CC'))}"'
 
+# Set the UART to use during the boot.
+EXTRA_OEMAKE += 'IMX_BOOT_UART_BASE=${ATF_BOOT_UART_BASE}'
```
thochstein, line 53: If someone sees `IMX_BOOT_UART_BASE=INVALID` in the build, they might think something is wrong and waste time figuring out that it is not wrong. Can you change this so nothing gets added to `EXTRA_OEMAKE` in this case?

otavio: I don't want to let this be "skipped" behind the scenes.

We have values for the supported SoCs inside `imx-base.inc`, and if a new SoC is added, it will fail to build as a platform does.

thochstein: Currently `IMX_BOOT_UART_BASE=INVALID` is passed for everything but `mx8m-generic-bsp`. Are you saying that you are not finished and you plan to define the rest?

otavio: No. Those are the only ones I could find. See:
```
otavio@yocto-env ..x-atf/2.6+gitAUTOINC+c6a19b1a35-r0/git (git)-[lf_v2.6] % git grep '^IMX_BOOT_UART_BASE'
plat/imx/imx8m/imx8mm/platform.mk:IMX_BOOT_UART_BASE    ?=    0x30890000
plat/imx/imx8m/imx8mn/platform.mk:IMX_BOOT_UART_BASE    ?=    0x30890000
plat/imx/imx8m/imx8mp/platform.mk:IMX_BOOT_UART_BASE    ?=    0x30890000
```
I found a wrong operator which caused an error. Could you check it?

thochstein: Okay, I think it makes sense to me now.

conf/machine/include/imx-base.inc
```diff
@@ -113,6 +113,10 @@ UBOOT_PROVIDES_BOOT_CONTAINER:imx-boot-container = "1"
 # Default TF-A provider to NXP downstream fork
 IMX_DEFAULT_ATF_PROVIDER ??= "imx-atf"
 
+# Allow setting the UART used during the boot by ATF.
+SOC_ATF_BOOT_UART_BASE:mx8m-generic-bsp   = "0x30890000"
```
thochstein, line 117: Do you not need to set the default?

```
SOC_ATF_BOOT_UART_BASE = ""
```

otavio: I do. Thanks. Fixed.

thochstein (approved):

PR #1168, opened 2022-08-09
kernel modules: Explicitly depends on virtual/kernel

angolini:
Although it is not needed for build or runtime dependency, the
dependency is needed to be explicit to link the spdx chain correctly and
avoid broken task dependencies.

Signed-off-by: Daiane Angolini <daiane.angolini@foundries.io>

### DIFF
--- a/recipes-extended/dpdk/dpdk-module_20.11.bb
+++ b/recipes-extended/dpdk/dpdk-module_20.11.bb
@@ -9,7 +9,7 @@ SRC_URI += " \
 inherit module
 
 #kernel module needs 'rte_build_config.h', which is generated at buid time
-DEPENDS += "dpdk"
+DEPENDS += "dpdk virtual/kernel"
 
 export S
 export STAGING_KERNEL_DIR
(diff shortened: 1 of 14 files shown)


otavio (approved):

thochstein:
Hmm, is it not the case that `module.bbclass` should provide this dependency?

otavio:
That is a valid question. @rpurdie?

rpurdie:
Maybe. It likely won't be straight forward due to circular dependency issues.

otavio:
@angolini are you looking at `module.bbclass` rework?

angolini:
I am not @otavio Who is on charge of this is @quaresmajose

otavio:
Thanks @angolini; @quaresmajose, please let me know if I can help.

quaresmajose:
@otavio I am working on it. What I am validating/checking is if kernel modules are rebuild when the kernel change their task hash signatures.

I find a way that shows the requirement of the `DEPENDS += "virtual/kernel"` in the `module.bbclass` but I need now to replicate the steps using only the OE-core and I am thinking in open a bug in Yocto to discuss this issue.

@rpurdie during my tests I don't see any circular dependency, can you please give any tips on how I can get into a case like this?

quaresmajose:
I have created the bug [1] with the steps to reproduce the issue and will send the patch for `module.bbclass` asap
[1] - https://bugzilla.yoctoproject.org/show_bug.cgi?id=14885

otavio:
As this is being worked on OE-Core, I will close this.

PR #1169, opened 2022-08-10
Rework fsl-eula-unpack.bbclass for reuse

thochstein:
We have an internal use for the license information calculated by class.

### DIFF
--- a/classes/fsl-eula-unpack.bbclass
+++ b/classes/fsl-eula-unpack.bbclass
@@ -104,6 +104,28 @@ do_fetch:prepend() {
         bb.fatal("The recipe LICENSE should include Proprietary but is " + d.getVar("LICENSE") + ".")
 }
 
+do_unpack[vardepsexclude] += "FSL_EULA_FILE"
+python do_unpack() {
+    eula = d.getVar('ACCEPT_FSL_EULA')
+    eula_file = d.getVar('FSL_EULA_FILE')
+    pkg = d.getVar('PN')
+    if eula == None:
+        bb.fatal("To use '%s' you need to accept the Freescale EULA at '%s'. "
+                 "Please read it and in case you accept it, write: "
+                 "ACCEPT_FSL_EULA = \"1\" in your local.conf." % (pkg, eula_file))
+    elif eula == '0':
+        bb.fatal("To use '%s' you need to accept the Freescale EULA." % pkg)
+    else:
+        bb.note("Freescale EULA has been accepted for '%s'" % pkg)
+
+    try:
+        bb.build.exec_func('base_do_unpack', d)
+    except:
+        raise
+
+    bb.build.exec_func('fsl_bin_do_unpack', d)
+}
+
 python fsl_bin_do_unpack() {
     src_uri = (d.getVar('SRC_URI') or "").split()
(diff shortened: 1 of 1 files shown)


classes/fsl-eula-unpack.bbclass
```diff
@@ -128,6 +128,8 @@ python fsl_bin_do_unpack() {
         bb.note("Handling file '%s' as a Freescale EULA-licensed archive." % url.basename)
         cmd = "sh %s --auto-accept --force" % (url.localpath)
         bb.fetch2.runfetchcmd(cmd, d, quiet=True, workdir=rootdir)
+    if not found:
+        bb.warn("The package includes 'fsl-eula-unpack.bbclass' but does not mark any URL in SRC_URI with the option 'fsl-eula=true'")
```
otavio, line 154: `bb.fatal` might enforce it.

classes/fsl-eula-unpack.bbclass
```diff
@@ -157,11 +157,21 @@ python fsl_bin_do_unpack() {
...
+    (layer_license, licenses, md5sums, found_layer_licenses, found_package_licenses) = findNxpEulaLicenses(d)
+    if not found_layer_licenses:
+        bb.fatal("The Freescale layer EULA '%s' is not listed in LIC_FILES_CHKSUM '%s'."
+                 % (layer_license, licenses))
+    if not found_package_licenses:
+        bb.fatal("A valid package EULA with md5sum in %s was not found in LIC_FILES_CHKSUM '%s'."
+                 % (md5sums.split(), licenses))
+}
+
+def findNxpEulaLicenses(d):
```
otavio, line 169: Python uses `find_nxp_eula_licenses(d):` so the snake_case.

otavio (changes requested):

classes/fsl-eula-unpack.bbclass
thochstein: Haha, old habit.

classes/fsl-eula-unpack.bbclass
```diff
@@ -171,14 +181,8 @@ python fsl_bin_do_unpack() {
...
-            found_package_license = True
-    if not found_layer_license:
-        bb.fatal("The Freescale layer EULA '%s' is not listed in LIC_FILES_CHKSUM '%s'."
-                 % (layer_license, licenses))
-    if not found_package_license:
-        bb.fatal("A valid package EULA with md5sum in %s was not found in LIC_FILES_CHKSUM '%s'."
-                 % (md5sums.split(), licenses))
-}
+            found_package_licenses += license
+    return (layer_license, licenses, md5sums, found_layer_licenses, found_package_licenses)
```
otavio, line 188: only the first two arguments are used. It could reduce the number of tuple items.

otavio (changes requested):

classes/fsl-eula-unpack.bbclass
thochstein: Not sure what's best. In my reuse case, I want the other values as well, so it does reduce a few lines of duplication.

otavio: That's fine then.

PR #1173, opened 2022-08-18
linux-imx: change LOCALVERSION to not use underscore

pberginkonsult:
Using underscore '_' in LOCALVERSION makes build failures if you use
deb packages.

Error message:
  Subprocess output:dpkg-deb: error: package name has characters that aren't lowercase alphanums or '-+.'

Change '_' to '-' to avoid this problem. It seems that linux-fslc-imx already have this convention
and previous versions of linux-imx also had.

Signed-off-by: Peter Bergin <peter@berginkonsult.se>


pberginkonsult:
If accepted I would request this change to be backported to kirkstone.

otavio (approved):

PR #1179, opened 2022-08-24
imx-atf: fix build error with binutils 2.39

### LINKED ISSUE #1178: imx-atf fails to build with binutils 2.39
In current master of poky (081a3655fbb10b83793f102b975e153e75240682) where binutils is updated to 2.39 the package imx-atf fails to build for the imx8mp-lpddr4-evk machine (and most probably others to).

The failure is:
```
|   LD      <...>/imx-atf/2.6+gitAUTOINC+c6a19b1a35-r0/git/build/imx8mp/release/bl31/bl31.elf
| aarch64-poky-linux-ld: warning: <...>/imx-atf/2.6+gitAUTOINC+c6a19b1a35-r0/git/build/imx8mp/release/bl31/bl31.elf has a LOAD segment with RWX permissions
| make: *** [Makefile:1254: <...>/imx-atf/2.6+gitAUTOINC+c6a19b1a35-r0/git/build/imx8mp/release/bl31/bl31.elf] Error 1
```
A similar case in systemd with solution: https://github.com/systemd/systemd-stable/commit/8a6f966be404897b5333c218701965ac3b5a0806
Bug report for atf: https://developer.trustedfirmware.org/T996

Looking at the references above my guess is that a solution would be to add '-Wl,--no-warn-rwx-segments' but haven't been able to find where in the makefile structure to do it.

MaxKrummenacher:
Fixes #1178

I tried to build imx-atf with this patch in kirkstone and it builds fine. However, I assume that binutils will not be backported to kirkstone in openbedded-core and this patch will not be needed in kirkstone.

### DIFF
--- a/recipes-bsp/imx-atf/imx-atf_2.6.bb
+++ b/recipes-bsp/imx-atf/imx-atf_2.6.bb
@@ -29,7 +29,9 @@ EXTRA_OEMAKE += " \
 
 # Let the Makefile handle setting up the CFLAGS and LDFLAGS as it is a standalone application
 CFLAGS[unexport] = "1"
-LDFLAGS[unexport] = "1"
+# Needed for binutils >= 2.39 to prevent build failure.
+# imx-atf links with ld, thus no '-Wl,' prefix
+LDFLAGS = "--no-warn-rwx-segments"
 AS[unexport] = "1"
 LD[unexport] = "1"
 
@@ -56,7 +58,6 @@ EXTRA_OEMAKE += 'IMX_BOOT_UART_BASE=${ATF_BOOT_UART_BASE}'
 do_configure[noexec] = "1"
 
 do_compile() {
-    # Clear LDFLAGS to avoid the option -Wl recognize issue
     oe_runmake bl31
     if ${BUILD_OPTEE}; then
         oe_runmake clean BUILD_BASE=build-optee


otavio:
I'd prefer to keep it in sync for now.

thochstein:
I tried this with our copied recipe on `kirkstone`.

```
| aarch64-poky-linux-ld: unrecognized option '--no-warn-rwx-segments'
```

otavio:
are you using a different toolchain?

thochstein:
Standard toolchain, as far as I know.

MaxKrummenacher:
I will have to check again on kirkstone. I thought I really did a `-c cleansstate` and then build from scratch.

otavio:
I did the revert . See dd9852630093745.

MaxKrummenacher:
Thanks, Sorry for my bad testing.

zandrey:
Hello @MaxKrummenacher,

Please have a look at commit [`977c5222d8c2 ("arm/trusted-firmware-a: work around RWX permission error on segment")`](https://github.com/jonmason/meta-arm/commit/977c5222d8c2f6f6c7e522e480a6e4cd40ba32d7) in _meta-arm_, it is solving the same issue but in a bit the different way.

I guess you can pick that one up and have a quick test round to see if that solves the build issue with `imx-atf`. I've picked up another patch from _meta-arm_ to address the `optee-os` issue with new _binutils_, and submitted it in #1184 

**Cc**: @thochstein

otavio:
This seems as a better patch indeed as it allows it to be merged in kirkstone as well.

MaxKrummenacher:
I think the meta-arm approach has two drawbacks.
- It completely drops 'all warnings are errors' which may hide another warning which may point to a weakness in the code.
- It is a patch and as such might not apply in the future
- Creates work (for me ;-) ).

I agree with Otavio that it is a plus that with the patch the fix is binutils version agnostic.

IMHO the negative aspects outweigh the positive ones, however if the community has a different view I can prepare a PR which reverts the existing fix and adds one which applies the patch from meta-arm.

Opinions?

zandrey:
Hello @MaxKrummenacher 

> I think the meta-arm approach has two drawbacks.
> 
>     * It completely drops 'all warnings are errors' which may hide another warning which may point to a weakness in the code.

This is indeed a valid point, which was not brought up in `meta-arm` ML during the patch review, see [1]. Cc: @jomason here, perhaps he can comment on this.

> 
>     * It is a patch and as such might not apply in the future

True, but since this issue is reported upstream in [2], there is a chance that is it will be addressed there and consecutive `TF-A` release would include a different solution. In this case this patch might be dropped.

> 
>     * Creates work (for me ;-) ).

😉 

> 
> 
> I agree with Otavio that it is a plus that with the patch the fix is binutils version agnostic.
> 
> IMHO the negative aspects outweigh the positive ones, however if the community has a different view I can prepare a PR which reverts the existing fix and adds one which applies the patch from meta-arm.
> 
> Opinions?

-- andrey

Link: [1]: https://lists.yoctoproject.org/g/meta-arm/topic/93266883#3717
Link: [2]: https://developer.trustedfirmware.org/T996

otavio:
Being a patch and conflicting is beneficial to check if it is still needed; I prefer the removal of the error in a way we can do the backport, so we avoid delta (for now) when possible.

PR #1184, opened 2022-08-26
optee-os: fix build with binutils 2.39

zandrey:
binutils 2.39 started to produce warning on RWX LOAD segments, which breaks the linking process of `optee-os`.

Upstream `optee-os` package has patch [`ac425901 ("arm/optee-os: backport RWX permission error patch")`](https://github.com/jonmason/meta-arm/commit/ac4259011f4c8cec01191828014487ccc5de73ca) integrated, which has been backported in OP-TEE from [`PR #5475`](https://github.com/OP-TEE/optee_os/pull/5474) targeting master in upstream.

Port upstream `optee-os` commit into NXP fork of `optee-os`.

**Cc**: @thochstein

### DIFF
--- a/recipes-security/optee-imx/optee-os/0008-no-warn-rwx-segments.patch
+++ b/recipes-security/optee-imx/optee-os/0008-no-warn-rwx-segments.patch
@@ -1,0 +1,64 @@
+Signed-off-by: Anton Antonov <Anton.Antonov@arm.com>
+Upstream-Status: Backport [https://github.com/OP-TEE/optee_os/pull/5474]
+
+From 0b8a917fa51a366806edc0f04b88cd23b24098c4 Mon Sep 17 00:00:00 2001
+From: Jerome Forissier <jerome.forissier@linaro.org>
+Date: Fri, 5 Aug 2022 09:48:03 +0200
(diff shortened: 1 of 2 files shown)


thochstein:
Is the option added only if supported, and thus won't break a kirkstone build?

zandrey:
> Is the option added only if supported, and thus won't break a kirkstone build?

Looking at OP-TEE PR and also Kernel commit [`0d362be5b142 ("Makefile: link with -z noexecstack --no-warn-rwx-segments")`](https://github.com/torvalds/linux/commit/0d362be5b14200b77ecc2127936a5ff82fbffe4), this is not new but rather promoted to WARN with _binutils_ `2.39`.

I did not try this on `kirkstone` since _binutils_ there is still of the version < `2.39`.

-- andrey

thochstein:
The reason I ask is that a [similar issue](https://github.com/Freescale/meta-freescale/pull/1179#issuecomment-1226222053) came up for `imx-atf`, and the patch there, when applied to `kirkstone`, did fail due to `unrecognized option`. I tried this on `kirkstone`, and it does seem to work. I think there may be some magic in the patch to not add the option if not supported.

PR #1194, opened 2022-09-03
imx-atf: Use clang when using toolchain-clang defaults

kraj:
When clang is default compiler, we need to ensure that its added as
dependency as well.

Reported-by: Cliff Brake <cbrake@bec-systems.com>


zandrey:
@kraj Thanks for fixing it!

I've just build-tested it for `imx8mm-lpddr4-evk` machine, `imx-atf` package compiles OK.

`Tested-by: Andrey Zhizhikin <andrey.z@gmail.com> # imx8mm-lpddr4-evk`

otavio (approved):

PR #1202, opened 2022-09-08
openssh: Disable rng-tools

thochstein:
All i.MX beginning with 6 have hardware support for entropy, so there is no need for rng-tools. Removing it helps startup performance.


otavio:
This is a distro option, so please send it to meta-freescale-distro.

thochstein:
https://github.com/Freescale/meta-freescale-distro/pull/94

PR #1208, opened 2022-09-14
gtk+3: Fix logical error in removing x11

kraj:
Last commit had a thinko, where it should be removing x11 only if x11 is not in distro features, but it ended up doing that for wayland

### DIFF
--- a/recipes-graphics/gtk+/gtk+3_%.bbappend
+++ b/recipes-graphics/gtk+/gtk+3_%.bbappend
@@ -1,6 +1,6 @@
 DEPENDS:append:imxgpu2d = " virtual/egl"
 
-WAYLAND = "${@bb.utils.contains('DISTRO_FEATURES', 'x11', '', 'wayland', d)}"
+WAYLAND = "${@bb.utils.contains('DISTRO_FEATURES', 'x11', '', 'x11', d)}"
 WAYLANDONLY = "${@bb.utils.contains('DISTRO_FEATURES', 'wayland', '${WAYLAND}', '', d)}"
 
 PACKAGECONFIG:remove:imxgpu2d = " ${WAYLANDONLY}"


otavio (approved):

recipes-graphics/gtk+/gtk+3_%.bbappend
```diff
@@ -1,6 +1,6 @@
 DEPENDS:append:imxgpu2d = " virtual/egl"
 
-WAYLAND = "${@bb.utils.contains('DISTRO_FEATURES', 'x11', '', 'wayland', d)}"
+WAYLAND = "${@bb.utils.contains('DISTRO_FEATURES', 'x11', '', 'x11', d)}"
 WAYLANDONLY = "${@bb.utils.contains('DISTRO_FEATURES', 'wayland', '${WAYLAND}', '', d)}"
```
thochstein, line 4: The point of the logic was to disable `x11` support if `wayland` is configured. This change just negates that, so might as well just drop the logic entirely, right? I'm doing a build now to see how this affects us.

thochstein: And why do you consider it a logical error? We've had to do this same logic for other packages because the package when configured for X11 would fail because our GPU drivers no longer support GLX.

kraj: it was my logic error :)

thochstein: Ah, of course, my logical error reading ;-)

thochstein: > The point of the logic was to disable `x11` support if `wayland` is configured. This change just negates that, so might as well just drop the logic entirely, right? I'm doing a build now to see how this affects us.

The change builds fine for me. Shall we drop this logic entirely?

PR #1209, opened 2022-09-15
gtk+3: Drop x11 config logic

thochstein:
The original intent of dropping x11 support from the package was so only wayland would be supported. The fix in efcb2fe and 7447f7c removes that. However, it keeps the idea that the x11 packageconfig should be disallowed from being added if x11 is not in the distro features. This was never the intent of the original logic, and since the default is to enable x11 packageconfig if and only if x11 is a distro feature, just drop the logic entirely.


otavio:
@thochstein could you prepare a PR for kirkstone?

thochstein:
Done: #1210

PR #1218, opened 2022-09-19
imx-gpu-viv: Mark libvulkan-imx unavailable for 8M Mini

thochstein:
The GPU drivers do not support 8M Mini, so properly exclude them from PACKAGES so the user gets an earlier indication of a problem:

```
$ bitbake vulkan-loader
ERROR: Nothing RPROVIDES 'libvulkan-imx' (but /.../poky/meta/recipes-graphics/vulkan/vulkan-loader_1.3.216.0.bb RDEPENDS on or otherwise requires it)
ERROR: Required build target 'vulkan-loader' has no buildable providers.
Missing or unbuildable dependency chain was: ['vulkan-loader', 'libvulkan-imx']
```

### DIFF
--- a/recipes-graphics/imx-gpu-viv/imx-gpu-viv-6.inc
+++ b/recipes-graphics/imx-gpu-viv/imx-gpu-viv-6.inc
@@ -71,8 +71,6 @@ PACKAGECONFIG ?= ""
 PACKAGECONFIG[valgrind] = ""
 
 # Note : If you add a package here, to prevent a naming conflict see the python_anonymous() futher below
-IMX_PACKAGES_GBM     = ""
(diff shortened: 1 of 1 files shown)


otavio (approved):

otavio:
@thochstein please check if mx8mm doesn't need SoC suffix change due to this as well.

thochstein:
Do you mean `MACHINE_SOCARCH_SUFFIX`? It is already unique.

munoz0raul:
Hi @thochstein, Do you know if Vulkan is supported for any other iMX8? If you think so, I can try to give it a try with my meta-flutter setup.

thochstein:
Yes, it is supported by all i.MX 8 with a GPU except for 8M Mini.

PR #1222, opened 2022-09-21
imx-gpu-viv: remove vulkan files if not packaged

MaxKrummenacher:
@thochstein 
Wouldn't we need to remove the OpenCL package(s) for the i.MX 8M Mini in the same way we do it for Vulkan?
AFAIK the Mini doesn't support OpenCL too.

@otavio 
This has already been backported to kirkstone and that is where I initially saw the error, so this would eventually need to go into kirkstone too.


thochstein:
Yes, good point. I'll do a PR unless you want to.

thochstein:
Actually let me do it since I've got another PR going as well.

PR #1229, opened 2022-09-26
Some fixes for imx-nxp-bsp

MaxKrummenacher:
The resent removal of gpu features for SoC which don't support the feature had some side effects.
@thochstein  I know that I proposed removing OpenCL, sorry about that.

The commits address three OE build errors.

### DIFF
--- a/recipes-graphics/imx-gpu-viv/imx-gpu-viv-6.inc
+++ b/recipes-graphics/imx-gpu-viv/imx-gpu-viv-6.inc
@@ -213,7 +213,9 @@ do_install () {
         # Framebuffer backend for i.MX 8 and beyond
         install -m 0644 ${S}/gpu-core/usr/lib/pkgconfig/egl.pc         ${D}${libdir}/pkgconfig/egl.pc
         cp -r ${S}/gpu-core/usr/lib/wayland/* ${D}${libdir}
+        set -f
         for f in ${FILES:libvdk-imx} ${FILES:libvdk-imx-dev}; do
+            set +f
             rm -rf ${D}/$f
         done
         rm -rf \
@@ -227,7 +229,9 @@ do_install () {
     fi
 
     if [ -z "${PACKAGES_OPENCL}" ]; then
+        set -f
         for f in ${FILES:libopencl-imx} ${FILES:libopencl-imx-dev} ${FILES:libclc-imx} ${FILES:libclc-imx-dev}; do
+            set +f
             rm -rf ${D}$f
         done
         rm -rf ${D}/opt/viv_samples/cl11
@@ -238,7 +242,9 @@ do_install () {
     fi
 
     if [ -z "${PACKAGES_OPENVX}" ]; then
+        set -f
         for f in ${FILES:libopenvx-imx} ${FILES:libopenvx-imx-dev}; do
+            set +f
             rm -rf ${D}$f
         done
     fi
@@ -254,7 +260,9 @@ do_install () {
         ln -s libvulkan_VSI.so.$FULL ${D}${libdir}/libvulkan_VSI.so.$MAJOR
         ln -s libvulkan_VSI.so.$FULL ${D}${libdir}/libvulkan_VSI.so
         if [ -z "${PACKAGES_VULKAN}" ]; then
+            set -f
             for f in ${FILES:libvulkan-imx} ${FILES:libvulkan-imx-dev}; do
+                set +f
                 rm -rf ${D}$f
             done
         fi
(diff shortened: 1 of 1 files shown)


recipes-graphics/imx-gpu-viv/imx-gpu-viv-6.inc
```diff
@@ -213,7 +213,9 @@ do_install () {
         # Framebuffer backend for i.MX 8 and beyond
         install -m 0644 ${S}/gpu-core/usr/lib/pkgconfig/egl.pc         ${D}${libdir}/pkgconfig/egl.pc
         cp -r ${S}/gpu-core/usr/lib/wayland/* ${D}${libdir}
+        set -f
```
otavio, line 216: @MaxKrummenacher, could you explain more about the need for `set -f`? I looked at this but couldn't find a reason. So after that, we can extend the commit log.

conf/machine/include/imx-base.inc
```diff
@@ -445,6 +445,8 @@ PREFERRED_VERSION_ffmpeg:mx8-nxp-bsp                    ??= "4.4.1"
 SOC_HAS_VIVANTE_KERNEL_DRIVER_SUPPORT        = "0"
 SOC_HAS_VIVANTE_KERNEL_DRIVER_SUPPORT:imxgpu = "1"
 
+DISTRO_FEATURES:remove:mx8mm-nxp-bsp = "vulkan"
```
otavio, line 448: We need to handle this in the mesa bbappend file, not removing this in distro.

recipes-graphics/imx-gpu-viv/imx-gpu-viv-6.inc
MaxKrummenacher: The FILES:xxx variable contains the path to to be deployed files, but does not prefix the path with ${D}.
`for f in ${FILES:libopencl-imx} ${FILES:libopencl-imx-dev} ${FILES:libclc-imx} ${FILES:libclc-imx-dev}; do` will get all elements in the FILES:xxx variables, expand it according to shell rules and the will set each resulting element into `f` and execute in our case the `rm ${D}$f`.

E.g. `${libdir}/libOpenCL${REALSOLIBS}` results in `/usr/lib/libOpenCL.so.*`. If you have on your build host a file or files which matches that pattern `/usr/lib/libOpenCL.so.*` will be replaced by that or those files. Later `rm ${D}$f` will try to remove files with those concreate filenames which likely not be the ones which were installed in the first place.
As a result do_package will the complain about those which were not removed.

If the file expansion didn't find any file matching in the build host filesystem the FILES:xxx entry is passed to `f` with the wildcard and all works. With `set -f` you tell the shell to not expand the wildcards and all works as if there were no matching files in your build host filesystem.

thochstein (approved):
I am working on this so a future release will not have this hard dependency from G2D to OpenCL.

conf/machine/include/imx-base.inc
otavio: So I think we cannot do the removal. We cannot vary distro features as a giant number of recipes would need to become socarch.

MaxKrummenacher: Then I will drop this patch in favor of adding it to a bunch of bbappends.

otavio: Maybe we should revert the vulkan removal and NXP can rework for it avoid creating the surface for Mini. @thochstein ?

thochstein: I will look into what can be done. What I wonder is whether the library already fails at runtime, and maybe the only problem was my attempt to exclude it.

thochstein: > Maybe we should revert the vulkan removal and NXP can rework for it avoid creating the surface for Mini. @thochstein ?

Is this expected to work? If a package is configured to use Vulkan can we really expect the package to work when the libary fails because Vulkan is not supported?

thochstein: Besides, the whole point of excluding Vulkan support was so the user would get a build time failure instead of runtime. Isn't that more important?

Should there be a `vulkan` machine feature?

otavio: Not necessarily, assuming the driver don't allow the misuse of vulkan. I'll ping you for a short call about it tomorrow ;-)

MaxKrummenacher: @otavio @thochstein Let me know how I can help. I now dropped the `remove vulkan from distro features` patch from the PR as it anyway is not the solution.

otavio (approved):

PR #1240, opened 2022-09-29
weston: Fix the append logic

kraj:
Previous commit did not really work because we are using bb.utils.filter and that removed the artificial space sadly. Therefore add it outside this function call.


kraj:
@otavio you were too soon to apply the previous fix. This one really fixes the problem in all cases.

otavio (approved):

PR #1249, opened 2022-10-03
vulkan-loader: iMX8M Mini doesn't support vulkan

quaresmajose:
Vulkan is disabled on meta-freescale for 8M Mini since ee92b5a8, so add the same logic to not recommends libvulkan-imx on this machines.

- imx-base.inc: Introduce imxvulkan override

All iMX8 modules with the imxgpu override support Vulkan with a
single exception for the iMX8M Mini that uses the mx8mm override.
 

- vulkan-loader: use the imxvulkan override in recommends

With this the libvulkan-imx is recommends only when the machine
supports Vulkan.

- imx-gpu-viv: use the imxvulkan override for libvulkan

Signed-off-by: Jose Quaresma <jose.quaresma@foundries.io>

### DIFF
--- a/recipes-graphics/vulkan/vulkan-loader_%.bbappend
+++ b/recipes-graphics/vulkan/vulkan-loader_%.bbappend
@@ -1,3 +1,3 @@
 # The i.MX implementation is dynamically loaded, so it requires an
 # explicit runtime dependency.
-RRECOMMENDS:${PN}:append:imxgpu = " libvulkan-imx"
+RRECOMMENDS:${PN}:append:imxvulkan = " libvulkan-imx"
--- a/conf/machine/include/imx-base.inc
+++ b/conf/machine/include/imx-base.inc
@@ -187,15 +187,15 @@ MACHINEOVERRIDES_EXTENDER:mx7ulp:use-nxp-bsp = "imx-generic-bsp:imx-nxp-bsp:imxf
 
 MACHINEOVERRIDES_EXTENDER:vf:use-nxp-bsp     = "imx-generic-bsp:imx-nxp-bsp:vf-generic-bsp:vf-nxp-bsp"
 
-MACHINEOVERRIDES_EXTENDER:mx8qm:use-nxp-bsp  = "imx-generic-bsp:imx-nxp-bsp:imxdrm:imxdpu:imxgpu:imxgpu2d:imxgpu3d:mx8-generic-bsp:mx8-nxp-bsp:mx8qm-generic-bsp:mx8qm-nxp-bsp"
+MACHINEOVERRIDES_EXTENDER:mx8qm:use-nxp-bsp  = "imx-generic-bsp:imx-nxp-bsp:imxdrm:imxdpu:imxgpu:imxgpu2d:imxgpu3d:imxvulkan:mx8-generic-bsp:mx8-nxp-bsp:mx8qm-generic-bsp:mx8qm-nxp-bsp"
 
 MACHINEOVERRIDES_EXTENDER:mx8mm:use-nxp-bsp  = "imx-generic-bsp:imx-nxp-bsp:imxdrm:imxvpu:imxgpu:imxgpu2d:imxgpu3d:mx8-generic-bsp:mx8-nxp-bsp:mx8m-generic-bsp:mx8m-nxp-bsp:mx8mm-generic-bsp:mx8mm-nxp-bsp"
-MACHINEOVERRIDES_EXTENDER:mx8mn:use-nxp-bsp  = "imx-generic-bsp:imx-nxp-bsp:imxdrm:imxgpu:imxgpu3d:mx8-generic-bsp:mx8-nxp-bsp:mx8m-generic-bsp:mx8m-nxp-bsp:mx8mn-generic-bsp:mx8mn-nxp-bsp"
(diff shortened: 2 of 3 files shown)


recipes-graphics/vulkan/vulkan-loader_%.bbappend
```diff
@@ -1,3 +1,5 @@
 # The i.MX implementation is dynamically loaded, so it requires an
 # explicit runtime dependency.
-RRECOMMENDS:${PN}:append:imxgpu = " libvulkan-imx"
+PACKAGES_VULKAN:aarch64       = "libvulkan-imx"
+PACKAGES_VULKAN:mx8mm-nxp-bsp = ""
+RRECOMMENDS:${PN}:append:imxgpu = " ${PACKAGES_VULKAN}"
```
otavio, line 5:
```
PACKAGES_VULKAN:imxgpu        = " libvulkan-imx"
PACKAGES_VULKAN:mx8mm-nxp-bsp = ""
RRECOMMENDS:${PN}:append      = "${PACKAGES_VULKAN}"
```

otavio (changes requested):

otavio (approved):

recipes-graphics/vulkan/vulkan-loader_%.bbappend
quaresmajose: look better! test local and works

thochstein: We do only support `aarch64`, so the original was correct.

Actually, I was working on this as well and have this:

```
RRECOMMENDS:${PN} += "${RRECOMMENDS_IMX}"

RRECOMMENDS_IMX                = ""
RRECOMMENDS_IMX:aarch64        = "${RRECOMMENDS_IMXGPU}"
RRECOMMENDS_IMX:mx8mm-nxp-bsp  = ""

RRECOMMENDS_IMXGPU             = ""
RRECOMMENDS_IMXGPU:imxgpu      = "libvulkan-imx"
```

otavio: It is not, as otherwise non-imx machines will end appending to it.

quaresmajose: I use the same variable PACKAGES_VULKAN to make it more consistent with the change https://github.com/Freescale/meta-freescale/commit/ee92b5a8b2cad45de267d3cf6dae9251f5ed11fd and as it is a consequence of this change.

@thochstein your solution can add the possibility to additionally select the vulkan lib so I will use your suggestion

otavio: I wonder if we shouldn't have an imxvulkan so we abstract it in imx-base.

thochstein: Yes, a new `imxvulkan` in `MACHINEOVERRIDES_EXTENDER` can work as well.

thochstein:
Note, there's a bigger issue here that we don't support the latest version of Vulkan. To get this working we also need to downgrade to 1.2.182.

thochstein:
It's for `kirkstone` as well.

recipes-graphics/vulkan/vulkan-loader_%.bbappend
quaresmajose: What should I do, I leave it as it is or create a new `imxvulkan` override?

otavio: Let's create the new override, it makes it easier to handle.

thochstein:
I'm working on the downgrade.

recipes-graphics/vulkan/vulkan-loader_%.bbappend
quaresmajose: Thanks, I will do that

quaresmajose: Don't know if this the way of creating an override in this project but here it is, any suggestion is welcome

recipes-graphics/vulkan/vulkan-loader_%.bbappend
```diff
@@ -1,3 +1,3 @@
 # The i.MX implementation is dynamically loaded, so it requires an
 # explicit runtime dependency.
-RRECOMMENDS:${PN}:append:imxgpu = " libvulkan-imx"
+RRECOMMENDS:${PN}:imxvulkan += "libvulkan-imx"
```
otavio, line 3: Don't mix override and `+=`. Use `:append` please.

otavio (changes requested):

recipes-graphics/vulkan/vulkan-loader_%.bbappend
quaresmajose: Sorry! fixed

otavio (approved):

recipes-graphics/vulkan/vulkan-loader_%.bbappend
thochstein: The main recipe has this:

`RRECOMMENDS:${PN} = "mesa-vulkan-drivers"`

We've been doing our releases like this, but I wonder if this was an oversight and should really be replaced by using assignment here instead of append:

`RRECOMMENDS:${PN}:imxvulkan = "libvulkan-imx"`

I can ask Prabhu.

angolini (approved):

thochstein (approved):
This looks good. I can always follow up if `mesa-vulkan-drivers` should be removed.

PR #1252, opened 2022-10-05
Fixes for i.MX Vulkan support

thochstein:


otavio (changes requested):
This must use SOC_ARCH as a package for non-i.MX and mainline BSP will use newer versions.

conf/machine/include/imx-base.inc
```diff
@@ -480,6 +480,10 @@ PREFERRED_VERSION_libdrm:mx6-nxp-bsp ??= "2.4.109.imx"
 PREFERRED_VERSION_libdrm:mx7-nxp-bsp ??= "2.4.109.imx"
 PREFERRED_VERSION_libdrm:mx8-nxp-bsp ??= "2.4.109.imx"
 
+PREFERRED_VERSION_vulkan-headers:imxvulkan ??= "1.2.182.0"
+PREFERRED_VERSION_vulkan-loader:imxvulkan  ??= "1.2.182.0"
+PREFERRED_VERSION_vulkan-tools:imxvulkan   ??= "1.2.182.0"
+
```
otavio, line 486: We need to move packages to SoC architecture when depending on those automatically. I think the following diff does the trick:

```diff
diff --git a/conf/machine/include/imx-base.inc b/conf/machine/include/imx-base.inc
index a0d6be13..dd1d8d5d 100644
--- a/conf/machine/include/imx-base.inc
+++ b/conf/machine/include/imx-base.inc
@@ -306,6 +306,11 @@ MACHINE_SOCARCH_FILTER:append:imxvpu = " \
     libimxvpuapi \
     virtual/imxvpu \
 "
+MACHINE_SOCARCH_FILTER:append:imxvulkan = " \
+    vulkan-headers \
+    vulkan-loader \
+    vulkan-tools \
+"
 MACHINE_SOCARCH_FILTER:append:imxgpu = " \
     virtual/egl \
     virtual/mesa \
```

recipes-graphics/vulkan/vulkan-headers_1.2.182.0.bb
```diff
@@ -0,0 +1,22 @@
...
+
+SRCREV = "37164a5726f7e6113810f9557903a117498421cf"
+
+S = "${WORKDIR}/git"
+
+inherit cmake
+
+FILES:${PN} += "${datadir}/vulkan"
+
+UPSTREAM_CHECK_GITTAGREGEX = "sdk-(?P<pver>\d+(\.\d+)+)"
```
otavio, line 22: I know this is an unmodified recipe copy, and I'd like to make it more prominent and easier to spot when someone is reviewing the layer, so I propose to move those to `recipes-downgrade`  so the reason for those is apparent.

otavio (changes requested):

otavio (approved):

PR #1292, opened 2022-11-11
Add Qt6 support

thochstein:


otavio (changes requested):
@thochstein, thanks for looking at this. I believe this is safe to backport to `kirkstone` as well. However, a few minor things can be improved before we merge this, such as:

- [ ] squash all commits as they can be included consolidated.
- [ ] add an extended commit log summarising what is supported.

PR #1293, opened 2022-11-11
imx-base.inc: Use weakest possible default settings

notapirate:
Use weakest possible default settings in imx-base.inc to ease meta-freescale integration into new projects. These changes should not change any currently set values.


notapirate:
If these commits change the default values heavily depends on the configuration, e.g. if config files are included before imx-base.inc that set these default values. Is there a standard config I can test against?

Part 1 and 3 should be save because they only set defaults new in imx-base.inc (part 1) or newly introduced in meta-freescale (part 3) and not redefined somewhere else in meta-freescale. As for part 2 and 4, I checked against the common default config files in the current poky master branch (bitbake.conf, defaultsetup.conf etc.) that they are either unset or weaker defaults.

zandrey:
@notapirate 
Just for the record, I saw you submitted also 2 patches in _OE-Core_:
- [`[PATCH] initial configs: set initial {DISTRO|MACHINE|IMAGE}FEATUREs as weak as possible`](https://lore.kernel.org/openembedded-core/20221112163103.42796-1-christoph.lauer@email.de/#r)
- [`[PATCH] initial configs: set PREFERRED_PROVIDERs as weaker defaults`](https://lore.kernel.org/openembedded-core/20221112155815.39110-1-christoph.lauer@email.de/#r)

Is this PR also related to those patches? Is it a part of bigger work that you do across all layers?

-- andrey

notapirate:
hi @zandrey, the OE-Core patches share the same motivation to be able to use weak defaults whenever possible, but there are no interdependencies. The OE-Core patches 'weaken' VIRTUAL_PROVIDER and FEATURE defaults, the meta-freescale patches do not touch these values.

angolini:
@otavio, could you please wait a bit before merging? I want to give it a try

zandrey:
@notapirate Looking at the progress of above mentioned patches in _OE-Core_, I saw that you decided that changes are not necessary.

Is it the case for this layer as well?

If not - then I would be interested in the answer to the question [Alex posted as a response to your patch](https://lore.kernel.org/openembedded-core/CANNYZj-CrvJpLCNFTi9Q__NWBc_25nUJqKwTqpEsHbU8wvCo6A@mail.gmail.com/).

-- andrey

notapirate:
@zandrey I'm providing a base layer for the machines that my company uses. So far I'm using weak or weakest values in my machine and distro files. I would like to include imx-base.inc to profit from all features of meta-freescale layer. To keep my current configs I either need at least weak defaults in imx-base.inc and include it at the end of my configs, or I use non-weak values and include imx-base.inc at the beginning of my configs.

With these patches I followed the first approach, but by now I believe the second approach should work as well since all values can be overwritten without a problem. An argument for weak values would also be that they could be set in local.conf or externally, but I don't need that feature, so that's more theoretical.

zandrey:
Hello @notapirate 

> @zandrey I'm providing a base layer for the machines that my company uses. So far I'm using weak or weakest values in my machine and distro files. I would like to include imx-base.inc to profit from all features of meta-freescale layer. To keep my current configs I either need at least weak defaults in imx-base.inc and include it at the end of my configs, or I use non-weak values and include imx-base.inc at the beginning of my configs.

Later I believe is the case that is used by vast majority of people (and layers) that are deriving their machines from `imx-base.inc`. So far this approach has been working, and I do not see the reason why it cannot be adopted in your case (considering that you've tried it out already).

Setting defaults using `??=` can have some dare consequences, and this was already pointed out in response to your patches in the _OE-Core_, where `qemu` target was broken.

> 
> With these patches I followed the first approach, but by now I believe the second approach should work as well since all values can be overwritten without a problem. An argument for weak values would also be that they could be set in local.conf or externally, but I don't need that feature, so that's more theoretical.

I would strongly discourage you (or anyone else) to include any setting in either `local.conf` or externally. Layer should have all those variables set, so that setup is deterministic and reproducible.

All-in-all, I see that this change is not required from your side anymore, right? If so - can you please close this PR then?

Thanks a lot!

-- andrey

thochstein:
This behavior of `+=` to generate an empty default is unfortunate. Generally I'd like to use `+=` since it's shorter but this side effect of negating `??=` is a gotcha. It does seem the bitbake manual should mention it.

I did recently find `?=` particularly limiting: https://github.com/Freescale/meta-freescale/commit/5b47898ff939791ab426fff17821aca3476ddd7a

notapirate:
> All-in-all, I see that this change is not required from your side anymore, right? If so - can you please close this PR then?
You're right, this change is not mandatory for me, and since there are some big uncertainties, I'm closing the pull request. Thanks @all for your time.

angolini:
I want to add a comment for the future.
I see value in this kind of change, but in the future, if this is needed again, let's try to tackle this in shorter PRs, making a few changes at a time.

PR #1302, opened 2022-11-16
linux-fslc: add dependency on lzop-native

twoerner:
linux-fslc do_compile fails with:
	...
	|   SYSMAP  System.map
	|   SORTTAB vmlinux
	|   OBJCOPY arch/arm/boot/Image
	|   Kernel: arch/arm/boot/Image is ready
	|   LDS     arch/arm/boot/compressed/vmlinux.lds
	|   AS      arch/arm/boot/compressed/head.o
	|   LZO     arch/arm/boot/compressed/piggy_data
	| /bin/sh: lzop: command not found


twoerner:
@otavio this adds a dependency on meta-openembedded. I was going to encode that in the README and conf/layer.conf, however the linux-imx recipe has had this dependency on lzop-native since 64d58e36 (linux-imx: Update kernel recipe to version 4.19.35, 2019-12-02)

This dependency is only needed if you're building either linux-imx or linux-fslc. iow it's not a dependency that is always needed.

Let me know if you want me to update the patch to add the dependency on the meta-openembedded layer.

zandrey:
@twoerner 

Thanks for the finding!

This actually was my fault in the latest Kernel PR, where I accidentally brought back the `CONFIG_KERNEL_LZO` Kernel config option.

It has been dropped in 3a2cc5d4995c2a0c45c2099b1fab03afa5090b8b, but I missed this part in latest update.

`lzop` has been dropped from _OE-Core_ by Ross with commit [`dea5e88637 ("lzop: remove recipe from oe-core")`](https://github.com/openembedded/openembedded-core/commit/dea5e8863792dc7bb3324b543e04da4c94a060aa), so I would not use it and re-introduce back in the layer.

There is a need to do a backport of 3a2cc5d4995c2a0c45c2099b1fab03afa5090b8b, you can do it in this PR or you prefer me to do it?

-- andrey

otavio:
@twoerner would you mind to rework the PR as @zandrey suggested?

twoerner:
Yes, I will rework the PR and do the backport.

otavio (approved):

PR #1311, opened 2022-11-25
imx-boot-container: add extra firmware copy from deploydir

gibsson:
Needed for platforms requiring other firmware files than the DDR (like the HDMI/DP firmware on i.MX 8MQ).

Signed-off-by: Gary Bisson <gary.bisson@boundarydevices.com>

### DIFF
--- a/classes/imx-boot-container.bbclass
+++ b/classes/imx-boot-container.bbclass
@@ -28,6 +28,16 @@
 ATF_MACHINE_NAME = "bl31-${ATF_PLATFORM}.bin"
 ATF_MACHINE_NAME:append = "${@bb.utils.contains('MACHINE_FEATURES', 'optee', '-optee', '', d)}"
 
+IMX_BOOT_CONTAINER_FIRMWARE_SOC = ""
+IMX_BOOT_CONTAINER_FIRMWARE_SOC:mx8mq-generic-bsp = " \
+    signed_dp_imx8m.bin \
+    signed_hdmi_imx8m.bin \
+"
+IMX_BOOT_CONTAINER_FIRMWARE ?= " \
+    ${IMX_BOOT_CONTAINER_FIRMWARE_SOC} \
+    ${DDR_FIRMWARE_NAME} \
+"
+
 # This package aggregates output deployed by other packages, so set the
 # appropriate dependencies for populate binaries task
 do_resolve_and_populate_binaries[depends] += " \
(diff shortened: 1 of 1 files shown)


gibsson:
After switching to `imx-boot-container` we've seen that the binary generated for imx8mq wasn't booting properly.
The issue is that the HDMI/DP firmware are not properly copied.
So this patch adds a copy of all `EXTRA_FIRMWARE_NAME` listed in the config file (to be added for nitrogen8m in our case).
Not sure it's the best option. I was tempted to have one single `IMX_BOOT_CONTAINER_FIRMWARE` variable with all the binaries to be copied from the deploy dir to the build one.
Let me know which approach you like the most.
Also the next question will be the OPTEE firmware, because it's in the dependency but AFAICT it's not copied.

otavio:
I'd personally prefer it to be handled with default values inside the class. I think there is no need for this to be defined for every machine which uses them.

angolini:
I like the approach of using default values in the class

angolini (approved):

gibsson:
@otavio indeed it makes sense to have it added automatically without changing the config.
So in that new revision I added the following, let me know if you're ok with it:
```
EXTRA_FIRMWARE_NAME = ""
EXTRA_FIRMWARE_NAME:mx8mq-generic-bsp = " \
    signed_dp_imx8m.bin \
    signed_hdmi_imx8m.bin \
"
```

otavio:
Could you use a more descriptive variable and make it overridable?

I think something as:

```c
SOC_IMX_BOOT_CONTAINER_FIRMWARE = ""
SOC_IMX_BOOT_CONTAINER_FIRMWARE:mx8mq-generic-bsp = " \
    signed_dp_imx8m.bin \
    signed_hdmi_imx8m.bin \
"
IMX_BOOT_CONTAINER_FIRMWARE ?= ${SOC_IMX_BOOT_CONTAINER_FIRMWARE}
```

gibsson:
Sure, i can do that. I'll respin it in an hour or so. Also, do you want to keep the DDR firmware apart or do you want them to be part of IMX_BOOT_CONTAINER_FIRMWARE? I'm not sure why it would be different, nor do I know what is the ddr_firmware_name is for currently.

otavio:
I'd love to consolidate this. @zandrey ?

thochstein:
Please use `IMX_BOOT_CONTAINER_FIRMWARE_SOC` instead. I think variables that are more specific should be named by appending, not prepending. It makes it easier to debug as well with `bitbake -g`.

gibsson:
@otavio see the latest version, it does a generic copy from deploydir to builddir for all firmware files.
Also I've followed @thochstein recommendation about the variable naming.

otavio (approved):

classes/imx-boot-container.bbclass
```diff
@@ -28,6 +28,16 @@
 ATF_MACHINE_NAME = "bl31-${ATF_PLATFORM}.bin"
 ATF_MACHINE_NAME:append = "${@bb.utils.contains('MACHINE_FEATURES', 'optee', '-optee', '', d)}"
 
+IMX_BOOT_CONTAINER_FIRMWARE_SOC = ""
+IMX_BOOT_CONTAINER_FIRMWARE_SOC:mx8mq-generic-bsp = " \
+    signed_dp_imx8m.bin \
+    signed_hdmi_imx8m.bin \
```
zandrey, line 34: Who is the provider of this firmware? It shall participate in `IMX_EXTRA_FIRMWARE`, so that the dependencies are resolved in `do_resolve_and_populate_binaries[depends]` invocation.

Otherwise, the build might break as `do_resolve_and_populate_binaries ()` would not be able to find those blobs in some cases.

zandrey (commented):
Sorry for late reply here, just one question from my side.

classes/imx-boot-container.bbclass
gibsson: It's already taken care of as part of firmware-imx:
https://github.com/Freescale/meta-freescale/blob/master/recipes-bsp/firmware-imx/firmware-imx-8m_8.17.bb#L17

zandrey: Great, thanks for confirming this! 👍🏻

PR #1322, opened 2022-12-07
jailhouse: update to lf-5.15.52_2.1.0

ricardosalveti:
Update jailhouse to the NXP lf-5.15.52_2.1.0 release, and clean up the packaging logic to align with the latest release.

Packaging changes:
- Drop build instructions for tools as that is now built as part of the
  main Makefile
- Drop JH_CONFIG and related tools variables as they are not required
- Change do_compile and do_install to prepend to allow the logic from
  setuptools3 to be executed (and package pyjailhouse properly)
- Drop kernel from PACKAGE_BEFORE_PN, as it wasn't really used in the
  end, and doesn't work correctly with usrmerge. Prefer removing libdir
  that gets appended by setuptools3 for module split to work correctly
  with and without usrmerge.
- Update FILES for pyjailhouse to include all the python related files
   produced by the build

### DIFF
--- a/recipes-extended/jailhouse/jailhouse_0.12.bb
+++ b/recipes-extended/jailhouse/jailhouse_0.12.bb
@@ -13,11 +13,10 @@ LIC_FILES_CHKSUM = "file://COPYING;md5=9fa7f895f96bde2d47fd5b7d95b6ba4d \
                  file://driver/jailhouse.h;beginline=9;endline=36;md5=2825581c1666c44a17955dc574cfbfb3 \
 "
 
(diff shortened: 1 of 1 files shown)


ricardosalveti:
It seems the cache metadata is not happy with kernel version there and PACKAGE_BEFORE_PN, testing it a bit more (moving to draft)

ricardosalveti:
All good now after removing the usage of kernel version and removing kernel from PACKAGE_BEFORE_PN.

ricardosalveti:
@quaresmajose @angolini ^^

quaresmajose:
Thanks for fixing the remaining issues.

PR #1332, opened 2022-12-15
weston-init: Drop redundant xwayland=true setting

thochstein:
The main recipe now sets xwayland=true in weston.ini, so drop it from here.


otavio (approved):

otavio:
Do you want this in kirkstone as well?

thochstein:
Ah, good question, I just checked and the `weston-init` update starts with `langdale`, so this is not for `kirkstone`.

PR #1346, opened 2022-12-21
Add NXP WiFi/BT support

thochstein:
Add NXP WiFi/BT firmware support, enabled by `MACHINE_FEATURES`, and tied into existing NXP development boards.

### DIFF
--- a/recipes-bsp/firmware-imx/firmware-nxp-wifi_1.0.bb
+++ b/recipes-bsp/firmware-imx/firmware-nxp-wifi_1.0.bb
@@ -82,7 +82,6 @@ FILES:${PN}-bcm4359-pcie = " \
     ${nonarch_base_libdir}/firmware/brcm/brcmfmac4359-pcie.* \
     ${sysconfdir}/firmware/BCM4349B1_*.hcd \
 "
-RPROVIDES:${PN}-bcm4359-pcie = "linux-firmware-bcm4359-pcie"
 
 FILES:${PN}-nxp-common = " \
     ${nonarch_base_libdir}/firmware/nxp/wifi_mod_para.conf \
--- a/recipes-bsp/firmware-imx/firmware-nxp-wifi_1.0.bb
+++ b/recipes-bsp/firmware-imx/firmware-nxp-wifi_1.0.bb
@@ -8,10 +8,11 @@ recipe in favor of upstream."
 
 SECTION = "kernel"
 LICENSE = "Proprietary"
-LIC_FILES_CHKSUM = "file://cyw-wifi-bt/EULA.txt;md5=80c0478f4339af024519b3723023fe28"
+LIC_FILES_CHKSUM = "file://EULA.txt;md5=be5ff43682ed6c57dfcbeb97651c2829"
 
-SRC_URI = "git://github.com/NXP/imx-firmware.git;protocol=https;branch=master"
-SRCREV = "484d38224fa2c26b8859a7bf20b7c4d49100f5bc"
+SRC_URI = "git://github.com/NXP/imx-firmware.git;protocol=https;branch=${SRCBRANCH}"
+SRCBRANCH = "lf-5.15.52_2.1.0"
+SRCREV = "b6f070e3d4cab23932d9e6bc29e3d884a7fd68f4"
 
 S = "${WORKDIR}/git"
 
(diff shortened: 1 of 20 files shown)


recipes-bsp/firmware-imx/firmware-nxp-wifi_1.0.bb
```diff
@@ -26,21 +27,97 @@ do_compile() {
...
+    ${sysconfdir}/firmware/BCM4349B1_*.hcd"
 RPROVIDES:${PN}-bcm4359-pcie = "linux-firmware-bcm4359-pcie"
+
+FILES:${PN}-nxp-common = " \
+    ${nonarch_base_libdir}/firmware/nxp/wifi_mod_para.conf"
+
+FILES:${PN}-nxp8801-sdio = " \
+    ${nonarch_base_libdir}/firmware/nxp/*8801*"
+RDEPENDS:${PN}-nxp8801-sdio  += "${PN}-nxp-common"
+RPROVIDES:${PN}-nxp8801-sdio += "linux-firmware-sd8801"
```
otavio, line 91: Why are you adding this `RPROVIDES` set? It doesn't come from `linux-firmware`, and we should use the right package in depends on allowing easier find of what provides them. It also allows easier rework if those are included in `linux-firmware`.

otavio (changes requested):

otavio (approved):

recipes-bsp/firmware-imx/firmware-nxp-wifi_1.0.bb
```diff
@@ -80,7 +80,6 @@ PACKAGES =+ " \
 FILES:${PN}-bcm4359-pcie = " \
     ${nonarch_base_libdir}/firmware/brcm/brcmfmac4359-pcie.* \
     ${sysconfdir}/firmware/BCM4349B1_*.hcd"
-RPROVIDES:${PN}-bcm4359-pcie = "linux-firmware-bcm4359-pcie"
```
angolini: I don't understand the reason for dropping it (the one provided in the commit log) Could you please elaborate a bit?

Is it only a matter of it being replaced?

angolini (approved):

recipes-bsp/firmware-imx/firmware-nxp-wifi_1.0.bb
thochstein: It was Otavio that had specific design reasons not to make an alias from this recipe to the `linux-firmware` packages.

@otavio, can you give me the reasoning as you see it and I'll add it to the commit message?

recipes-bsp/firmware-imx/firmware-nxp-wifi_1.0.bb
```diff
@@ -26,21 +27,90 @@ do_compile() {
...
     ${PN}-bcm4359-pcie \
-"
+    ${PN}-nxp-common \
+    ${PN}-nxp8801-sdio \
+    ${PN}-nxp8987-sdio \
+    ${PN}-nxp8997-common \
+    ${PN}-nxp8997-pcie \
+    ${PN}-nxp8997-sdio \
+    ${PN}-nxp9098-pcie \
+    ${PN}-nxp9098-sdio"
```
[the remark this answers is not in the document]
thochstein: I've gotten to the point where I don't like putting a quotation mark on its own line. It does in fact also add a line between this assignment and the next assignment, which is not ideal, while the benefit of having each value line end consistently with `\` doesn't seem particularly valuable.

For example, I prefer the `VARIABLE_A` syntax to `VARIABLE_B`:
```
VARIABLE_A = " \
    ${VARIABLE_A1} \
    ${VARIABLE_A2}"
VARIABLE_A1 = " \
    a1-1 \
    a1-2"
VARIABLE_A2 = " \
    a2-1 \
    a2-2"

VARIABLE_B = " \
    ${VARIABLE_B1} \
    ${VARIABLE_B2} \
    "
VARIABLE_B1 = " \
    b1-1 \
    b1-2 \
    "
VARIABLE_B2 = " \
    b2-1 \
    b2-2 \
    "
```

recipes-bsp/firmware-imx/firmware-nxp-wifi_1.0.bb
otavio: Sure. This package doesn't aim to be a replacement for `linux-firmware` but to offer different firmware files. Forcing user to use the `firmware-nxp-wifi` allow for easier find for them if we eventually have a replacement in `linux-firmware`.

recipes-bsp/firmware-imx/firmware-nxp-wifi_1.0.bb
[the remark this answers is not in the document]
otavio: Personally, I disagree. I prefer:
```
VARIABLE_B = " \
    ${VARIABLE_B1} \
    ${VARIABLE_B2} \
"
VARIABLE_B1 = " \
    b1-1 \
    b1-2 \
"
VARIABLE_B2 = " \
    b2-1 \
    b2-2 \
"
```
as it has some nice properties:
- reduce the number of changed lines if adding new lines
- allow visual check where the block ends
 
-

otavio (approved):

PR #1352, opened 2022-12-30
Upgrade kernel & u-boot

### LINKED ISSUE #1341: Upgrade Linux kernel and U-Boot (i.MX variants)

thochstein:
Fixes #1341.


recipes-bsp/u-boot/u-boot-imx-mfgtool_2022.04.bb
```diff
@@ -2,8 +2,6 @@
 # Copyright (C) 2014-2016 Freescale Semiconductor
 # Copyright 2017-2019 NXP
 
-FILESEXTRAPATHS:prepend := "${THISDIR}/u-boot-imx:"
```
otavio, line 5: Really?

otavio (approved):

recipes-bsp/u-boot/u-boot-imx-mfgtool_2022.04.bb
thochstein: The trouble comes from including the `u-boot-imx` recipe.

otavio: But why is it causing problems? Usually, for this kind of recipe, we use the base recipe files as a place for patches and similarly used on both recipes.

If the `u-boot-imx` recipe has a patch file, I believe it will fail to find it.

thochstein: No problems. I dropped it because I misunderstood the deeper reason for the line and thought I was just doing a simple cleanup. I will restore it and add a comment to make the intent clear.

otavio (approved):

PR #1389, opened 2023-01-13
imx-base.inc: imx8mn: add Asynchronous Sample Rate Converter firmware

arpaqt:
For all i.MX8MN devices, we ought to add the `firmware-imx-easrc-imx8mn` by
default so the driver can work as expected.

See commit
https://github.com/nxp-imx/linux-imx/commit/970406eaef3a5304cc1513d8a4aae23e183f7ba8

Also for Yocto/Kirkstone, etc.


otavio (changes requested):
I am fine with the change but the commit log needs improvement. Please rework and extend it.

arpaqt:
@otavio Do you mean commit message should follow some formal template or maybe that it should be a bit more verbose; Rationale itself is quite clear in the commit referred?
_easrc: easrc@300c0000 {
  compatible = "fsl,imx8mn-easrc";
  firmware-name = "imx/easrc/easrc-imx8mn.bin";_

otavio:
Something as:
```
imx-base.inc: imx8mn: add Asynchronous Sample Rate Converter firmware

For all i.MX8MN devices, we ought to add the `firmware-imx-easrc-imx8mn` by
default so the driver can work as expected.
```

otavio:
You missed the signed-off-by.

PR #1393, opened 2023-01-19
Topic/u boot fix

mdrodrigo:


recipes-bsp/u-boot/files/0001-Remove-video-config.patch
```diff
@@ -0,0 +1,24 @@
+From 7ff28a5520475c207b67c04b05fe32d42461215e Mon Sep 17 00:00:00 2001
+From: "Rodrigo M. Duarte" <rodrigo.duarte@ossystems.com.br>
+Date: Mon, 16 Jan 2023 15:05:00 -0300
+Subject: [PATCH] Remove video config
+
+---
+ configs/sandbox_defconfig | 1 -
```
otavio, line 7: I believe we should use the `tools-only_defconfig` so we avoid adding this extra patch.

otavio (changes requested):

recipes-bsp/u-boot/u-boot-fslc-mxsboot_2023.01.bb
```diff
@@ -3,7 +3,19 @@ require u-boot-fslc-common_${PV}.inc
 DESCRIPTION = "U-boot bootloader mxsboot tool"
 SECTION = "bootloader"
 
-DEPENDS = "swig-native bison-native gnutls-native dtc openssl"
+inherit python3native
+
+PACKAGE_ARCH = "${MACHINE_ARCH}"
```
otavio, line 8: Why?

mdrodrigo: Fixed. Forgotten in the Last review.

PR #1398, opened 2023-01-24
Add i.MX 8ULP Machines

angolini:
I have only built the source code. I don't have the board so I cannot test it. In case anyone is available to test, please, let me know the result.

It's important to note that the `IMXBOOT_TARGETS` tries to use dualboot, but the `m33` binaries are not included:

```
IMXBOOT_TARGETS_SD   = "flash_singleboot flash_dualboot"
IMXBOOT_TARGETS_FSPI = "flash_dualboot_flexspi"
IMXBOOT_TARGETS_ND   = ""
```

### DIFF
--- a/conf/machine/imx8ulp-lpddr4-evk.conf
+++ b/conf/machine/imx8ulp-lpddr4-evk.conf
@@ -1,0 +1,27 @@
+#@TYPE: Machine
+#@NAME: i.MX 8ULP EVK
+#@SOC: i.MX8ULP
+#@DESCRIPTION: Machine configuration for NXP i.MX 8ULP Evaluation Kit with LPDDR4
+#@MAINTAINER: Jun Zhu <junzhu@nxp.com>
+
+require include/imx8ulp-evk.inc
+
+KERNEL_DEVICETREE_BASENAME = "imx8ulp-evk"
+
+KERNEL_DEVICETREE += " \
+    freescale/${KERNEL_DEVICETREE_BASENAME}-epdc.dtb \
+    freescale/${KERNEL_DEVICETREE_BASENAME}-flexio-i2c.dtb \
+    freescale/${KERNEL_DEVICETREE_BASENAME}-i3c.dtb \
+    freescale/${KERNEL_DEVICETREE_BASENAME}-lpa.dtb \
+    freescale/${KERNEL_DEVICETREE_BASENAME}-lpspi-slave.dtb \
+    freescale/${KERNEL_DEVICETREE_BASENAME}-nd.dtb \
+    freescale/${KERNEL_DEVICETREE_BASENAME}-rk055hdmipi4m.dtb \
+    freescale/${KERNEL_DEVICETREE_BASENAME}-rk055hdmipi4mv2.dtb \
+    freescale/${KERNEL_DEVICETREE_BASENAME}-sof-btsco.dtb \
+"
+
+UBOOT_CONFIG_BASENAME = "imx8ulp_evk"
+
+MACHINE_FEATURES += "soc-reva1"
+UPOWER_FIRMWARE_NAME = "upower_a1.bin"
+IMX_DEFAULT_BSP = "nxp"
--- a/conf/machine/include/imx8ulp-evk.inc
+++ b/conf/machine/include/imx8ulp-evk.inc
@@ -1,0 +1,44 @@
+MACHINEOVERRIDES =. "mx8:mx8ulp:"
+
+require conf/machine/include/imx-base.inc
+require conf/machine/include/arm/armv8a/tune-cortexa35.inc
+
+MACHINE_FEATURES += "pci wifi bluetooth optee jailhouse"
+
+KERNEL_DEVICETREE = " \
+    freescale/${KERNEL_DEVICETREE_BASENAME}.dtb \
+"
+
+IMX_DEFAULT_BOOTLOADER:use-nxp-bsp = "u-boot-imx"
+IMX_DEFAULT_BOOTLOADER:use-mainline-bsp = "u-boot-fslc"
+
+LOADADDR = ""
+UBOOT_SUFFIX = "bin"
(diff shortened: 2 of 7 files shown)


thochstein:
+@junzhuimx 

Thanks, @angolini, but unfortunately 8ULP is not in production yet. Our website does show a single board:

https://www.nxp.com/products/power-management/pmics-and-sbcs/pmics/pca9460-evaluation-board:KIT-PCA9460-EVB

The name is different. Maybe we should hold off on this?

conf/machine/imx8ulp-9x9-lpddr4-evk.conf
```diff
@@ -0,0 +1,23 @@
+#@TYPE: Machine
+#@NAME: i.MX 8ULP 9x9 EVK
+#@SOC: i.MX8ULP
+#@DESCRIPTION: Machine configuration for NXP i.MX 8ULP 9x9 Evaluation Kit with LPDDR4
```
thochstein, line 4: These boards with `9x9` in the name seem to be only for internal use, so not upstreamed.

conf/machine/imx8ulpevk.conf
```diff
@@ -0,0 +1,20 @@
+#@TYPE: Machine
+#@NAME: i.MX 8ULP EVK
+#@SOC: i.MX8ULP
+#@DESCRIPTION: Machine configuration for NXP i.MX 8ULP Evaluation Kit
```
thochstein, line 4: Don't include this file either, we still haven't fully adapted since the machine renames.

angolini: Thanks @thochstein , let me rework the patchset.

angolini: @thochstein done!

angolini:
@thochstein I still don't have access to the board, but in the future, we are going to have it. It would be good to test with the reviewed patchset. Would you mind reviewing the PR?

conf/machine/imx8ulp-lpddr4-evk.conf
```diff
@@ -0,0 +1,28 @@
...
+KERNEL_DEVICETREE += " \
+    freescale/imx8ulp-evk-epdc.dtb \
+    freescale/imx8ulp-evk-flexio-i2c.dtb \
+    freescale/imx8ulp-evk-i3c.dtb \
+    freescale/imx8ulp-evk-lpa.dtb \
+    freescale/imx8ulp-evk-lpspi-slave.dtb \
+    freescale/imx8ulp-evk-nd.dtb \
+    freescale/imx8ulp-evk-rk055hdmipi4m.dtb \
+    freescale/imx8ulp-evk-rk055hdmipi4mv2.dtb \
+    freescale/imx8ulp-evk-sof-btsco.dtb \
```
thochstein, line 20: Use `${KERNEL_DEVICETREE_BASENAME}` here.

conf/machine/include/imx8ulp-evk.inc
```diff
@@ -0,0 +1,44 @@
+MACHINEOVERRIDES =. "mx8:mx8ulp:"
```
thochstein, line 1: I believe this should be just `MACHINEOVERRIDES =. "mx8ulp"`.

That said, when I grep I see the same pattern for a few more SOCs.

```
conf/machine/imx7dsabresd.conf:MACHINEOVERRIDES =. "mx7:mx7d:"
conf/machine/imx7ulpevk.conf:MACHINEOVERRIDES =. "mx7:mx7ulp:"
conf/machine/imx8qm-mek.conf:MACHINEOVERRIDES =. "mx8:mx8qm:"
```

@otavio, can you please confirm that `mx8` is not needed here?

conf/machine/imx8ulp-lpddr4-evk.conf
thochstein: Thanks, @angolini, but there are two more parts, using `append:use-nxp-bsp`, and moving to `.inc`. Actually, leaving it here is fine for now. We can move anything common to `.inc` if/when we add `ddr4` version.

recipes-bsp/firmware-sentinel/firmware-sentinel_0.8.bb
```diff
@@ -0,0 +1,25 @@
...
+
+inherit fsl-eula-unpack use-imx-security-controller-firmware deploy
+
+SRC_URI = "${FSL_MIRROR}/${BP}.bin;fsl-eula=true"
+SRC_URI[md5sum] = "be47a5e59c1192ee36246af97d5d1532"
+SRC_URI[sha256sum] = "1003d4c6773c153ea341911a74e25c249423644f70f3d8f8d085599e00770b3f"
+
+do_compile[noexec] = "1"
+
+do_install() {
```
otavio, line 16:
```
do_install[noexec] = "1"
```

otavio: Please check if the nopackages class might be used.

conf/machine/include/imx8ulp-evk.inc
otavio: It is. It is meant to reuse the `mx8` common override.

otavio (changes requested):

conf/machine/include/imx8ulp-evk.inc
thochstein: If that is the case, then it seems surprising we don't see the pattern more.

angolini:
@otavio @thochstein please let me know if there is anything else I'm missing, because I think there is nothing else from my side so far

otavio (approved):

conf/machine/include/imx8ulp-evk.inc
thochstein: Isn't `mx8` unused since we changed all the overrides to `mx8-*-bsp`.

PR #1414, opened 2023-02-07
Initial support for i.MX 93 machines

angolini:
I have tested the PR on `imx93-11x11-lpddr4x-evk` and it boots.

While it's been reviewed, I'm working to fix `linux-fslc-imx` because today it does not boot for this board


conf/machine/include/imx-base.inc
```diff
@@ -164,6 +164,7 @@ DEFAULTTUNE:mx8m-generic-bsp   ?= "cortexa53-crypto"
 DEFAULTTUNE:mx8qm-generic-bsp  ?= "cortexa72-cortexa53-crypto"
 DEFAULTTUNE:mx8x-generic-bsp   ?= "cortexa35-crypto"
 DEFAULTTUNE:mx8ulp-generic-bsp ?= "cortexa35-crypto"
+DEFAULTTUNE:mx9-generic-bsp    ?= "cortexa35-crypto"
```
zandrey, line 167: Not sure this tune is correct here.

IIRC, `imx93` has _A55_ and it is closer to _A53_ than to _A35_.

There is no `DEFAULTTUNE` for _A55_ available in _OE-Core_, but in this case I would rather set the tune to `cortexa53-crypto` instead.

recipes-bsp/firmware-imx/firmware-imx-9_8.18.bb
```diff
@@ -0,0 +1,21 @@
...
+    # Synopsys DDR
+    for ddr_firmware in ${DDR_FIRMWARE_NAME}; do
+        install -m 0644 ${S}/firmware/ddr/synopsys/${ddr_firmware} ${DEPLOYDIR}
+    done
+}
+addtask deploy after do_install before do_build
+
+PACKAGE_ARCH = "${MACHINE_SOCARCH}"
+
+COMPATIBLE_MACHINE = "(mx9-nxp-bsp)"
```
zandrey, line 21: Why is this file required? It is a copy of `firmware-imx-8m`, which can be used in this context as well.

conf/machine/imx93evk.conf
```diff
@@ -0,0 +1,20 @@
+#@TYPE: Machine
```
thochstein, line 1: This file is used by NXP for internal testing and shouldn't be upstreamed.

conf/machine/include/imx-base.inc
thochstein: Looks like A55 is available?

https://github.com/openembedded/openembedded-core/blob/master/meta/conf/machine/include/arm/armv8-2a/tune-cortexa55.inc

recipes-kernel/linux/linux-fslc-imx/mx9-nxp-bsp/defconfig
```diff
@@ -0,0 +1,1071 @@
+CONFIG_SYSVIPC=y
```
thochstein, line 1: What is the purpose of the default `defconfig` for all `mx9`? At the least the commit should explain why we add it, but perhaps even better would be to not add it until it is actually used?

conf/machine/include/imx-base.inc
zandrey: True, thanks for the hint!

Sorry, I was searching in the different folder (`armv8a`) and did not realize that _OE-Core_ had a separate folder laid out for _ARMv8.2-A_ architecture.

Then it's indeed not correct - the tune shall be set to point to the correct one.

classes/use-imx-security-controller-firmware.bbclass
```diff
@@ -24,6 +24,7 @@ SECO_FIRMWARE_NAME:mx8qxp-nxp-bsp = \
 SECO_FIRMWARE_NAME:mx8dx-nxp-bsp  = "mx8qxc0-ahab-container.img"
 SECO_FIRMWARE_NAME:mx8dxl-nxp-bsp = "mx8dxla1-ahab-container.img"
 SECO_FIRMWARE_NAME:mx8ulp-nxp-bsp = "mx8ulpa1-ahab-container.img"
+SECO_FIRMWARE_NAME:mx9-nxp-bsp    = "mx93a0-ahab-container.img"
```
otavio, line 27: Please, use `mx93` as override.

conf/machine/include/imx-base.inc
```diff
@@ -372,6 +379,7 @@ IMX_EXTRA_FIRMWARE:mx8-generic-bsp   = "firmware-imx-8 imx-sc-firmware imx-seco"
 IMX_EXTRA_FIRMWARE:mx8m-generic-bsp  = "firmware-imx-8m"
 IMX_EXTRA_FIRMWARE:mx8x-generic-bsp  = "imx-sc-firmware imx-seco"
 IMX_EXTRA_FIRMWARE:mx8ulp-generic-bsp = "firmware-upower firmware-sentinel"
+IMX_EXTRA_FIRMWARE:mx9-nxp-bsp        = " firmware-imx-8 firmware-sentinel"
```
otavio, line 382: Please, use `mx93` override.

recipes-bsp/firmware-imx/firmware-imx-9_8.18.bb
otavio: This file could be removed, right?

recipes-kernel/linux/linux-fslc-imx/mx9-nxp-bsp/defconfig
otavio: I'd let the `mx93` one, only.

otavio (changes requested):

recipes-kernel/linux/linux-fslc-imx/mx9-nxp-bsp/defconfig
angolini: I will fix the commit log, but they are the same defconfig, but one for `linux-imx`and one for `linux-fslc-imx`. It happens that today they are the same.
Is there a better way to have one file only for both linux?

conf/machine/include/imx-base.inc
angolini: changed

recipes-kernel/linux/linux-fslc-imx/mx9-nxp-bsp/defconfig
thochstein: Ah, I did not notice the two files are for different linux. For `linux-imx`, I see that this is the same `defconfig` as for all i.MX 8. Perhaps we can link `linux-imx/imx9-nxp-bsp/defconfig` to `linux-imx/imx8-nxp-bsp/defconfig`? Or better, can we replace both with `linux-imx/aarch64/defconfig`?

otavio: Using `aarch64` seems like a good option.

conf/machine/include/imx-base.inc
thochstein: FYI, I started to suggest that we use `mx9` to denote the family as we do for `mx8m` and `mx8x`, but I remembered that the naming scheme has changed and `mx8m` is analogous to `mx93` or `mx95`. This is an improvement on the naming since for i.MX 8 we had the ambiguity of an i.MX 8 Series and an i.MX 8 Family.

recipes-bsp/firmware-imx/firmware-imx-9_8.18.bb
angolini: I'm currently testing this

conf/machine/include/imx-base.inc
zandrey: @angolini 

There is care to be taken here, as changing `DEFAULTTUNE` to plain `cortexa55` here would not imply that Advanced SIMD instructions would be enabled as well... This was enabled implicitly by `-crypto` suffix for `imx8m` series, but as far as I can see this suffix is not (yet) supported in _OE_Core_ for _A55_ profile.

I'm not sure if it's a miss in _OE-Core_ or it is not brought up deliberately, but without it - performance might suffer as ASIMD instructions would not be emitted by compiler.

@thochstein @otavio I'm a bit torn here: from one side - the new tune matches to what SOC has in the Cortex-A complex, but on the other hand - resulting compiled binaries might suffer performance or size difference if it is enabled like this.

Shall it be taken as it is presented in this PR, or perhaps first `-crypto` suffix shall be submitted to _OE-Core_?

I'm not sure: if `imx93` implements full _A55_ profile(@thochstein ❔) - then `-crypto` suffix would really be desired; if not - then this change is good to go.

-- andrey

thochstein: I will have to ask. We have been setting `DEFAULTTUNE:mx9-nxp-bsp    ?= "armv8a-crc-crypto"` in our distros, and I see now that we neglected to add an optimized tuning from the BSP layer.

otavio: I'd add it for our tune, so it uses by default @zandrey.

angolini: so far, I will let it be. Waiting to see if @thochstein brings some news

recipes-bsp/firmware-imx/firmware-imx-9_8.18.bb
angolini: I got it working somehow, I will update the code so you can review. After review I can test other SoCs

conf/machine/include/imx-base.inc
junzhuimx, line 382: firmware-imx-9 contains for DDR firmware for i.MX93. 
it is not right to use firmware-imx-8

recipes-bsp/firmware-imx/firmware-imx-9_8.18.bb
thochstein: @otavio, wouldn't it be better to combine all these different `firmware-imx` recipes since they are all using the same `SRC_URI`?

zandrey: > @zandrey it is no the copy of firmware-imx-8m. recently, firmware-imx-9 is for i.MX 9 series firmware only.


From the recipe content perspective, it is a copy.

I tend to agree with @thochstein comment above to combine `firmware-imx` into one recipe, since `8m` and `9` in this context point to the same `SRC_URI`.

conf/machine/include/imx-base.inc
zandrey: @thochstein

Good catch! Yes, indeed it is!

It is actually coming from the top include, and shall be enabled for `cortexa55`.

I would suggest to set `DEFAULTTUNE` straight to `cortexa55` here.

thochstein: @zandrey, great! I've also received confirmation that i.MX 93 implements full A55 profile.

angolini: great! So I'm resolving this as we all agreed (and my current version is already fixed)

recipes-kernel/linux/linux-fslc-imx/mx9-nxp-bsp/defconfig
angolini: then using `aarch64` the `imx-nxp-bsp` is avoided which causes an error for all the machines, so I suggest we use `imx8-nxp-bsp` and `imx9-nxp-bsp`

conf/machine/include/imx-base.inc
angolini: I have a patch on that recipe to install the right files. Let me push my latest

recipes-kernel/linux/linux-fslc-imx/mx9-nxp-bsp/defconfig
thochstein: Can we change `imx-nxp-bsp` to `arm`?

recipes-bsp/firmware-imx/firmware-imx-8_8.18.bb
```diff
@@ -18,6 +18,6 @@ addtask deploy after do_install before do_build
 
 PACKAGE_ARCH = "${MACHINE_SOCARCH}"
 
-COMPATIBLE_MACHINE      = "(mx8-nxp-bsp)"
+COMPATIBLE_MACHINE      = "(mx8-nxp-bsp|mx9-nxp-bsp)"
```
thochstein, line 21: With `firmware-imx-9` back, this change needs to be dropped as well.

angolini:
@otavio @thochstein @zandrey 
The changes are covered. The missing points are:

* for linux-imx/linux-fslc-imx what should be the directory structure?
currently we have (after this PR):
```
recipes-kernel/linux/linux-fslc
├── armv8a
├── defconfig
└── mxs-generic-bsp
recipes-kernel/linux/linux-fslc-imx
├── imx-nxp-bsp
├── mx8-nxp-bsp
└── mx93-nxp-bsp
recipes-kernel/linux/linux-fslc-lts
├── armv8a
├── defconfig
└── mxs-generic-bsp
recipes-kernel/linux/linux-imx
├── imx-nxp-bsp
├── mx8-nxp-bsp
└── mx93-nxp-bsp
```
* firmware-imx* should be merged?
currently after this PR we have:
```
./recipes-bsp/firmware-imx/firmware-imx-8_8.18.bb
./recipes-bsp/firmware-imx/firmware-imx_8.18.bb
./recipes-bsp/firmware-imx/firmware-imx-8m_8.18.bb
./recipes-bsp/firmware-imx/firmware-imx-8.18.inc
```

recipes-bsp/firmware-imx/firmware-imx-8_8.18.bb
junzhuimx: @angolini @otavio 
firmware-imx-9 is integrated to firmware-imx-8 recipe. 
How about also merge firmware-imx-8m into formware-imx-8?

otavio: @junzhuimx, totally agree. See #1426.

angolini:
From my POV all the requests were covered (or are tracked in an issue)

PR updated with the latest working version

otavio (approved):

thochstein (approved):

PR #1418, opened 2023-02-08
u-boot-imx: Change the imx bsp override to generic bsp

mdrodrigo:
With the imx bsp override the u-boot-imx does
not compile imx-boot-tools. This causes an error
in imx-boot because it does not find imx-boot-tools.

Signed-off-by: Rodrigo M. Duarte <rodrigo.duarte@ossystems.com.br>

### DIFF
--- a/recipes-bsp/u-boot/u-boot-imx_2022.04.bb
+++ b/recipes-bsp/u-boot/u-boot-imx_2022.04.bb
@@ -10,13 +10,13 @@ PROVIDES += "u-boot"
 inherit uuu_bootloader_tag
 
 UUU_BOOTLOADER            = ""
-UUU_BOOTLOADER:mx6-nxp-bsp        = "${UBOOT_BINARY}"
-UUU_BOOTLOADER:mx7-nxp-bsp        = "${UBOOT_BINARY}"
+UUU_BOOTLOADER:mx6-generic-bsp        = "${UBOOT_BINARY}"
+UUU_BOOTLOADER:mx7-generic-bsp        = "${UBOOT_BINARY}"
 UUU_BOOTLOADER_TAGGED     = ""
-UUU_BOOTLOADER_TAGGED:mx6-nxp-bsp = "u-boot-tagged.${UBOOT_SUFFIX}"
-UUU_BOOTLOADER_TAGGED:mx7-nxp-bsp = "u-boot-tagged.${UBOOT_SUFFIX}"
+UUU_BOOTLOADER_TAGGED:mx6-generic-bsp = "u-boot-tagged.${UBOOT_SUFFIX}"
+UUU_BOOTLOADER_TAGGED:mx7-generic-bsp = "u-boot-tagged.${UBOOT_SUFFIX}"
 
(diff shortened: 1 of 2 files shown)


recipes-bsp/u-boot/u-boot-imx_2022.04.bb
```diff
@@ -16,7 +16,7 @@ UUU_BOOTLOADER_TAGGED     = ""
 UUU_BOOTLOADER_TAGGED:mx6-nxp-bsp = "u-boot-tagged.${UBOOT_SUFFIX}"
 UUU_BOOTLOADER_TAGGED:mx7-nxp-bsp = "u-boot-tagged.${UBOOT_SUFFIX}"
 
-do_deploy:append:mx8m-nxp-bsp() {
+do_deploy:append() {
```
otavio, line 19: The right override to use is `mx8m-generic-bsp`.  All those need to be `generic-bsp` as the bootloader can also be used with mainline BSP.

mdrodrigo: FIxed

otavio: Please change all.

otavio (approved):

recipes-bsp/u-boot/u-boot-imx_2022.04.bb
mdrodrigo: Changed at all

thochstein (commented):
I assume that you are using `mainline-bsp`, since otherwise the `nxp-bsp` overrides would be working. I think the commit message should state this.

I'm concerned that we have other places in the layer where we are using `nxp-bsp` that should be `generic-bsp`. One obvious issue is with a recent change to `imx-boot`:

https://github.com/Freescale/meta-freescale/blame/master/recipes-bsp/imx-mkimage/imx-boot_1.0.bb#L68

@angolini , I believe at one time recently you talked about providing documentation regarding the different usage modes of the layer, mainline vs. nxp vs. other?

angolini:
The doc update is in [1] @thochstein 
[1] https://github.com/Freescale/Documentation/commit/6d7d61376c8b03a2f3fde909e9473b1e7f0a14cb

I'm not sure README is the best place for that kind of info, but I'm up to discuss Documentation improvments.

angolini (approved):

thochstein (approved):

thochstein:
> The doc update is in [1] @thochstein [1] [Freescale/Documentation@6d7d613](https://github.com/Freescale/Documentation/commit/6d7d61376c8b03a2f3fde909e9473b1e7f0a14cb)
> 
> I'm not sure README is the best place for that kind of info, but I'm up to discuss Documentation improvments.

Thanks, Daiane, I'll have a look. A quick glance shows this is more detailed than what I was thinking of, so yes, it doesn't go in a README. What I was thinking of was a more basic usage guide for a user of the layer, someone wanting to create an image and needing to understand the choices available in the layer and how they might choose from the different configuration options.

otavio (approved):

PR #1432, opened 2023-02-17
Update opencv and wayland-protocols for i.MX 9

thochstein:


recipes-support/opencv/opencv_4.6.0.imx.bb
```diff
@@ -276,14 +276,17 @@ SRC_URI += " \
...
 
-PACKAGECONFIG:remove        = "eigen"
-PACKAGECONFIG:append:mx8-nxp-bsp    = " dnn text"
-PACKAGECONFIG_OPENCL        = ""
+PACKAGECONFIG:remove = "eigen"
+
+PACKAGECONFIG:append:mx8-nxp-bsp = " dnn text"
+PACKAGECONFIG:append:mx9-nxp-bsp = " dnn text"
+
+PACKAGECONFIG:append = " ${PACKAGECONFIG_OPENCL}"
```
otavio, line 284: Please align this properly.

otavio (changes requested):

recipes-support/opencv/opencv_4.6.0.imx.bb
thochstein: I actually like it this way because the first line is easier to read, and you can see that it is a simple assignment to `PACKAGECONFIG_OPENCL`, which is aligned itself below the first line.

otavio (approved):

PR #1444, opened 2023-03-06
add 1.20.3 gstreamer1.0-rtsp-server recipe

kraj:


otavio:
@kraj we merged #1445 as it was more complete. Please check.

kraj:
seems good.

PR #1445, opened 2023-03-08
gstreamer: add recipes for upstream 1.20.3 versions

sapiippo:
oe-core has moved to gstreamer 1.22.0 and recipes which don't have an imx version can no longer be built. Add older version of those components to be used with the gstreamer 1.20.3.imx.


otavio (approved):
@sapiippo thanks for preparing it.

quaresmajose:
These new gstreamer recipes should use `imx` version `1.20.3.imx` instead of `1.20.3` like all the others upstream forks used in the layer. Maybe it can cause some troubles for folks who intend to use the upstream version `1.20.3`

MaxKrummenacher:
@quaresmajose 
The recipes here are not NXP forks of the upstream project. However if you use the -imx version of gstreamer these provide the matching older 1.20.3 version.

quaresmajose:
This is not a big troubles because the `1.20.3` version is not available anymore in oe-core master but is more safe IMO to have this recipes with the version named `1.20.3.imx`.

otavio:
I see your point of view, @quaresmajose. We could add a comment inside the recipe explaining the motivation for it. Can you prepare a  patch renaming them?

quaresmajose:
Sure, renamed and added a comment on https://github.com/Freescale/meta-freescale/pull/1448

PR #1453, opened 2023-03-13
xdg-desktop-portal-wlr: Disable build when using imx gpu driver

kraj:


otavio (approved):

otavio (approved):

thochstein:
@otavio, this does break `langdale` as the recipe does not exist there:

```
ERROR: No recipes in default available for:
  sources/meta-freescale/dynamic-layers/multimedia-layer/recipes-support/xdg-desktop-portal-wlr/xdg-desktop-portal-wlr_%.bbappend
```

otavio:
Wait, it shouldn't have it used as it has the dynamic-layers entry. @kraj ?

kraj:
its a new recipe, so perhaps its doable. Armin will have to decide, regardless will be good to send a backport.

recipes-multimedia/gstreamer/gstreamer1.0-plugins-base_1.20.3.imx.bb
```diff
@@ -85,6 +85,14 @@ EXTRA_OEMESON += " \
...
 "
 
+# Remove this meta package when moving to gstreamer 1.22+, this is for making
+# gst packagegroups from master work with gstreamer 1.20.3.imx
+PACKAGES += "${PN}-videoconvertscale"
+
+ALLOW_EMPTY:${PN}-videoconvertscale = "1"
+
+RDEPENDS:${PN}-videoconvertscale = "${PN}-videoconvert ${PN}-videoscale"
+
```
thochstein, line 95: This change does also break our `langdale` build, but in a different way that I don't understand. We see a warning:

```
WARNING: preferred version 1.22.0.imx of gstreamer1.0-plugins-base not available (for item gstreamer1.0-plugins-base-videoconvertscale)
```

And finally an error:

```
ERROR: Multiple versions of gstreamer1.0-plugins-base are due to be built (/langdale/sources/meta-imx/meta-bsp/recipes-multimedia/gstreamer/gstreamer1.0-plugins-base_1.22.0.imx.bb /langdale/sources/meta-freescale/recipes-multimedia/gstreamer/gstreamer1.0-plugins-base_1.20.3.imx.bb). Only one version of a given PN should be built in any given build. You likely need to set PREFERRED_VERSION_gstreamer1.0-plugins-base to select the correct version or don't depend on multiple versions.
```

We are using new recipes based on 1.22 from `master`. I don't understand how dynamic packages works. Perhaps our failure now is because Yocto chooses this explicit package from 1.20.3.imx recipe instead of using a dynamic package from the preferred 1.22.0.imx recipe?

kraj: if you are using 1.22.x then it should already be providing `gstreamer1.0-plugins-base-videoconvertscale` and you should not need 1.20.3.imx recipe. Since I dont see 1.22.0.imx recipe I dont know whats going on. It seems to work ok with 1.22.0 recipes from oe-core in mix.

thochstein: Hmm, don't know, we copied the 1.22 recipe verbatim. Without your change 1.22.0.imx is picked fine, with your change it is marked with the warning as `not available`. I will use `BBMASK` to hide the `meta-freescale` recipes.

otavio: @thochstein you could add an empty package to allow it to apply.

thochstein: Thanks. Then we have a workaround in a 1.22.0.imx recipe for a workaround in a 1.20.3.imx recipe for the 1.22.0 recipe.

PR #1457, opened 2023-03-15
gstreamer1.0-plugins-base: set default videosink for IMXQM/IMXQXP

liuming50:
The default 'autovideosink' gives very bad performance when playing H264/H265 videos on IMXQM/IMXQXP machines. Let's choose 'imxvideoconvert_g2d ! autovideosink' as the default videosink, so the end users dont need manually pass 'imxvideoconvert_g2d' pipeline to gst-play-1.0.


otavio (commented):
@thochstein, please take a look.

I don't like this idea as it overrides users' choices. If we consider adding this we'd need to support the other SoCs as well.

I'd like to understand your point of view on this.

huhaihua:
Hi @liuming50 , please don't change gstreamer community playback tool, we have a gplay-1.0 just like the gst-play-1.0, but it will auto the right video sink. package imx-gst1.0-plugins contains this tool, you can have a check

otavio:
I prefer to not diverge here, hence I am closing it.

PR #1458, opened 2023-03-16
isp-imx: Update isp-imx_4.2.2.19.0.bb to commit a7ef904

gibsson:
- NXP decided to change the binaries without updating the version [1]
- Here is NXP changelog: [LF-8298] Fix intermittent build break "Error copying file ... isi_common.h".

[1] https://github.com/nxp-imx/meta-imx/commit/43c0642e

### DIFF
--- a/recipes-bsp/isp-imx/isp-imx_4.2.2.19.0.bb
+++ b/recipes-bsp/isp-imx/isp-imx_4.2.2.19.0.bb
@@ -1,4 +1,4 @@
-# Copyright 2020-2022 NXP
+# Copyright 2020-2023 NXP
 
 DESCRIPTION = "i.MX Verisilicon Software ISP"
 LICENSE = "Proprietary"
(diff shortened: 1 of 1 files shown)


gibsson:
@thochstein how can updating binaries _after_ a release be authorized? Moreover there is a versioning number for that, why not name it 4.2.2.19.1?

jviguera:
The same happened in the Kirkstone branch. NXP updated the binaries without changing the version/revision:

https://github.com/nxp-imx/meta-imx/commit/ca4d220fb70af7a0d0b547216d18d94188f38dae

That said, @gibsson I would instead update to the latest version `isp-imx_4.2.2.20.0`, which seems to be the one used by NXP in their Kirkstone-based release.

gibsson:
@jviguera yes updating should be done but requires to update all the multimedia packages at once which I don't have time to do today. So this patch is to first fix the issue as-is.
Would you be willing to offer a PR to update all the multimedia packages to 5.15-2.2.0 release?

gibsson:
Any update on this? To be honest I think we should merge it (and backport to Kirkstone) as it simply breaks the build currently. But we can keep the comment section open.

thochstein:
Sorry for the trouble, but we did need to modify the binaries. We should have also upstreamed fixes here but forgot.

PR #1461, opened 2023-03-22
Allow generation of mfgtool bundle

otavio:
The class provides the infrastructure for MFGTOOL generation and is tied
to images. To generate the bundle, the task populate_mfgtool must be
called. For example:

,----[ Running populate_mfgtool for core-image-minimal image ]
| $: bitbake core-image-minimal -c populate_mfgtool
`----

The class behavior is controlled through the MFGTOOL_SCRIPT variable. The
MFGTOOL_SCRIPT variable itself specifies a space-separated list of the
scripts to enable. Following the scripts, you can determine the behavior
of each script by providing up to two order-dependent arguments, which
are separated by commas. You can omit any argument you like but must
retain the separating commas. The order is important and specifies the
following:

  1. Extra dependencies that should be added to the do_populate_mfgtool
     task, if the script is enabled.
  2. Extra binaries that should be added to the bundle, if the script is
     enabled.

For example:

,----[ Defining bootloader.uuu.in script ]
| MFGTOOL_SCRIPT = "bootloader.uuu.in"
| MFGTOOL_SCRIPT[bootloader.uuu.in] = "virtual/bootloader:do_deploy,${UBOOT_BINARY}"
`----

The virtual/bootloader:do_deploy is added to do_populate_mfgtool
dependencies and ${UBOOT_BINARY} copied to the bundle, only if the
script is enabled.

During the mfgtool bundle generation, the uuu.in files are processed and
some variables replaced. The variables are:

  - MACHINE
  - UBOOT_BINARY
  - SPL_BINARY
  - IMAGE_BASENAME

---

TODO Items before merging:
- [ ] add an i.MX8MM uuu script set
- [ ] add an i.MX6QDL uuu script set

### DIFF
--- a/classes/image_populate_mfgtool.bbclass
+++ b/classes/image_populate_mfgtool.bbclass
@@ -1,0 +1,139 @@
+# Allow generation of mfgtool bundle
+#
+# The class provides the infrastructure for MFGTOOL generation and is tied to images. To generate
+# the bundle, the task populate_mfgtool must be called. For example:
+#
+# ,----[ Running populate_mfgtool for core-image-minimal image ]
+# | $: bitbake core-image-minimal -c populate_mfgtool
+# `----
+#
+# The class behavior is controlled through the MFGTOOLCONFIG (analogous to PACKAGECONFIG)
+# variable. The MFGTOOLCONFIG variable itself specifies a space-separated list of the script to
+# enable. Following the script, you can determine the behavior of each script by providing up to two
+# order-dependent arguments, which are separated by commas. You can omit any argument you like but
+# must retain the separating commas. The order is important and specifies the following:
+#
+#   1. Extra dependencies that should be added to the do_populate_mfgtool task, if the script is
+#      enabled.
+#   2. Extra binaries that should be added to the bundle, if the script is enabled.
+#
+# For example:
+#
+# ,----[ Defining foo.uuu.in and bar.uuu script ]
+# | MFGTOOLCONFIG = "foo.uuu.in bar.uuu"
+# | MFGTOOLCONFIG[foo.uuu.in] = "dep-foo1:do_deploy dep-foo2:do_deploy,file-foo1"
+# | MFGTOOLCONFIG[bar.uuu] = "dep-bar1:do_deploy,file-bar1 file-bar2"
+# `----
+#
+# The dep-foo1:do_deploy, dep-foo2:do_deploy, and dep-bar1:do_deploy are added to
+# do_populate_mfgtool dependencies. In addition, file-foo1, file-bar1, and file-bar2 are copied to
+# the bundle, only if the script is enabled.
+#
+# During the mfgtool bundle generation, the uuu.in files are processed and some variables
+# replaced. The variables are:
+#
+#   - MACHINE
+#   - UBOOT_BINARY
+#   - SPL_BINARY
+#   - IMAGE_BASENAME
+#
+# Copyright 2022-2023 (C) O.S. Systems Software LTDA.
+#
+# SPDX-License-Identifier: MIT
+
+MFGTOOL_FILESPATH ??= " \
+    ${@base_set_filespath(["%s/mfgtool" % p for p in "${BBPATH}".split(":")] \
+                             + ["${FILE_DIRNAME}/${BP}/mfgtool", \
+                                "${FILE_DIRNAME}/${BPN}/mfgtool", \
+                                "${FILE_DIRNAME}/files/mfgtool"] \
+                          , d)} \
+"
+
+MFGTOOLDIR = "${WORKDIR}/mfgtool-${PN}"
+do_populate_mfgtool[dirs] = "${MFGTOOLDIR}"
+do_populate_mfgtool[cleandirs] = "${MFGTOOLDIR}"
+
+addtask populate_mfgtool after do_image_complete do_unpack before do_deploy
+do_populate_mfgtool[dirs] ?= "${DEPLOY_DIR_IMAGE} ${WORKDIR}"
+do_populate_mfgtool[nostamp] = "1"
+do_populate_mfgtool[recrdeptask] += "do_deploy"
+do_populate_mfgtool[depends] += "uuu-bin:do_populate_sysroot"
+
+python () {
+    depends = []
+    deploy_files = []
+    scripts = (d.getVar('MFGTOOLCONFIG') or "").split()
+    scripts_and_flags = d.getVarFlags('MFGTOOLCONFIG') or {}
+    for flag, flagval in sorted(scripts_and_flags.items()):
+        items = flagval.split(",")
+        num = len(items)
+        if num > 2:
+            bb.error("%s: MFGTOOLCONFIG[%s] Only \"depends,deploy files\" can be specified!" % (d.getVar("PN"), flag))
+
+        if flag in scripts:
+            if num >= 2 and items[1]:
+                deploy_files.append(items[1])
+            if num >= 1 and items[0]:
+                depends.append(items[0])
+
+    d.appendVarFlag('do_populate_mfgtool', 'depends', ' ' + ' '.join(depends))
+    d.setVar('_SCRIPT_DEPLOY_FILES', ' '.join(deploy_files))
+}
+
+python do_populate_mfgtool() {
+    # For MFGTOOLCONFIG items we use BitBake's fetcher module allowing a consistent behavior.
+    scripts = (d.getVar('MFGTOOLCONFIG') or "").split()
+    src_uri = ["file://%s" % f for f in scripts]
+    if not src_uri:
+        bb.fatal("MFGTOOLCONFIG is empty so populate_mfgtool cannot be run.")
+        return
+    bb.debug(1, "following scripts are used: %s" % ', '.join(scripts))
+
+    localdata = bb.data.createCopy(d)
+    filespath = (d.getVar('MFGTOOL_FILESPATH') or "")
+    localdata.setVar('FILESPATH', filespath)
+
+    try:
+        fetcher = bb.fetch2.Fetch(src_uri, localdata)
+        fetcher.unpack(localdata.getVar('WORKDIR'))
+    except bb.fetch2.BBFetchException as e:
+        bb.fatal("BitBake Fetcher Error: " + repr(e))
+
+    # Generate MFGTOOL bundle.
+    bb.build.exec_func('generate_mfgtool_bundle', d)
+}
+
+generate_mfgtool_bundle() {
+    bbnote "Processing uuu files ..."
+    for src in $(ls -1 ${WORKDIR}/*.uuu.in); do
+        dest=$(echo $src | sed 's,.in$,,g')
+        bbnote " - $src -> $dest"
+        sed -e 's/@@MACHINE@@/${MACHINE}/g' \
+            -e "s,@@UBOOT_BINARY@@,${UBOOT_BINARY},g" \
+            -e "s,@@SPL_BINARY@@,${SPL_BINARY},g" \
+            -e "s,@@IMAGE_BASENAME@@,${IMAGE_BASENAME},g" \
+            $src > $dest
+    done
+
+    bbnote "Deploying uuu files ..."
+    for src in $(ls -1 ${WORKDIR}/*.uuu); do
+        dest=$(basename $src)
+        bbnote " - $src -> ${MFGTOOLDIR}/${PN}-${MACHINE}/$dest"
+        install -D -m 0644 $src ${MFGTOOLDIR}/${PN}-${MACHINE}/$dest
+    done
+
+    bbnote "Copying uuu binaries..."
+    cp -v -s ${STAGING_LIBDIR}/uuu/* ${MFGTOOLDIR}/${PN}-${MACHINE}/
+
+    bbnote "Copying MFGTOOL extra deploy files..."
+    for f in ${_SCRIPT_DEPLOY_FILES}; do
+        mkdir -p ${MFGTOOLDIR}/${PN}-${MACHINE}/binaries
+        cp -v -s ${DEPLOY_DIR_IMAGE}/$f ${MFGTOOLDIR}/${PN}-${MACHINE}/binaries/
+    done
+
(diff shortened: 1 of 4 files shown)


classes/image_populate_mfgtool.bbclass
```diff
@@ -0,0 +1,142 @@
+# Allow generation of mfgtool bundle
```
angolini, line 1: there are 2 typos in the commit log:
tow/two
exemple/example

classes/image_populate_mfgtool.bbclass
```diff
@@ -0,0 +1,142 @@
...
+# The class provides the infrastructure for MFGTOOL generation and is tied to images. To generate
+# the bundle, the task populate_mfgtool must be called. For exemple:
+#
+# ,----[ Running populate_mfgtool for core-image-minimal image ]
+# | $: bitbake core-image-minimal -c populate_mfgtool
+# `----
+#
+# The class behavior is controlled through the MFGTOOLSCRIPT variable. The MFGTOOLSCRIPT variable
+# itself specifies a space-separated list of the scripts to enable. Following the scripts, you can
+# determine the behavior of each script by providing up to tow order-dependent arguments, which are
```
angolini, line 12: tow/two

classes/image_populate_mfgtool.bbclass
```diff
@@ -0,0 +1,142 @@
+# Allow generation of mfgtool bundle
+#
+# The class provides the infrastructure for MFGTOOL generation and is tied to images. To generate
+# the bundle, the task populate_mfgtool must be called. For exemple:
```
angolini, line 4: exemple/example

classes/image_populate_mfgtool.bbclass
```diff
@@ -0,0 +1,142 @@
...
+#
+# ,----[ Running populate_mfgtool for core-image-minimal image ]
+# | $: bitbake core-image-minimal -c populate_mfgtool
+# `----
+#
+# The class behavior is controlled through the MFGTOOLSCRIPT variable. The MFGTOOLSCRIPT variable
+# itself specifies a space-separated list of the scripts to enable. Following the scripts, you can
+# determine the behavior of each script by providing up to tow order-dependent arguments, which are
+# separated by commas. You can omit any argument you like but must retain the separating commas. The
```
angolini, line 13: the arguments are separated by commas. Maybe the list (for each argument) is separated by spaces?

angolini (commented):
I like this, let me test on my side to make sure I understand how to use

classes/image_populate_mfgtool.bbclass
otavio: The use of `,` is important as we can have multi dependencies and files.

otavio: I think I improved it.

classes/image_populate_mfgtool.bbclass
```diff
@@ -0,0 +1,142 @@
...
+
+MFGTOOL_FILESPATH ??= " \
+    ${@base_set_filespath(["%s/mfgtool" % p for p in "${BBPATH}".split(":")] \
+                             + ["${FILE_DIRNAME}/${BP}/mfgtool", \
+                                "${FILE_DIRNAME}/${BPN}/mfgtool", \
+                                "${FILE_DIRNAME}/files/mfgtool"] \
+                          , d)} \
+"
+
+MFGTOOLDIR = "${WORKDIR}/mfgtool-${PN}"
```
quaresmajose, line 50: The `${PN}` is already present on the `${WORKDIR}` of the image so maybe it can be dropped.

classes/image_populate_mfgtool.bbclass
```diff
@@ -0,0 +1,142 @@
...
+
+            if flag in mfgtoolscript:
+                if num >= 2 and items[1]:
+                    mfgtoolscript_deploy_files.append(items[1])
+                if num >= 1 and items[0]:
+                    mfgtoolscript_depends.append(items[0])
+
+        d.appendVarFlag('do_populate_mfgtool', 'depends', ' ' + ' '.join(mfgtoolscript_depends))
+
+    d.setVar('MFGTOOLSCRIPT_DEPLOY_FILES', ' '.join(mfgtoolscript_deploy_files))
```
quaresmajose, line 83: as `MFGTOOLSCRIPT_DEPLOY_FILES` is an internal variable of the class, something like `_MFGTOOLSCRIPT_DEPLOY_FILES` can be better.

quaresmajose (commented):
LGTM

classes/image_populate_mfgtool.bbclass
```diff
@@ -0,0 +1,142 @@
...
+#
+# The class provides the infrastructure for MFGTOOL generation and is tied to images. To generate
+# the bundle, the task populate_mfgtool must be called. For example:
+#
+# ,----[ Running populate_mfgtool for core-image-minimal image ]
+# | $: bitbake core-image-minimal -c populate_mfgtool
+# `----
+#
+# The class behavior is controlled through the MFGTOOLSCRIPT variable. The MFGTOOLSCRIPT variable
+# itself specifies a space-separated list of the scripts to enable. Following the scripts, you can
```
thochstein, line 11: To improve usability, I suggest updating the variable name with `_` and `S`: `MFGTOOL_SCRIPTS`.

classes/image_populate_mfgtool.bbclass
```diff
@@ -0,0 +1,142 @@
...
+    for src in $(ls -1 ${WORKDIR}/*.uuu); do
+        dest=$(basename $src)
+        bbnote " - $src -> ${MFGTOOLDIR}/${PN}-${MACHINE}/$dest"
+        install -D -m 0644 $src ${MFGTOOLDIR}/${PN}-${MACHINE}/$dest
+    done
+
+    bbnote "Copying uuu binaries..."
+    cp -v -s ${STAGING_LIBDIR_NATIVE}/uuu/* ${MFGTOOLDIR}/${PN}-${MACHINE}/
+
+    bbnote "Copyng MFGTOOL extra deploy files..."
```
MrCry0, line 131: `s/Copyng/Copying/`

With that and previous notes from other colleagues - LGTM.

classes/image_populate_mfgtool.bbclass
otavio: Fixed.

classes/image_populate_mfgtool.bbclass
otavio: Fixed.

classes/image_populate_mfgtool.bbclass
otavio: Fixed.

classes/image_populate_mfgtool.bbclass
```diff
@@ -0,0 +1,142 @@
...
+
+addtask populate_mfgtool after do_image_complete do_unpack before do_deploy
+do_populate_mfgtool[dirs] ?= "${DEPLOY_DIR_IMAGE} ${WORKDIR}"
+do_populate_mfgtool[nostamp] = "1"
+do_populate_mfgtool[recrdeptask] += "do_deploy"
+do_populate_mfgtool[depends] += "uuu-bin-native:do_populate_sysroot"
+
+python () {
+    # Process MFGTOOL_SCRIPTS and handle its fields.
+    mfgtoolscript_flags = d.getVarFlags('MFGTOOL_SCRIPTS') or {}
```
thochstein, line 62: It helps readability to maintain consistency, so `mfgtoolscript_flags` should be `mfgtool_scripts_flags`.

classes/image_populate_mfgtool.bbclass
```diff
@@ -0,0 +1,142 @@
...
+    mfgtoolscript_depends = []
+    mfgtoolscript_deploy_files = []
+    if mfgtoolscript_flags:
+        mfgtoolscript = (d.getVar('MFGTOOL_SCRIPTS') or "").split()
```
thochstein, line 66: `mfgtoolscript` -> `mfgtool_scripts`.

otavio: Fixed.

classes/image_populate_mfgtool.bbclass
otavio: Fixed.

otavio:
I ended up using singular for the variable name. The main reason for using singular is because every entry sets properties for a _specific_ script. So the dependencies and binaries for installation belong to one particular script.

thochstein:
The variable `MFGTOOL_SCRIPTS` represents a _set of scripts_ and so should be plural. The variable `MFGTOOL_SCRIPTS[script]` is itself singular, representing the configuration for the single member named `script`.

classes/image_populate_mfgtool.bbclass
```diff
@@ -0,0 +1,142 @@
...
+                    script_depends.append(items[0])
+
+        d.appendVarFlag('do_populate_mfgtool', 'depends', ' ' + ' '.join(script_depends))
+
+    d.setVar('_SCRIPT_DEPLOY_FILES', ' '.join(script_deploy_files))
+}
+
+python do_populate_mfgtool() {
+    # For MFGTOOL_SCRIPT items we use BitBake's fetcher module allowing a consistent behavior.
+    script = (d.getVar('MFGTOOL_SCRIPT') or "").split()
```
thochstein, line 88: Consider if you wanted to do a `for-loop`, best would be `for script in scripts`.

classes/image_populate_mfgtool.bbclass
```diff
@@ -0,0 +1,142 @@
...
+do_populate_mfgtool[dirs] ?= "${DEPLOY_DIR_IMAGE} ${WORKDIR}"
+do_populate_mfgtool[nostamp] = "1"
+do_populate_mfgtool[recrdeptask] += "do_deploy"
+do_populate_mfgtool[depends] += "uuu-bin-native:do_populate_sysroot"
+
+python () {
+    # Process MFGTOOL_SCRIPT and handle its fields.
+    script_flags = d.getVarFlags('MFGTOOL_SCRIPT') or {}
+    script_depends = []
+    script_deploy_files = []
```
thochstein, line 64: I see that this is where naming is tricky. The problem is `script_depends` and `script_deploy_files` here are sets of sets, so it's hard to come up with an accurate name that is not cumbersome. Although it could be argued that `script_depends` would be clearly a single set for a single `script`, while `scripts_depends` would be the full 'set of sets'.

How about using `flags`, `depends`, and `deploy_files` here, and `script_depends` and `script_deploy_files` for a single `script`?

classes/image_populate_mfgtool.bbclass
otavio: Sure, but the `MFGTOOL_SCRIPT` is analogous to `PACKAGECONFIG`. It defines a script to have flags (depends and files) defined.

thochstein: Let me try again: I consider that `PACKAGECONFIG` has a special definition, that a configuration is a set of values defining the singular configuration. I see no such definition for `MFGTOOL_SCRIPT`. It is not a set of values defining a singular script, so it is a bad analog.

thochstein: Hmm, how about using `MFGTOOLCONFIG`? I guess that makes your analog clearer to me.

otavio: `MFGTOOL_SCRIPT`: I consider it defines a script with a set of values. Consider this:

```
MFGTOOL_SCRIPT = "foo.uuu.in bar.uuu"
MFGTOOL_SCRIPT[foo.uuu.in] = "dep-foo1:do_deploy dep-foo2:do_deploy,file-foo1 file-foo2"
MFGTOOL_SCRIPT[bar.uuu] = "dep-bar1:do_deploy dep-bar2:do_deploy,file-bar1 file-bar2"
```

This defines two scripts, and each has a specific set of values. A convention is to use the same variable (here `MFGTOOL_SCRIPT`) to look up the flags.

thochstein: The question is whether the variable name can be improved. Calling it `MFGTOOL_SCRIPT` gives a natural connotation that only one script is allowed. A knowing user will go to the docs to see if multiples are allowed. Making it plural makes it clear from the name.

As I said, I think all this goes away if you just make the variable `MFGTOOL_CONFIG`, aligning it with `PACKAGECONFIG` and `UBOOT_CONFIG`, so giving the user another clue about how it works just from the name.

thochstein (approved):

thochstein:
Thanks :-)

classes/image_populate_mfgtool.bbclass
otavio: Fixed.

MrCry0 (approved):

recipes-devtools/uuu/uuu-bin-native_1.4.243.bb
```diff
@@ -0,0 +1,37 @@
...
+    install -D -m 0755 ${WORKDIR}/uuu-${PV}_mac ${D}${libdir}/uuu/uuu_mac
+    install -D -m 0644 ${WORKDIR}/uuu-${PV}.exe ${D}${libdir}/uuu/uuu.exe
+}
+
+SYSROOT_DIRS_NATIVE = "${libdir}/uuu"
+
+# HACK! We are not aiming to run those binaries during the build but copy then for MFGTOOL
+# bundle. If make the binaries relocatable they will rely on OpenEmbedded native sysroot paths but
+# we want them to run at user machine.
+SYSROOT_PREPROCESS_FUNCS:remove = "relocatable_binaries_preprocess relocatable_native_pcfiles"
```
quaresmajose, line 37: @otavio maybe is better to use only the `do_deploy` with `INHIBIT_DEFAULT_DEPS = "1"` for this propose. with that, the `inherit native` can also be removed.

otavio: I changed the way it works, and it looks good to me. Please take a look.

quaresmajose: even better deploy the archive with everything inside.

quaresmajose (approved):
LGTM

kraj (approved):
lgtm

PR #1465, opened 2023-03-29
Remove conditional imx-boot-container configuration from machines and move to extender

otavio:
This pull request includes two commits that make changes related to the `imx-boot-container` configuration.

The first commit removes the conditional `imx-boot-container` `MACHINEOVERRIDES` from machines. The conditional belongs to the bootloader recipe and can be removed from machines to simplify the configuration.

The second commit moves the `imx-boot-container` `MACHINEOVERRIDES` to extender as it is SoC specific and should not be set inside the machine. The commit moves it to the `imx-base.inc` file to ensure proper configuration.

Together, these changes improve the configuration of `imx-boot-container` in our software project by removing unnecessary conditional statements from machines and moving the configuration to a more appropriate location.

### DIFF
--- a/conf/machine/imx8mnul-ddr3l-evk.conf
+++ b/conf/machine/imx8mnul-ddr3l-evk.conf
@@ -4,7 +4,7 @@
 #@DESCRIPTION: Machine configuration for NXP i.MX 8M Nano UltraLite Evaluation Kit with DDR3L
 #@MAINTAINER: Jun Zhu <junzhu@nxp.com>
 
-MACHINEOVERRIDES =. "imx-boot-container:mx8mnul:"
+MACHINEOVERRIDES =. "mx8mnul:imx-boot-container:"
 
 require conf/machine/include/imx-base.inc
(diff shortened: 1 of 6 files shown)


thochstein:
I think it makes sense.

Note that the `imx-boot` recipe is used by NXP for all i.MX 8 families. Do you know why we limit `imx-boot-container` to 8M?

thochstein:
Actually, I think you should set `UBOOT_PROVIDES_BOOT_CONTAINER` with a weak default.

otavio:
> Actually, I think you should set `UBOOT_PROVIDES_BOOT_CONTAINER` with a weak default.

This variable can be removed as well. The class would need more rework, but I prefer to do that as another PR so we don't mix changes.

otavio:
About limiting it to i.MX8M SoCs, @zandrey did most of this work. But first, we must verify which platforms already use `binman` inside U-Boot.

Either way, we should first clean up this, and later, we can extend the functionality.

Issue #1469, opened 2023-03-31
Dunfell: CodeAurora links are broken


GabeDia:
Hello,

Per https://bye.codeaurora.org/ the NXP repositories have moved to their github counterpart and thus any recipes pointing to CodeAurora are broken (linux-imx, imx-atf, gst-*). Dunfell is supported until next year - can someone fix the affected recipes?

Thanks,
Gabe

joelsand:
I have the same issue 😞  Hope this can get fixed without too much trouble.

GabeDia:
No it needs to be ported to dunfell - thanks @thochstein for the quick fix.

PR #1470, opened 2023-03-31
Move from codeaurora to github

thochstein:
Fixes: #1469


otavio:
Some repo still needs to change the protocol. git protocol is not accepted on GitHub anymore.

otavio (approved):

PR #1478, opened 2023-04-03
Piglit OpenCL and GL

thochstein:


otavio (approved):

recipes-graphics/piglit/piglit_%.bbappend
```diff
@@ -1,4 +1,4 @@
-PACKAGECONFIG:remove:imxgpu = "x11 glx"
+PACKAGECONFIG:remove:imxgpu = "glx"
```
thochstein, line 1: @kraj, please note. You removed GL a few months ago, but it seems okay now.

kraj: what fixed it ?

thochstein: I don't know. We've been running with GL for a bit, and I only noticed this while working on an unrelated issue. Maybe something in the upstreaming has fixed it. I built 8QM MEK.

Can you check if it works for you now?

PR #1479, opened 2023-04-04
jailhouse: update to lf-5.15.71_2.2.0

ricardosalveti:
Update jailhouse to the lf-5.15.71_2.2.0 release and also extend the default compatible machine list to include 8ulp and mx93, since they are now supported on this release.


ricardosalveti:
Also another one that would be nice to backport to kirkstone once merged (to align on the bsp base version used there).

otavio (approved):

PR #1485, opened 2023-04-07
imx-base.inc: Fix 8M overrides

thochstein:
8MQ video is not working. The problem is traced to a change in `weston.ini`, specifically that the file is now from OE-Core layer instead of meta-freescale. This problem is traced to the move of the `imx-boot-container` override [1], specifically that the override was moved with a trailing `:` that inserts an empty override.
```
MACHINEOVERRIDES="aarch64:armv8a:use-nxp-bsp:imx-generic-bsp:imx-nxp-bsp:imxdrm:imxvpu:imxgpu:imxgpu3d:imxvulkan:mx8-generic-bsp:mx8-nxp-bsp:mx8m-generic-bsp:mx8m-nxp-bsp:mx8mq-generic-bsp:mx8mq-nxp-bsp:imx-boot-container::imx8mq-evk"
```

[1] https://github.com/Freescale/meta-freescale/commit/0ee4cb24eba5171da75ab8eeb72c7eab06083339


thochstein:
It's good, you didn't backport the original `imx-boot-container` rework to `kirkstone`, so nothing to fix.

PR #1493, opened 2023-04-12
linux-*-mfgtool, u-boot-*-mfgtool: Remove specific recipes

otavio:
The use of a specific recipes are not need anymore.

For the U-Boot specific case, the UBOOT_CONFIG mechanism allows multiple
binaries using different configurations so we can remove this recipes
and default to the IMX_DEFAULT_BOOTLOADER as provider.

The Linux specific case is completely unnecessary as the USB gadget is
supported in all versions we support.


kraj (approved):
LGTM

quaresmajose (approved):
LGTM

Issue #1290, opened 2022-11-10
imx6qdlsabresd: UBOOT_CONFIG not defined if kernel-fitimage is in KERNEL_CLASSES


sebfabre:
Hello,

I am working on imx6qdlsabresd with kirkstone branch (I think we have the same issue with imx6qdlsabreauto). If kernel-fitimage is added in KERNEL_CLASSES, I have the following build errors:
_Nothing PROVIDES ‘virtual/kernel’
....
Either UBOOT_MACHINE or UBOOT_CONFIG must be set in the imx6qdlsabresd machine configuration._

In imx6qdlsabresd.conf, UBOOT_CONFIG is only defined for uboot recipes:
_UBOOT_CONFIG:pn-u-boot-fslc ??= "sd"
UBOOT_CONFIG:pn-u-boot-imx ??= " \
    sd \
    ${@bb.utils.contains('MACHINE_FEATURES', 'optee', 'sd-optee', '', d)} \
"_
kernel-fitimage class inherits uboot-sign which inherits uboot-config. uboot-config class checks UBOOT_CONFIG and it is not defined.

https://github.com/Freescale/meta-freescale/blob/kirkstone/recipes-bsp/u-boot/libubootenv_%25.bbappend seems to manage this case but only for libubootenv.

zandrey:
I actually stumbled upon the same issue recently, but in a bit the different context.

I guess this has to be addressed in _OE-Core_ and not in this layer, as this dependency chain is hard-coded there.

-- andrey

PR #1494, opened 2023-04-13
imx6qdlsabre*: ensure UBOOT_CONFIG setting is always available

otavio:
We cannot tie the UBOOT_CONFIG setting to the U-Boot recipe otherwise if
kernel-fitimage is added in KERNEL_CLASSES, the following build errors
is raised:

,----
| Nothing PROVIDES ‘virtual/kernel’
| ...
| Either UBOOT_MACHINE or UBOOT_CONFIG must be set in the imx6qdlsabresd machine configuration.
`----

kernel-fitimage class inherits uboot-sign which inherits
uboot-config. uboot-config class checks UBOOT_CONFIG and it is not
defined.

The fix is essentially to move the conditional for the variable
assignment instead of using the recipe override.

Fixes: #1290

### DIFF
--- a/conf/machine/imx6qdlsabreauto.conf
+++ b/conf/machine/imx6qdlsabreauto.conf
@@ -32,17 +32,17 @@ KERNEL_DEVICETREE:use-mainline-bsp = " \
     imx6dl-sabreauto.dtb \
 "
 
+UBOOT_CONFIG ??= " \
+    sd \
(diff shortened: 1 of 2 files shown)


thochstein (approved):

kraj (approved):
lgtm

quaresmajose (approved):
LGTM

PR #1499, opened 2023-04-13
classes/image_populate_mfgtool: avoid error copying duplicated files

otavio:
Eventually, multiple `uuu` script may install same binaries. We need to
ensure we install them once, otherwise it'll fail as it might try to
override a existing binary.

This commit ensure all dependencies and binaries are mentioned once,
removing duplication.

Fixes: ae7af6f8 ("classes/image_populate_mfgtool: init")


angolini (approved):
I tested on my side and it worked.

PR #1510, opened 2023-04-20
Fix firmware-nxp-wifi update

thochstein:


recipes-bsp/firmware-imx/firmware-nxp-wifi_1.0.bb
```diff
@@ -28,11 +28,16 @@ do_compile() {
...
     # - bcm4359-pcie
-    install -D -m 0644 cyw-wifi-bt/*_CYW*/brcmfmac4359-pcie* ${D}${nonarch_base_libdir}/firmware/brcm
-    install -m 0644 cyw-wifi-bt/*_CYW*/BCM4349B1*.hcd ${D}${sysconfdir}/firmware
+    for f in cyw-wifi-bt/*_CYW*/brcmfmac4359-pcie*; do
+        install -D -m 0644 $f ${D}${nonarch_base_libdir}/firmware/brcm/$(basename $f)
+    done
+
+    for f in cyw-wifi-bt/*_CYW*/BCM4349B1*.hcd; do
+        install -D -m 0644 $f ${D}${sysconfdir}/firmware/$(basename $f)
+    done
```
thochstein, line 37: I have to say I am not a fan of using `install -D`. I've spent too much time debugging problems with it due to its non-standard handling of the destination. My preference would be to disallow it except for trivial cases and just make a separate line creating the directory. I mean, just look at the extra logic I had to write to make something sane out of this.

MrCry0 (approved):
LGTM.  Thanks!

Yeah, I faced problems with `install -D` in another project. It seems not reliable in some situations :(

Issue #1427, opened 2023-02-13
Rework linux kernel recipes to use in-tree `defconfig`


otavio:
In the past, we preferred to use the `defconfig` inside the layer, so it served as a reference for users. However, a lot of time passed, and users are now more used to using configuration fragments and Linux kernel forks, so it seems to be time to make it easier for us, maintainers, to manage the kernel and to use in-tree `defconfig`.

PR #1513, opened 2023-04-24
Remove SoC-specific defconfig, linux-fslc.inc and linux-fslc-lts 5.15 recipe

otavio:
This PR includes three commits that remove unnecessary files and configurations from the linux-fslc repository.

The first commit removes the SoC-specific defconfig and replaces it with the in-tree defconfig. The defconfig inside the layer served as a reference for users, but now it's easier for maintainers to manage the kernel using the in-tree defconfig. This commit resolves issue #1427.

The second commit removes the `linux-fslc.inc` file. This file was dead simple, and splitting it just added complexity, so it was removed.

The third commit removes the `linux-fslc-lts` 5.15 recipe since the repository moved to version 6.1, and the recipe is untested.


quaresmajose (approved):

MrCry0 (approved):
LGTM

PR #1529, opened 2023-05-11
dynamic-layers: meta-arm: don't force build nor install target in TF-A

clementperon:
Some board or user could like to change the build or install target of the ARM trusted firmware.

Only set those variables if they are empty.

### DIFF
--- a/dynamic-layers/meta-arm/recipes-bsp/trusted-firmware-a/trusted-firmware-a_%.bbappend
+++ b/dynamic-layers/meta-arm/recipes-bsp/trusted-firmware-a/trusted-firmware-a_%.bbappend
@@ -1,16 +1,20 @@
-# Common Build targets
-TFA_BUILD_TARGET = "all"
-TFA_INSTALL_TARGET = "bl31"
-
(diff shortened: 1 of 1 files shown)


dynamic-layers/meta-arm/recipes-bsp/trusted-firmware-a/trusted-firmware-a_%.bbappend
```diff
@@ -1,6 +1,6 @@
 # Common Build targets
-TFA_BUILD_TARGET = "all"
-TFA_INSTALL_TARGET = "bl31"
+TFA_BUILD_TARGET ?= "all"
+TFA_INSTALL_TARGET ?= "bl31"
```
otavio, line 3: I agree with you, but I would like to suggest overriding it only for the machines supported on the `bbappend`.

otavio (changes requested):

otavio (approved):

clementperon:
@otavio FYI i'm using kirkstone if it's still maintained would be glad if it's backported, thanks!

PR #1533, opened 2023-05-15
imx-boot-firmware-files: Simplify logic

thochstein:
Use variable overrides instead of function overriding.


otavio:
At the moment, I am inclined to decline this pull request. It is important for us to consider the possibility of supporting secure boot in the future, which may require generating files through programming methods.

thochstein:
My suggestion would be to keep DDR and Cadence with this simpler mechanism. Using the more complex function overriding does duplicate logic and doesn't seem to help.

If a more complex mechanism is needed for secure boot, we may find that it is simpler to add that specifically for secure boot. It seems better to deal with that problem then versus trying to guess now.

angolini:
I don't have an strong opinion, but the simpler logic seems ok to me, and when we get to the point to support secure boot, we can revert this commit and come back, if this is the case

PR #1537, opened 2023-05-17
weston upgrade, xwayland compatibility update

thochstein:

PR #1538, opened 2023-05-17
weston-init

thochstein:


otavio (approved):

recipes-graphics/wayland/weston-init.bbappend
```diff
@@ -8,64 +8,72 @@ REQUIRED_DISTRO_FEATURES:remove = "${IMX_REQUIRED_DISTRO_FEATURES_REMOVE}"
...
+
+PACKAGECONFIG_USE_G2D                ?= ""
+PACKAGECONFIG_USE_G2D:imxgpu2d       ?= "use-g2d"
+PACKAGECONFIG_USE_G2D:mx8qm-nxp-bsp  ?= ""
+PACKAGECONFIG_USE_G2D:mx8qxp-nxp-bsp ?= ""
+PACKAGECONFIG_USE_G2D:mx8dx-nxp-bsp  ?= ""
+
+USE_G2D_VALUE             = "true"
+USE_G2D_VALUE:mx6-nxp-bsp = "1"
+USE_G2D_VALUE:mx7-nxp-bsp = "1"
```
angolini, line 42: Could you, please, elaborate why there is "true" and "1"?

recipes-graphics/wayland/weston-init.bbappend
```diff
@@ -8,64 +8,72 @@ REQUIRED_DISTRO_FEATURES:remove = "${IMX_REQUIRED_DISTRO_FEATURES_REMOVE}"
...
+    fi
+
+    if [ "${@bb.utils.contains('PACKAGECONFIG', 'size', 'yes', 'no', d)}" = "yes" ]; then
+        sed -i -e "/^\[shell\]/a size=${SIZE_VALUE}" ${D}${sysconfdir}/xdg/weston/weston.ini
+    fi
+
+    if [ "${@bb.utils.contains('PACKAGECONFIG', 'use-g2d', 'yes', 'no', d)}" = "yes" ]; then
+        sed -i -e "/^\[core\]/a use-g2d=${USE_G2D_VALUE}" ${D}${sysconfdir}/xdg/weston/weston.ini
+    else
+        sed -i -e "/^\[core\]/a #use-g2d=${USE_G2D_VALUE}" ${D}${sysconfdir}/xdg/weston/weston.ini
```
angolini, line 75: Are line 75 inserting a commented line? I'm not against it, but I would like to double-check if I understand why.

angolini (approved):
LGTM

recipes-graphics/wayland/weston-init.bbappend
MrCry0: > Are line 75 inserting a commented line? I'm not against it, but I would like to double-check if I understand why.

@angolini I think it's intentional: if use-g2d is not in PACKAGECONFIG, it is good to add the commented option as a reference.

thochstein: Correct.

recipes-graphics/wayland/weston-init.bbappend
thochstein: The option was updated to a boolean in the weston 11 fork. The option remains as an int for weston 10. I'll update the commit message.

otavio: We now use weston 11 everywhere so we could use `true`, no?

thochstein: No, weston 11 has fully dropped `fbdev`, and we chose to lock i.MX 6 and 7 on weston 10. See #1537.

otavio: My fault. Thx.

PR #1539, opened 2023-05-18
xserver-xorg: fix patch fuzz

dankm:
Reformat the patch so that the do_patch step stops complaining about patch fuzz.


kraj (approved):
lgtm

PR #1540, opened 2023-05-19
Add screen machine feature to all machine configurations

JohannesKauffmann:
This unbreaks `IMAGE_FEATURES:append = " splash"`, as since poky `46d287faa6b72778dbe7652cde71e5def0f94747`, the "splash" image feature depends on the "screen" machine feature. Since `MACHINE_FEATURES` is an unconditional set (` = `) in `imx-base.inc`, appending to `MACHINE_FEATURES` in `local.conf` is not possible.

### DIFF
--- a/conf/machine/imx23evk.conf
+++ b/conf/machine/imx23evk.conf
@@ -19,5 +19,3 @@ UBOOT_SUFFIX = "sb"
 UBOOT_MACHINE = "mx23evk_config"
 
 KERNEL_DEVICETREE = "imx23-evk.dtb"
-
-MACHINE_FEATURES = "usbgadget usbhost vfat touchscreen"
--- a/conf/machine/imx28evk.conf
+++ b/conf/machine/imx28evk.conf
@@ -25,8 +25,6 @@ KERNEL_DEVICETREE = "imx28-evk.dtb"
 
 SERIAL_CONSOLES = "115200;ttyAMA0"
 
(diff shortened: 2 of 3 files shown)


conf/machine/imx23evk.conf
```diff
@@ -20,4 +20,4 @@ UBOOT_MACHINE = "mx23evk_config"
 
 KERNEL_DEVICETREE = "imx23-evk.dtb"
 
-MACHINE_FEATURES = "usbgadget usbhost vfat touchscreen"
+MACHINE_FEATURES = "usbgadget usbhost vfat touchscreen screen"
```
otavio, line 23: Could you please remove the override and switch to using the `imx-base` one? Thank you.

otavio (changes requested):

conf/machine/imx23evk.conf
JohannesKauffmann: Since 1f8df81eaa8c22329d99a473c45b3c368a363c31, the difference is that imx23evk no longer has `alsa`; apparently it wasn't ready at the time. I don't know about the state of imx23evk audio support, is it okay to add it back now?

otavio: I think it makes sense.

JohannesKauffmann: Done

PR #1543, opened 2023-05-23
conf: set the list of required layers

tprrt:
LAYERDEPENDS Lists the layers on which this recipe depends.

meta-oe is required by the linux-fscl recipe because it depends of lzop-native.

### DIFF
--- a/conf/layer.conf
+++ b/conf/layer.conf
@@ -9,6 +9,7 @@ BBFILE_COLLECTIONS += "freescale-layer"
 BBFILE_PATTERN_freescale-layer := "^${LAYERDIR}/"
 BBFILE_PRIORITY_freescale-layer = "5"
 LAYERSERIES_COMPAT_freescale-layer = "langdale mickledore"
+LAYERDEPENDS_freescale-layer = "core"
 
 # Add the Freescale-specific licenses into the metadata
 LICENSE_PATH += "${LAYERDIR}/custom-licenses"


conf/layer.conf
```diff
@@ -9,6 +9,7 @@ BBFILE_COLLECTIONS += "freescale-layer"
 BBFILE_PATTERN_freescale-layer := "^${LAYERDIR}/"
 BBFILE_PRIORITY_freescale-layer = "5"
 LAYERSERIES_COMPAT_freescale-layer = "langdale mickledore"
+LAYERDEPENDS_freescale-layer = "core openembedded-layer"
```
otavio, line 12: Hello! I was wondering if I could make a request regarding `linux-fslc`. Is it possible to take out the need for `lzop-native` and make `meta-oe` something that isn't required? If it's not too much trouble, could you create a PR for this? Thank you so much!

otavio (changes requested):

conf/layer.conf
otavio: Please check #1544.

conf/layer.conf
```diff
@@ -9,6 +9,7 @@ BBFILE_COLLECTIONS += "freescale-layer"
 BBFILE_PATTERN_freescale-layer := "^${LAYERDIR}/"
 BBFILE_PRIORITY_freescale-layer = "5"
 LAYERSERIES_COMPAT_freescale-layer = "langdale mickledore"
+LAYERDEPENDS_freescale-layer = "core"
```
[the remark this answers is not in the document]
tprrt: Done, sorry for forgetting.

PR #1544, opened 2023-05-23
Remove the need of lzop-native from layer

otavio:
There are three commits related to the Linux kernel configuration in the imx platform. This changes
remove the KERNEL_LZO configuration Linux kernel configuration and uses the KERNEL_GZIP, which is
the default. This is enough to remove the lzop-native dependency.

### DIFF
--- a/recipes-kernel/linux/linux-imx/ARM-imx_v7_defconfig-Remove-KERNEL_LZO-config.patch
+++ b/recipes-kernel/linux/linux-imx/ARM-imx_v7_defconfig-Remove-KERNEL_LZO-config.patch
@@ -1,0 +1,27 @@
+From ad9ab3b3c553cbc3c61f233b6e2cd5abdd2a624b Mon Sep 17 00:00:00 2001
+From: Otavio Salvador <otavio@ossystems.com.br>
+Date: Tue, 23 May 2023 13:16:05 -0300
+Subject: [PATCH] ARM: imx_v7_defconfig: Remove KERNEL_LZO config
+
+The KERNEL_GZIP is used in most config and is the default, there is no
+clear reason to diverge so let default be used.
+
+Upstream-Status: Pending
+
+Signed-off-by: Otavio Salvador <otavio@ossystems.com.br>
+---
+ arch/arm/configs/imx_v7_defconfig | 1 -
+ 1 file changed, 1 deletion(-)
+
+diff --git a/arch/arm/configs/imx_v7_defconfig b/arch/arm/configs/imx_v7_defconfig
+index 6de3049689191..94e9457e2bab1 100644
+--- a/arch/arm/configs/imx_v7_defconfig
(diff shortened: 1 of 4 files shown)


recipes-kernel/linux/linux-imx/ARM-imx_v7_defconfig-Remove-KERNEL_LZO-config.patch
```diff
@@ -0,0 +1,27 @@
...
+---
+ arch/arm/configs/imx_v7_defconfig | 1 -
+ 1 file changed, 1 deletion(-)
+
+diff --git a/arch/arm/configs/imx_v7_defconfig b/arch/arm/configs/imx_v7_defconfig
+index 6de3049689191..94e9457e2bab1 100644
+--- a/arch/arm/configs/imx_v7_defconfig
++++ b/arch/arm/configs/imx_v7_defconfig
+@@ -1,4 +1,3 @@
+-CONFIG_KERNEL_LZO=y
```
kraj, line 21: will this patch eventually show up in imx kernel too ?

kraj (approved):

recipes-kernel/linux/linux-imx/ARM-imx_v7_defconfig-Remove-KERNEL_LZO-config.patch
otavio: I hope so. @thochstein?

fabioestevam (commented):
@otavio I suggest you submit the same change to arch/arm/configs/imx_v6_v7_defconfig upstream as well.

CONFIG_KERNEL_LZO=y was introduced in 2011 by the following commit below:

commit 06965c39b4c63933fa0a1cde2237ef85477c5655
Author: Sascha Hauer <s.hauer@pengutronix.de>
Date:   Wed Aug 24 09:15:30 2011 +0200

    ARM i.MX5: update defconfig
    
    Rename it from mx51_defconfig to mx5_defconfig since the i.MX53
    is now supported aswell. Also, remove the DEBUG_LL and enable
    some more drivers useful for i.MX5 boards. For supporting
    1GiB of SDRAM vmsplit 2/2 is enabled.
    
    Signed-off-by: Sascha Hauer <s.hauer@pengutronix.de>

Only imx_v6_v7_defconfig and sdk7786_defconfig select CONFIG_KERNEL_LZO=y, so it seems fine to remove it as it looks like an old leftover.

otavio:
Sent. :-)

recipes-kernel/linux/linux-imx/ARM-imx_v7_defconfig-Remove-KERNEL_LZO-config.patch
thochstein: We have planned it for our Q3 release.

PR #1546, opened 2023-05-24
imx-gpu-viv: Move cl_viv_vx_ext.h out of libclc-imx-dev

thochstein:


otavio:
It is not. Check i.MX6 please.

thochstein:
It is working fine for me. Since we don't define `libclc-imx-dev`, by definition it is empty, so I don't understand your concern. There are headers in `libopencl-imx-dev`, if that's what you are thinking of:

```
build/tmp/work/cortexa9t2hf-neon-mx6qdl-fsl-linux-gnueabi/imx-gpu-viv/1_6.4.11.p1.0-aarch32-r0/packages-split$ tree libclc-imx* libopencl-imx*
libclc-imx
└── usr
    ├── include
    │   └── CL
    │       └── cl_viv_vx_ext.h
    └── lib
        └── libCLC.so
libopencl-imx
├── etc
│   └── OpenCL
│       └── vendors
│           └── Vivante.icd
└── usr
    └── lib
        ├── libOpenCL.so.1 -> libOpenCL.so.1.2.0
        ├── libOpenCL.so.1.2.0
        └── libOpenCL.so.3.0.0
libopencl-imx-dev
└── usr
    ├── include
    │   └── CL
    │       ├── cl2.hpp
    │       ├── cl_d3d10.h
    │       ├── cl_d3d11.h
    │       ├── cl_dx9_media_sharing.h
    │       ├── cl_dx9_media_sharing_intel.h
    │       ├── cl_egl.h
    │       ├── cl_ext.h
    │       ├── cl_ext_intel.h
    │       ├── cl_ext_viv.h
    │       ├── cl_gl_ext.h
    │       ├── cl_gl.h
    │       ├── cl.h
    │       ├── cl_half.h
    │       ├── cl.hpp
    │       ├── cl_icd.h
    │       ├── cl_layer.h
    │       ├── cl_platform.h
    │       ├── cl_va_api_media_sharing_intel.h
    │       ├── cl_version.h
    │       ├── opencl.h
    │       └── opencl.hpp
    └── lib
        └── libOpenCL.so -> libOpenCL.so.1.2.0
```

otavio:
Here:
```
libclc-imx
└── usr
    ├── include
    │   └── CL
    │       └── cl_viv_vx_ext.h
    └── lib
        └── libCLC.so
```
The `cl_viv_vx_ext.h` cannot be in this package.

thochstein:
It is needed at runtime.

thochstein:
Perhaps you're right, I know I went back and forth a bit with the dev team on this.

thochstein:
Oh, maybe the point is that `libclc-imx` is providing the CL compiler, so it doesn't make sense to include it in the rootfs without the header.

otavio:
The `.so` should be in `-dev` and not the other way around.

thochstein:
The `.so` is not versioned, hence needs to be in the main package.

thochstein:
Note that upstream does appear to [contain SOC-specifics in the main package](https://packages.debian.org/sid/all/libclc-15/filelist), and [only generics in `-dev`](https://packages.debian.org/sid/all/libclc-15-dev/filelist). So, keeping the original implementation would seem to be aligned.

otavio:
Thank you for sharing your thoughts. However, I must point out that the comparison is not entirely fair since the package in question is architecture-independent, while `libclc-imx` and `libclc-imx-dev` are not. 

The main concern here is whether the `cl_viv_vx_ext.h` header file is utilized during application runtime. Based on my understanding, it is not. Rather, it is mostly used during OpenCL development to enable applications to link with the `libclc.so` library (which should ideally be versioned, but that is a separate matter altogether). As such, it would make sense for the header to be included in `libclc-imx-dev`.

recipes-graphics/imx-gpu-viv/imx-gpu-viv-6.inc
```diff
@@ -328,7 +328,6 @@ do_install:append:libc-musl() {
 ALLOW_EMPTY:${PN} = "1"
 
 FILES:libclc-imx = "${libdir}/libCLC${SOLIBS}"
-FILES:libclc-imx-dev = "${includedir}/CL/cl_viv_vx_ext.h"
```
otavio, line 331: This is fine but we need a: `RPROVIDER:libopencl-imx-dev = "libclc-imx-dev"`

thochstein: Can you explain why? For backwards compatibility? I think it was wrong to put the header in libclc-imx to begin with. The recipe may have putting all the OpenCL headers in libclc-imx-dev as well (in fact it was trying to put the CL headers in other packages as well), which I guess was pretty wrong and we fixed without adding RPROVIDER.

otavio (approved):

otavio:
Please squash the commits and mark it as ready ;-)

PR #1554, opened 2023-05-31
uuu: drop bogus COMPATIBLE_MACHINE

Villemoes:
This apparently prevents the recipe from actually building for the most useful variants, native and nativesdk - it's really rare that one would actually want to run uuu itself _on_ an imx target.

Moreover, we have a test setup where we use Raspberry Pis for power cycling, exposing the serial console etc. from various boards, so we do want to build uuu for the BSPs we put on those RPis in order to automatically test bootstrapping of imx boards.

### DIFF
--- a/recipes-devtools/uuu/uuu_git.bb
+++ b/recipes-devtools/uuu/uuu_git.bb
@@ -15,6 +15,4 @@ S = "${WORKDIR}/git"
 
 DEPENDS = "libusb zlib bzip2 openssl"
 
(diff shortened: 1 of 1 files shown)


Villemoes:
Thanks for merging quickly.

Any chance this could be backported to the branches that are still active? And how does that work in practice, should I create separate backport-1554-to-kirkstone etc. branches and create separate MRs, or is there some automation that could help?

otavio:
Done 👍🏼

PR #1562, opened 2023-06-06
weston-init: disable xwayland for mainline BSPs

liuming50:
A regression was introduced by commit 47f0410be1:
[ weston-init: Rework weston.ini configuration ]

xwayland had been dropped from weston.ini for mainline BSPs before that commit, since it's not supported by mainline BSPs, but after that, it's added back again.

Introduce a xwayland PACKAGECONFIG to control that, drop it from weston.ini from mainline BSPs.

### DIFF
--- a/recipes-graphics/wayland/weston-init.bbappend
+++ b/recipes-graphics/wayland/weston-init.bbappend
@@ -10,11 +10,15 @@ SRC_URI:append:mx6sl-nxp-bsp = " file://weston.config"
 
 PACKAGECONFIG ??= " \
     no-idle-timeout \
+    ${@bb.utils.contains('DISTRO_FEATURES', 'x11', 'xwayland', '', d)} \
     ${PACKAGECONFIG_GBM_FORMAT} \
     ${PACKAGECONFIG_REPAINT_WINDOW} \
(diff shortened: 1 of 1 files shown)


MrCry0 (approved):

thochstein:
Looks good, my only request is to add back the data from the original commit so it's simpler to see:

```
The main recipe now sets xwayland=true in weston.ini, but xwayland does
not really work for mainline BSPs, the following error was observed on
both imx6/imx7 machines, as follows:

| # gst-play-1.0 testvideo_h264_360p.mp4
| Press 'k' to see a list of keyboard shortcuts.
| Now playing testvideo_h264_360p.mp4
| X connection to :0.0 broken (explicit kill or server shutdown).

after removing "xwayland=true" from weston.ini, it works again.

This patch mainly aims to resolve this problem by introducing a variable
INI_COMMENT_ASSIGNMENTS that the end users could set to comment a line in
weston.ini.
```

liuming50:
Changes in patch set V2:

1 Adjust git commit message according to @thochstein 's commnts.

otavio (approved):

thochstein (approved):

PR #1563, opened 2023-06-06
imx-gpu-viv: Add runtime dependency for OpenCL VX Intrinsic Extension

thochstein:
The OpenCL implementation includes an extension for VX Intrinsics. To use it, an app requires the CL compiler and /usr/include/CL/cl_viv_vx_ext.h, so add the appropriate runtime dependency to libopenvx-imx.

### DIFF
--- a/recipes-graphics/imx-gpu-viv/imx-gpu-viv-6.inc
+++ b/recipes-graphics/imx-gpu-viv/imx-gpu-viv-6.inc
@@ -188,10 +188,6 @@ IMX_SOC:mx8ulp-nxp-bsp = "mx8ulp"
 
 LIBVULKAN_API_VERSION   = "1.2.182"
 
-SUPPORTS_OPENCL_VX_INTRINSIC_EXTENSION               = "false"
-SUPPORTS_OPENCL_VX_INTRINSIC_EXTENSION:mx8qm-nxp-bsp = "true"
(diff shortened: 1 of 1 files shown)


otavio (approved):

otavio:
Thank you, you found the solution to be very nice!

thochstein:
Thanks! I'm still building and looking for internal feedback as well.

thochstein:
@otavio, this is ready.

kraj:
There is a problem this causes. Its crossing the runtime dependency boundary between a package and its development libs and headers package now. 
```
OPENCL_VX_INTRINSIC_EXTENSION_PACKAGES:mx8qm-nxp-bsp = "libclc-imx libopencl-imx-dev"
```
One can add `INSANE_SKIP` on `dev-deps` I think libopencl-imx-dev does not have recursive dependencies on other -dev packages which is a good sign, it only rdepends on libopencl-imx

I am testing a local patch and will post if it works.

thochstein:
Thanks, @kraj . Darn, I did assume I would have such a problem, but never saw it during builds. I did just do `bitbake imx-gpu-viv -c do_package_qa -f` and now see the problem. Perhaps my builds were somehow skipping that step? You'd think though that with the change in packaging that qa would be rerun.

kraj:
One practice I usually use is to test changes with `bitbake -ccleanall <recipe> && bitbake <recipe>`

PR #1566, opened 2023-06-08
Fixes with latest oe-core master changes and clang16

kraj:
setuptools3-base FILES assignments have been simplified, therefore the internal libraries should be now packaged explicitly.

Fixes
ERROR: QA Issue: gstreamer1.0-python: Files/directories were installed but not shipped in any package:
  /usr/lib/gstreamer-1.0/libgstpython.so


otavio (approved):

thochstein (changes requested):
The first commit header is wrong, the recipe is `gstreamer1.0-python`. Otherwise, the rest is good.

kraj:
> The first commit header is wrong, the recipe is `gstreamer1.0-python`. Otherwise, the rest is good.

good catch ! fixed thusly.

thochstein (approved):

PR #1567, opened 2023-06-09
meta-freescale: fmlib: map kernel source path to target path

limeng-linux:
When building package fmlib based on Yocto version mickledore, there is below warning reported:
WARNING: fmlib-git-r1 do_package_qa: QA Issue: File /usr/lib64/libfm-arm.a in package fmlib-staticdev contains reference to TMPDIR [buildpaths] Because the KERNEL_SRC is passed into Makefile, and not mapped to a target path like the source of this package. To void this warning, it also needs to add the path map in CFLAGS  for kernel source.


limeng-linux:
There  are 2 patches to fix the TMPDIR issue for fmlib and tsntool package.

thanks,
Limeng

otavio (approved):

PR #1568, opened 2023-06-09
add vulkan-validation-layers support

EasyIP2023:
PR fixes

https://errors.yoctoproject.org/Errors/Details/707981/

### DIFF
--- a/recipes-downgrade/vulkan/vulkan-validation-layers_1.2.182.0.imx.bb
+++ b/recipes-downgrade/vulkan/vulkan-validation-layers_1.2.182.0.imx.bb
@@ -1,0 +1,52 @@
+SUMMARY = "Vulkan Validation layers"
+DESCRIPTION = "Khronos official Vulkan validation layers to assist developers \
+in verifying that their applications correctly use the Vulkan API"
+HOMEPAGE = "https://www.khronos.org/vulkan/"
+BUGTRACKER = "https://github.com/KhronosGroup/Vulkan-ValidationLayers"
+SECTION = "libs"
+
+LICENSE = "Apache-2.0"
+LIC_FILES_CHKSUM = "file://LICENSE.txt;md5=8df9e8826734226d08cb412babfa599c"
+
+SRC_URI = "git://git@github.com/KhronosGroup/Vulkan-ValidationLayers.git;branch=sdk-1.2.182;protocol=https"
+SRCREV = "91fcffa8eca1a9573c8d736a54a028035ed4e06c"
+
+S = "${WORKDIR}/git"
+
+REQUIRED_DISTRO_FEATURES = "vulkan"
+
+DEPENDS = "vulkan-headers vulkan-loader spirv-headers spirv-tools glslang"
+
+# BUILD_TESTS            - Not required for OE builds
+# BUILD_WERROR           - There are too many compiler warnings/errors due to upgrades in version
+#                          of clang. Requiring a number of patches from upstream. Disable compiler
+#                          -Werror to bypass build issues.
+# USE_ROBIN_HOOD_HASHING - Provides substantial performance improvements on all platforms.
+#                          Yocto project doesn't contain a recipe for package so disabled it.
+EXTRA_OECMAKE = "\
+    -DBUILD_TESTS=OFF \
+    -DBUILD_WERROR=OFF \
+    -DUSE_ROBIN_HOOD_HASHING=OFF \
+    -DGLSLANG_INSTALL_DIR=${STAGING_LIBDIR} \
+    -DVULKAN_HEADERS_INSTALL_DIR=${STAGING_DATADIR} \
+    -DSPIRV_HEADERS_INSTALL_DIR=${STAGING_DATADIR} \
+    "
+
+PACKAGECONFIG[x11] = "-DBUILD_WSI_XLIB_SUPPORT=ON -DBUILD_WSI_XCB_SUPPORT=ON, -DBUILD_WSI_XLIB_SUPPORT=OFF -DBUILD_WSI_XCB_SUPPORT=OFF, libxcb libx11 libxrandr"
+PACKAGECONFIG[wayland] = "-DBUILD_WSI_WAYLAND_SUPPORT=ON, -DBUILD_WSI_WAYLAND_SUPPORT=OFF, wayland"
+
+PACKAGECONFIG ??= "${@bb.utils.filter('DISTRO_FEATURES', 'x11 wayland', d)}"
+
+inherit cmake features_check pkgconfig
+
+FILES:${PN} += "${datadir}/vulkan"
+
+SOLIBS = ".so"
+FILES_SOLIBSDEV = ""
+
+# These recipes need to be updated in lockstep with each other:
+# glslang, vulkan-headers, vulkan-loader, vulkan-tools,
+# vulkan-validation-layers, spirv-headers, spirv-tools
+# The tags versions should always be sdk-x.y.z, as this is what
+# upstream considers a release.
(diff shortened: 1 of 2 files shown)


thochstein:
We do already have this recipe here, with one less hyphen in the name:

https://github.com/Freescale/meta-freescale-distro/tree/master/recipes-graphics/vulkan-validationlayers

EasyIP2023:
@otavio and @thochstein Hey, I just recently merged the same recipe in OE-core.

Above error came up during @kraj testing.

Guess the question I have, do we want to stick to what's already implemented or have a recipe name compatible with OE-core recipe and remove from `meta-freescale-demo`?

thochstein:
We certainly want to be aligned. I wonder if OE-Core should be fixed so the recipe name does match the repo name?

thochstein:
Note that what is in `meta-freescale-distro` does have an important patch as well. I wonder if it should be upstreamed? Not sure of the details.

https://github.com/Freescale/meta-freescale-distro/blob/master/recipes-graphics/vulkan-validationlayers/vulkan-validationlayers/0001-CMakeLists.txt-Change-the-installation-path-of-JSON-.patch

EasyIP2023:
From what I see in the patch. `/etc/vulkan/explicit_layer.d` is used to store `VkLayer_khronos_validation.json` instead of `/usr/share/vulkan/explicit_layer.d`. As to why on `INSTALL_DEFINES` line they update `"library_path": "libVkLayer_khronos_validation.so"` to `"library_path": "/usr/lib/libVkLayer_khronos_validation.so"` is questionable. That shouldn't be required to run Vulkan code with the validation layers extension enabled.

I'd drop the patch as it seems more like developer installation preference. Also `VK_LAYER_PATH` path can be set during testing if need be, but vulkan should automatically search `/usr/share/vulkan/explicit_layer.d` and find layer json file.

I have a preference for the name `vulkan-validation-layers`.

thochstein:
As I remember, the patch was necessary to fix some build error(s), so if it's building for you, we're probably okay. We did more recently do some rework in imx-gpu-viv related to this, so perhaps the patch was necessary before the rework.

EasyIP2023:
Yeah, master builds for me with latest OE-core, meta-clang (using clang compiler), meta-openembedded, etc... 

@otavio Do you still want me to add `.imx` to recipe name? Would you like me to also rename the other recipes in `recipes-downgrade/vulkan` to include `.imx` in recipe name?

otavio:
Sure. It makes easier to find it out.

recipes-downgrade/vulkan/vulkan-validation-layers_1.2.182.0.bb
```diff
@@ -0,0 +1,52 @@
...
+    -DUSE_ROBIN_HOOD_HASHING=OFF \
+    -DGLSLANG_INSTALL_DIR=${STAGING_LIBDIR} \
+    -DVULKAN_HEADERS_INSTALL_DIR=${STAGING_DATADIR} \
+    -DSPIRV_HEADERS_INSTALL_DIR=${STAGING_DATADIR} \
+    "
+
+PACKAGECONFIG[x11] = "-DBUILD_WSI_XLIB_SUPPORT=ON -DBUILD_WSI_XCB_SUPPORT=ON, -DBUILD_WSI_XLIB_SUPPORT=OFF -DBUILD_WSI_XCB_SUPPORT=OFF, libxcb libx11 libxrandr"
+PACKAGECONFIG[wayland] = "-DBUILD_WSI_WAYLAND_SUPPORT=ON, -DBUILD_WSI_WAYLAND_SUPPORT=OFF, wayland"
+
+PACKAGECONFIG ?= "${@bb.utils.filter('DISTRO_FEATURES', 'x11 wayland', d)}"
```
thochstein, line 38: This is normally a weak default assignment `PACKAGECONFIG ??=`, and it goes above the individual feature definitions `PACKAGECONFIG[f1]`.

https://docs.yoctoproject.org/singleindex.html#term-PACKAGECONFIG

EasyIP2023: Changed

thochstein: Please fix the order as well:

```
PACKAGECONFIG ??= "${@bb.utils.filter('DISTRO_FEATURES', 'x11 wayland', d)}"

PACKAGECONFIG[x11] = "-DBUILD_WSI_XLIB_SUPPORT=ON -DBUILD_WSI_XCB_SUPPORT=ON, -DBUILD_WSI_XLIB_SUPPORT=OFF -DBUILD_WSI_XCB_SUPPORT=OFF, libxcb libx11 libxrandr"
PACKAGECONFIG[wayland] = "-DBUILD_WSI_WAYLAND_SUPPORT=ON, -DBUILD_WSI_WAYLAND_SUPPORT=OFF, wayland"
```

EasyIP2023: Fixed

thochstein (approved):

recipes-downgrade/vulkan/vulkan-validation-layers_1.2.182.0.imx.bb
```diff
@@ -0,0 +1,52 @@
+SUMMARY = "Vulkan Validation layers"
```
kraj, line 1: why do we need a copy of recipe ? if there are differences to oe-core one, then it can be dealt with in a bbappend.

EasyIP2023: Following standards currently in repo. Have to ask @thochstein @otavio, but I think it's to be explicit and make it obvious what version of Vulkan to utilize with i.MX GPU.

I'm up for utilizing a `vulkan-validation-layers_%.bbappend` and only updating PV/SRC_URI/SRCREV. Following design in layer.

EasyIP2023: Also think decision was made because we can't set an explicit version number in the name of the file with bbappend if a bb with the same version number doesn't exist.

otavio:
@EasyIP2023 rebase please.

thochstein:
> We do already have this recipe here, with one less hyphen in the name:
> 
> https://github.com/Freescale/meta-freescale-distro/tree/master/recipes-graphics/vulkan-validationlayers

I assumed you would remove this as well?

otavio:
Sure 👍🏼

thochstein:
This does fail in `do_configure`:

```
| CMake Error at /opt/work/upstream/fsl-xwayland/tmp/work/cortexa72-cortexa53-crypto-mx8-fsl-linux/vulkan-validation-layers/1.2.182.0.imx-r0/recipe-sysroot-native/usr/share/cmake-3.26/Modules/FindPackageHandleStandardArgs.cmake:230 (message):
|   Could NOT find VulkanHeaders (missing: VulkanHeaders_INCLUDE_DIR)
```

Updating `EXTRA_OECMAKE` to match the upstream recipe does fix the configure error:

```
-    -DGLSLANG_INSTALL_DIR=${STAGING_LIBDIR} \
-    -DVULKAN_HEADERS_INSTALL_DIR=${STAGING_DATADIR} \
-    -DSPIRV_HEADERS_INSTALL_DIR=${STAGING_DATADIR} \
+    -DGLSLANG_INSTALL_DIR=${STAGING_DATADIR} \
+    -DVULKAN_HEADERS_INSTALL_DIR=${STAGING_EXECPREFIXDIR} \
+    -DSPIRV_HEADERS_INSTALL_DIR=${STAGING_EXECPREFIXDIR} \
```

PR to follow.

PR #1573, opened 2023-06-13
vulkan-validation-layers: Fix do_configure failure

thochstein:
The build does fail in do_configure:
```
| CMake Error at /opt/work/upstream/fsl-xwayland/tmp/work/cortexa72-cortexa53-crypto-mx8-fsl-linux/vulkan-validation-layers/1.2.182.0.imx-r0/recipe-sysroot-native/usr/share/cmake-3.26/Modules/FindPackageHandleStandardArgs.cmake:230 (message):
|   Could NOT find VulkanHeaders (missing: VulkanHeaders_INCLUDE_DIR)
Updating EXTRA_OECMAKE to match the upstream recipe does fix the configure error
```

Checking the upstream recipe, there are three changes in `EXTRA_OECMAKE`. Aligning those does fix the failure.

Fixes: #1568


EasyIP2023:
@thochstein Thanks. I'll try to cancel some of the PR changes I made upstream as well. That's odd thought I saw the `VulkanHeadersConfig.cmake` be located in `<recipes-sysroot>/usr/share/cmake`.

PR #1576, opened 2023-06-16
fsl-kernel-localversion: fix regression setting LOCALVERSION

jviguera:
CONFIG_LOCALVERSION is a string setting. It needs to go between quotes in the .config file.

$ grep CONFIG_LOCALVERSION= .config
CONFIG_LOCALVERSION="-dey"


classes/fsl-kernel-localversion.bbclass
```diff
@@ -20,7 +20,7 @@ do_kernel_localversion() {
 
 	# Fallback for recipes not able to use LINUX_VERSION_EXTENSION
 	if [ "${@bb.data.inherits_class('kernel-yocto', d)}" = "False" ]; then
-		echo "CONFIG_LOCALVERSION=${LOCALVERSION}" >> ${B}/.config
+		echo "CONFIG_LOCALVERSION=\"${LOCALVERSION}\"" >> ${B}/.config
```
quaresmajose, line 23: Avoiding escape chars is more readable in my opinion
```
echo '"CONFIG_LOCALVERSION="${LOCALVERSION}"' >> ${B}/.config
```

quaresmajose (approved):

classes/fsl-kernel-localversion.bbclass
otavio: Avoid the escales, it is nicer indeed.

jviguera: @otavio updated the commit using single quotes to avoid escaping the double quotes.

MrCry0 (approved):

otavio:
We need to fix this on master, first.

PR #1586, opened 2023-06-25
alsa-state: Add FILESEXTREAPATHS for NXP specific and Mainline BSP

junzhuimx:
Current implement can only apply the setting under folder imx-generic-bsp. Add paths for the NXP and mainline specific setting.

### DIFF
--- a/recipes-bsp/alsa-state/alsa-state.bbappend
+++ b/recipes-bsp/alsa-state/alsa-state.bbappend
@@ -1,4 +1,4 @@
 # Prepend path to override files from upstream recipe
-FILESEXTRAPATHS:prepend:imx-generic-bsp := "${THISDIR}/${PN}/imx-generic-bsp:"
+FILESEXTRAPATHS:prepend := "${THISDIR}/${PN}:"
 
 PACKAGE_ARCH:imx-generic-bsp = "${MACHINE_ARCH}"


kraj (approved):

recipes-bsp/alsa-state/alsa-state.bbappend
```diff
@@ -1,4 +1,6 @@
 # Prepend path to override files from upstream recipe
 FILESEXTRAPATHS:prepend:imx-generic-bsp := "${THISDIR}/${PN}/imx-generic-bsp:"
+FILESEXTRAPATHS:prepend:imx-nxp-bsp := "${THISDIR}/${PN}/imx-nxp-bsp:"
+FILESEXTRAPATHS:prepend:imx-mainline-bsp := "${THISDIR}/${PN}/imx-mainline-bsp:"
```
thochstein, line 4: I believe the way to fix this is to just drop the BSP overrides:

`FILESEXTRAPATHS:prepend := "${THISDIR}/${PN}:"

junzhuimx: @thochstein You are right. It is much more simple to implement it. Let me update it

recipes-bsp/alsa-state/alsa-state.bbappend
```diff
@@ -1,4 +1,4 @@
 # Prepend path to override files from upstream recipe
-FILESEXTRAPATHS:prepend:imx-generic-bsp := "${THISDIR}/${PN}/imx-generic-bsp:"
+FILESEXTRAPATHS:prepend := "${THISDIR}/${PN}:"
```
kraj, line 2: What happens when you have other BSP layers in project ? Would it insert itself into those too ?

junzhuimx: will not impact your case:
- if other BSP layers defines "imx-generic/nxp/mainline-bsp" in MACHINEOVERRIDES will pick related files under the special folder
- if no defined, they can add bbappend and add the same ILESEXTRAPATHS:prepend := "${THISDIR}/${PN}:" to pick up the version in their layers.
alsa-state
├── imx-generic-bsp
│   └── asound.state
├── imx-mainline-bsp
│   └── asound.conf
└── imx-nxp-bsp
    └── asound.conf

kraj: That’s sounds good then. I am fine with this change

PR #1598, opened 2023-07-05
image_types_fsl: IMAGE_NAME_SUFFIX is now directly included

quaresmajose:
The IMAGE_NAME_SUFFIX is now directly included in both IMAGE_NAME and IMAGE_LINK_NAME.

https://git.yoctoproject.org/poky/commit/?id=6f6c79029bc2020907295858449c725952d560a1

### DIFF
--- a/classes/image_types_fsl.bbclass
+++ b/classes/image_types_fsl.bbclass
@@ -40,5 +40,5 @@ do_image_wic[depends] += " \
 IMAGE_CMD:wic:append:mxs-generic-bsp() {
 	# Change partition type for mxs processor family
 	bbnote "Setting partition type to 0x53 as required for mxs' SoC family."
-	echo -n S | dd of=$out${IMAGE_NAME_SUFFIX}.wic bs=1 count=1 seek=450 conv=notrunc
+	echo -n S | dd of=${out}.wic bs=1 count=1 seek=450 conv=notrunc
 }


quaresmajose:
I don't use this IMAGE_CLASSES so this PR has not been tested not even build but i think it is necessary

classes/image_types_fsl.bbclass
```diff
@@ -40,5 +40,5 @@ do_image_wic[depends] += " \
 IMAGE_CMD:wic:append:mxs-generic-bsp() {
 	# Change partition type for mxs processor family
 	bbnote "Setting partition type to 0x53 as required for mxs' SoC family."
-	echo -n S | dd of=$out${IMAGE_NAME_SUFFIX}.wic bs=1 count=1 seek=450 conv=notrunc
+	echo -n S | dd of=$out.wic bs=1 count=1 seek=450 conv=notrunc
```
MaxKrummenacher, line 43: While I do not use this image_types class, I guess this should be:
s/$out.wic/${out}.wic/

quaresmajose: good catch! not testing is always very expensive.
Thanks

quaresmajose:
fixed in https://github.com/Freescale/meta-freescale/pull/1690

PR #1602, opened 2023-07-12
Upgrade NXP Wi-Fi driver to L6.1.22-2.0.0 and rename it to kernel-module-nxp-wlan

junzhuimx:
The NXP W-Fi driver is the all-in-one driver to support all of NXP Wi-Fi modules.
Rename the recipe name to provide the right information.


recipes-kernel/kernel-modules/kernel-module-nxp-wlan_git.bb
otavio, line 1: Please add an RPROVIDES for the old name.

otavio (changes requested):

recipes-kernel/kernel-modules/kernel-module-nxp-wlan_git.bb
junzhuimx: Update recipe already

PR #1610, opened 2023-07-19
[kirkstone] smw: Add "Security Middleware Library" recipe

matteofacchinetti:
I'm interested in pkcs11 API provided by this package and I tested on imx8qm:
 
 TEST tests_pkcs11_random PASSED
 Ran 15 tests with 0 failures

### DIFF
--- a/recipes-security/swm/smw_git.bb
+++ b/recipes-security/swm/smw_git.bb
@@ -1,0 +1,63 @@
+# Copyright 2020-23 NXP
+
+SUMMARY = "NXP i.MX Security Middleware Library"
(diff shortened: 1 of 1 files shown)


matteofacchinetti:
Unfortunately this recipes is valid only for this branch.
For the master, is needed to add some changes due to the different OP-TEE version.

otavio:
Even if there are changes for kirkstone we need to get it in master first

matteofacchinetti:
OK, I have investigated: last version works properly in this branch and also in the master. And recipe is compatible.
So updated. I will push in master too.

PR #1614, opened 2023-07-21
[master] smw: Add "Security Middleware Library" recipe

matteofacchinetti:
I'm interested in pkcs11 API provided by this package and I tested on imx8qm:

TEST tests_pkcs11_random PASSED
Ran 15 tests with 0 failures


recipes-security/swm/smw_git.bb
```diff
@@ -0,0 +1,63 @@
+# Copyright 2020-23 NXP
```
angolini, line 1: Is this recipe from a NXP BSP? Could you, please, add in the commit log which tag/version exactly?

PR #1616, opened 2023-07-22
imx93-evk: use correct name for the iw612 machine feature

ricardosalveti:
imx-base.inc uses nxpiw612-sdio instead of iw612 when adding the relevant firmware packages.


ricardosalveti:
@angolini ^^

PR #1619, opened 2023-07-24
gstreamer: Upgrade 1.20.3.imx -> 1.22.0.imx

thochstein:


conf/machine/include/imx-base.inc
```diff
@@ -478,17 +478,10 @@ MACHINE_GSTREAMER_1_0_PLUGIN:mx8qxp-nxp-bsp ?= "imx-gst1.0-plugin"
...
-PREFERRED_VERSION_gstreamer1.0-libav:mx8-nxp-bsp        ??= "1.20.3.imx"
-PREFERRED_VERSION_gstreamer1.0-rtsp-server:mx8-nxp-bsp  ??= "1.20.3.imx"
-PREFERRED_VERSION_gstreamer1.0-python:mx8-nxp-bsp       ??= "1.20.3.imx"
-PREFERRED_VERSION_gstreamer1.0-vaapi:mx8-nxp-bsp        ??= "1.20.3.imx"
-PREFERRED_VERSION_gstreamer1.0-omx:mx8-nxp-bsp          ??= "1.20.3.imx"
-PREFERRED_VERSION_gst-devtools:mx8-nxp-bsp              ??= "1.20.3.imx"
+PREFERRED_VERSION_gstreamer1.0:mx8-nxp-bsp              ??= "1.22.0.imx"
+PREFERRED_VERSION_gstreamer1.0-plugins-base:mx8-nxp-bsp ??= "1.22.0.imx"
+PREFERRED_VERSION_gstreamer1.0-plugins-good:mx8-nxp-bsp ??= "1.22.0.imx"
+PREFERRED_VERSION_gstreamer1.0-plugins-bad:mx8-nxp-bsp  ??= "1.22.0.imx"
```
otavio, line 484: What about the other recipe files? We should add them as we did in past.

otavio (changes requested):

conf/machine/include/imx-base.inc
thochstein: I wasn't sure. You want recipe copies even when they are the same? Shall I copy the latest, 1.22.4, or better to align and use 1.22.0?

otavio: I'd use the ones that match the version inside the i.MX fork.

recipes-multimedia/gstreamer/gstreamer1.0-plugins-bad_1.22.0.imx.bb
```diff
@@ -3,7 +3,7 @@
 # recipe. The second section customizes the recipe for i.MX.
 
 ########### OE-core copy ##################
-# Upstream hash: a21649109374fde44cf77de845cfb3cb6cbfb138
+# Upstream hash: fb2d28e0315ece6180c87c7047587673024a09f7
 
 require recipes-multimedia/gstreamer/gstreamer1.0-plugins-common.inc
 require recipes-multimedia/gstreamer/gstreamer1.0-plugins-license.inc
```
thochstein, line 9: This should use the local copy.

otavio (approved):

recipes-multimedia/gstreamer/gstreamer1.0-plugins-bad_1.22.0.imx.bb
otavio: Could you please provide me with more details?

thochstein: The original commit that copied the downgraded recipes didn't copy the .inc files. I guess if we're taking a snapshot we should copy the .inc files as well, right?

otavio: Sure, it makes sense. Thanks for looking at this.

otavio (approved):

recipes-multimedia/gstreamer/gstreamer1.0-plugins-bad_1.22.0.imx.bb
quaresmajose: yeah using the inc files available on oe-core will get us out of sync quickly. however since the names are the same I need to confirm locally which inc are used, to check if we use the ones on the layer or the upstream.

conf/machine/include/imx-base.inc
```diff
@@ -478,18 +478,21 @@ MACHINE_GSTREAMER_1_0_PLUGIN:mx8qxp-nxp-bsp ?= "imx-gst1.0-plugin"
...
+
+# GStreamer copied recipes
 PREFERRED_VERSION_ffmpeg:mx8-nxp-bsp                    ??= "4.4.1"
+PREFERRED_VERSION_gst-devtools:mx8-nxp-bsp              ??= "1.22.0.imx"
+PREFERRED_VERSION_gstreamer1.0-libav:mx8-nxp-bsp        ??= "1.22.0.imx"
+PREFERRED_VERSION_gstreamer1.0-omx:mx8-nxp-bsp          ??= "1.22.0.imx"
+PREFERRED_VERSION_gstreamer1.0-plugins-ugly:mx8-nxp-bsp ??= "1.22.0.imx"
+PREFERRED_VERSION_gstreamer1.0-python:mx8-nxp-bsp       ??= "1.22.0.imx"
+PREFERRED_VERSION_gstreamer1.0-rtsp-server:mx8-nxp-bsp  ??= "1.22.0.imx"
+PREFERRED_VERSION_gstreamer1.0-vaapi:mx8-nxp-bsp        ??= "1.22.0.imx"
```
quaresmajose, line 495: Look like the `mx9-nxp-bsp` machines still uses the upstream version oe-core, is this what is intended?
is `imx9` not supported on the `1.22.0.imx` fork?

thochstein: `imx9` is supported by the fork. I'll fix it.

recipes-multimedia/gstreamer/gstreamer1.0-plugins-bad_1.22.0.imx.bb
thochstein: I'm not sure if you are suggesting that we should keep the full path. In case you are, to be clear, these lines were changed to the full path after copying the recipes in order to find the `.inc` files in oe-core. Changing the lines back just re-aligns them with what is in oe-core.

conf/machine/include/imx-base.inc
thochstein: The `ffmpeg` downgrade was done specifically for gstreamer, so I guess that's why it is grouped together. What do you think, @otavio?

otavio: We can move but add a comment so it is documented why it is need.

otavio (approved):

quaresmajose (approved):
LGTM

PR #1643, opened 2023-08-23
imx8dxl*evk.conf: Rename machines for compatibility

thochstein:
The meta-imx layer does already use these machine config files, but without a hyphen before the rev in the machine name. The hyphen is arguably better, and might have been preferred at the beginning, but at this point breaking compatibility is not worth it.


MrCry0 (approved):

otavio:
I suggest using an alias for the meta-imx machines, as this will make it easier to convert to them in the future.

PR #1645, opened 2023-08-26
NXP WiFi firmware

thochstein:
Revised fix for #1639. Depends on OE-Core update to `linux-firmware`. That is currently pending a final internal review before I submit the change to OE-Core.

PR #1646, opened 2023-08-29
splitter brcm Wi-Fi firmware to firmware-brcm-wifi

junzhuimx:
firmware-nxp-wifi contains Wi-Fi firmware from NXP only


thochstein:
Hi Jun, what is the benefit of moving this package to a separate recipe?

junzhuimx:
firmware-nxp-wifi is defined for NXP Wi-Fi but bcm4359 is a "black sheep", as it is used for Cypress (Broadcom) Wi-Fi module.
And BRCM4359 does not upstream to linux-firmware.git, so the separate recipe should be more reasonable.

junzhuimx:
@otavio 
There is  no more updates for bcm4359 firmware in imx-firmware.git. So after thus splitting, no more maintain needs for the recipe "firmware-brcm-wifi" until Cypress upstream them to linux-firmware.git.
We only focus on the firmware-nxp-wifi, which provides the latest NXP firmware for community/customers to "earlier" access.

otavio:
#1645 seems to address all issues. Can we close this?

junzhuimx:
@otavio 
https://github.com/Freescale/meta-freescale/pull/1645 is for NXP Wi-Fi firmware only.
This PR is just splitter BRCM firmware to new recipe. and does not conflict with https://github.com/Freescale/meta-freescale/pull/1645

PR #1667, opened 2023-10-17
Bump linux-imx to lf-6.1.36_2.1.0

angolini:
Change:
* update the kernel source code
* remove the useless patches
* include the new dts


angolini:
build test only for imx8mm-lpddr4-evk so far. Performing build test for imx93-11x11-lpddr4x-evk right now

conf/machine/imx8mp-lpddr4-evk.conf
```diff
@@ -42,6 +42,7 @@ KERNEL_DEVICETREE:append:use-nxp-bsp = " \
     freescale/imx8mp-evk-rpmsg.dtb \
     freescale/imx8mp-evk-rpmsg-lpv.dtb \
     freescale/imx8mp-evk-sof-wm8960.dtb \
+    freescale/imx8mp-evk-sof-pdm.dtb \
```
angolini, line 45: @otavio, @thochstein  two questions here. 

I expected to find KERNEL_DEVICETREE for imx8mp-ddr4-evk.conf and imx8mp-lpddr4-evk.conf in the imx8mp-evk.inc file. Is there a reason? For me it looks like the .inc file is a better place.

Should we be using ${KERNEL_DEVICETREE_BASENAME} instead?

I don´t  know those boards so I'm not sure about this.

thochstein: I guess we don't support both board flavors with the same devicetrees?

Using `KERNEL_DEVICETREE_BASENAME` is a good idea.

angolini: I don't know. The dtb filenames are "basic", that's why I guessed it would go to the .inc file.

thochstein: I guess you could check the kernel to see.

thochstein: I guess you could check the kernel to see.

otavio (approved):

MrCry0 (approved):
LGTM

PR #1668, opened 2023-10-17
firmware-nxp-wifi: make it compatible with imx-nxp-bsp

quaresmajose:
There multiple providers for this so make this one imx-nxp-bsp compatible and chose linux-firmware provider in other bsp.
Otherwise when other bsp layers are available we need to set the PREFERRED_RPROVIDER.


otavio (changes requested):
Put this as end of file.

quaresmajose:
> Put this as end of file.

Fixed. Thanks

MrCry0 (approved):

recipes-bsp/firmware-imx/firmware-nxp-wifi_1.0.bb
```diff
@@ -194,3 +194,5 @@ RDEPENDS:${PN}-nxpiw612-sdio += "${PN}-nxp-common"
 RPROVIDES:${PN}-nxpiw612-sdio = "linux-firmware-nxpiw612-sdio"
 RREPLACES:${PN}-nxpiw612-sdio = "linux-firmware-nxpiw612-sdio"
 RCONFLICTS:${PN}-nxpiw612-sdio = "linux-firmware-nxpiw612-sdio"
+
+COMPATIBLE_MACHINE = "(imx-nxp-bsp)"
```
thochstein, line 198: Is it compatible with `imx-mainline-bsp`?

quaresmajose: yes, it is also compatible with `imx-mainline-bsp` but imo for mainline is better to use version provided by `linux-firmware`

thochstein: So shouldn't it be marked compatible but not recommended for mainline?

PR #1678, opened 2023-10-31
ARM 32-bit device-tree reorganization

twoerner:
The 32-bit ARM device trees in the Linux kernel were reorganized in a manner similar to how the 64-bit ARM device trees have always been organized: by placing them in vendor+family subdirectories. Therefore update the KERNEL_DEVICETREE definitions to match.

### DIFF
--- a/conf/machine/imx23evk.conf
+++ b/conf/machine/imx23evk.conf
@@ -18,4 +18,4 @@ UBOOT_SUFFIX = "sb"
 
 UBOOT_MACHINE = "mx23evk_config"
 
-KERNEL_DEVICETREE = "imx23-evk.dtb"
+KERNEL_DEVICETREE = "nxp/mxs/imx23-evk.dtb"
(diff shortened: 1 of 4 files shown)


twoerner:
Oops! This patch needs a v2

twoerner:
Okay, ready.

angolini (approved):
Thanks for taking care of those boards!
I would like to hear if there are someone out there still using the machines!

thochstein:
Doesn't this break older kernel builds?

twoerner:
@angolini I am still actively using the Olimex board: olinuxino-maxi

twoerner:
@thochstein hmm... it does. It's frustrating that a single yocto branch includes pre- and post- dts directory cleanup versions. Is there a way to provide version-specific KERNEL_DEVICETREE variables?

Applying this patch fixes 6.6 builds and breaks 6.1 builds.
Not applying this patch breaks 6.6 builds and allows 6.1 to continue working.
6.6 is the default.

otavio:
I prefer fixing the issue in the LTS kernel recipe instead of delaying it.

otavio:
I suggest we fix the old kernel recipes instead of reworking the device tree path, to avoid future issues. So when kernel 6.1 is removed the code goes away with it.

PR #1681, opened 2023-11-01
imx-atf: Explicitly demand BFD linker

kraj:
This component uses BFD linker specific options which may not be available when default ld is not GNU BFD LD

Fixes

| aarch64-yoe-linux-ld: error: unknown argument '--fix-cortex-a53-835769'


MrCry0 (approved):

jviguera:
@kraj @otavio is this something we should backport to Kirkstone (after all it is a long-term support release)?

kraj:
> @kraj @otavio is this something we should backport to Kirkstone (after all it is a long-term support release)?

Yes it won’t hurt

PR #1682, opened 2023-11-02
 imx-mcore-demos: Fix install for multilib  Replace  `base_libdir` with `nonarch_base_libdir`

valijec:
When using multilib configuration there is a build break at packaging step

### DIFF
--- a/recipes-fsl/mcore-demos/imx-mcore-demos.inc
+++ b/recipes-fsl/mcore-demos/imx-mcore-demos.inc
@@ -40,9 +40,8 @@ S = "${WORKDIR}/${SOC}-${MCORE_TYPE}-demo-${PV}"
 SCR = "SCR-${SOC}-${MCORE_TYPE}-demo.txt"
 
 do_install () {
-    # install elf format binary to /lib/firmware
-    install -d ${D}${base_libdir}/firmware
-    install -m 0644 ${S}/*.elf ${D}${base_libdir}/firmware
+    install -d ${D}${nonarch_base_libdir}/firmware
+    install -m 0644 ${S}/*.elf ${D}${nonarch_base_libdir}/firmware
 }
 
 DEPLOY_FILE_EXT       ?= "bin"


recipes-fsl/mcore-demos/imx-mcore-demos.inc
```diff
@@ -57,6 +57,6 @@ addtask deploy after do_install
 
 PACKAGE_ARCH = "${MACHINE_SOCARCH}"
 
-FILES:${PN} = "${nonarch_base_libdir}/firmware"
+FILES:${PN} = "${baselib}/firmware"
```
thochstein, line 60: It's `do_install` that needs to be fixed, installing to `nonarch_base_libdir` instead.

thochstein (changes requested):
It's `do_install` that needs to be fixed, installing to `nonarch_base_libdir` instead.

recipes-fsl/mcore-demos/imx-mcore-demos.inc
valijec: Thank you @thochstein, fixed that

recipes-fsl/mcore-demos/imx-mcore-demos.inc
```diff
@@ -40,9 +40,9 @@ S = "${WORKDIR}/${SOC}-${MCORE_TYPE}-demo-${PV}"
 SCR = "SCR-${SOC}-${MCORE_TYPE}-demo.txt"
 
 do_install () {
-    # install elf format binary to /lib/firmware
-    install -d ${D}${base_libdir}/firmware
-    install -m 0644 ${S}/*.elf ${D}${base_libdir}/firmware
+    # install elf format binary to /usr/lib/firmware
```
quaresmajose, line 43: It will be better to drop this line as it depends on multilib or usermerge

thochstein (approved):

PR #1685, opened 2023-11-02
linux: Add compatibility logic for 32-bit dtb move

thochstein:
Newer kernels have moved the dtbs to a vendor sub-folder, including the ones for i.MX 32-bit, which now reside in nxp/imx. In order to maintain one KERNEL_DEVICETREE for both new and old kernels, provide logic that can be enabled in older kernels to strip the new sub-folder.

Fixes: 0d838c68

### DIFF
--- a/recipes-kernel/linux/linux-fslc-lts_6.1.bb
+++ b/recipes-kernel/linux/linux-fslc-lts_6.1.bb
@@ -12,6 +12,8 @@ upstreaming in any form."
 
 require linux-imx.inc
 
+KERNEL_DEVICETREE_32BIT_COMPATIBILITY_UPDATE = "1"
+
 SRC_URI = "git://github.com/Freescale/linux-fslc.git;branch=${KBRANCH};protocol=https"
 
 # PV is defined in the base in linux-imx.inc file and uses the LINUX_VERSION definition
--- a/recipes-kernel/linux/linux-imx.inc
+++ b/recipes-kernel/linux/linux-imx.inc
@@ -38,3 +38,29 @@ KCONFIG_MODE="--alldefconfig"
 # We need to pass it as param since kernel might support more then one
 # machine, with different entry points
 KERNEL_EXTRA_ARGS += "LOADADDR=${UBOOT_ENTRYPOINT}"
+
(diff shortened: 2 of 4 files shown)


recipes-kernel/linux/linux-imx.inc
```diff
@@ -38,3 +38,29 @@ KCONFIG_MODE="--alldefconfig"
...
+python kernel_devicetree_imx_32bit_compatibility_update() {
+    import os.path
+    import re
+    if d.getVar('KERNEL_DEVICETREE_IMX_32BIT_COMPATIBILITY_UPDATE') != "1" or d.getVar('TUNE_ARCH') != "arm":
+        return
+    input = d.getVar('KERNEL_DEVICETREE').split()
+    output = ""
+    stripped = ""
+    for original in input:
+        if re.match("^nxp/imx", original):
```
MrCry0, line 55: there are some machines supported in `meta-freescale` with dts in `nxp/mxs` directory as well

thochstein: Thanks! Will add it.

MrCry0: @thochstein actually, as there are no subdirs in `arch/arm/boot/dts` in kernels < 6.6, we could just strip any subdirs in `original` ;-)

otavio: Absolutely. That makes perfect sense.

recipes-kernel/linux/linux-fslc-lts_6.1.bb
```diff
@@ -12,6 +12,8 @@ upstreaming in any form."
 
 require linux-imx.inc
 
+KERNEL_DEVICETREE_32BIT_COMPATIBILITY_UPDATE = "1"
```
thochstein, line 15: Hmm, I just tested if I could override this in local.conf and it didn't work. Maybe need to use `?=` here.

otavio: Sure, I don't expect this to be overrideable inside `local.conf` as it's version-specific, not machine-specific.

MrCry0 (approved):
LGTM

otavio (approved):

PR #1690, opened 2023-11-10
image_types_fsl.bbclass: Drop use of IMAGE_NAME_SUFFIX

kraj:
This is no longer required since its already merged into IMAGE_NAME and IMAGE_LINK_NAME via [1] in OE-Core. Infact it causes problems now if we add .rootfs explicitly, this created non-existent filenames and resulting images are wrong.

[1] https://git.openembedded.org/openembedded-core/commit/meta/classes-recipe/image-artifact-names.bbclass?id=26d97acc71379ab6702fa54a23b6542a3f51779c


quaresmajose (approved):
LGTM

Issue #961, opened 2022-01-25
SCMVERSION ?= "y" Image creation fails when `linux-imx` is populated from sstate cache.


niziak:
Hello.
When external additional kernel modules are built by separate recipes following error is shown:
```
ERROR: XXX-image-1.0-r0 do_rootfs: Unable to install packages.
...
Solver encountered 2 problem(s):
Problem 1/2:

- package jailhouse-0.12-r0.imx8mmevk requires kernel-module-jailhouse-5.10.35-lts-5.10.y+gd5366be4cdf6, but none of the providers can be installed
- conflicting requests
- nothing provides kernel-5.10.35-lts-5.10.y+gd5366be4cdf6 needed by kernel-module-jailhouse-5.10.35-lts-5.10.y+gd5366be4cdf6-0.12-r0.imx8mmevk

Solution 1:
- do not ask to install a package providing jailhouse
```

Problem occurs in some specific scenario, where linux-imx is downloaded directly from sstate cache and additional modules (i.e. `jailhouse`) are built separately. Modules requires kernel source. Kernel source is unpacked and then patched by our local patches.
Every patch generates new commit in kernel source (every time different hash because current timestamp is part of hash base),.

When SCMVERSION is enabled (which is default) the kernel package name is different each time so the built module package name doesn't match to kernel.

Solution is to disable SCMVERSION by setting `SCMVERSION=""`

otavio:
Is there a way for me to reproduce it?

angolini:
I will close this issue. If there is a way to reproduce it on our side, please reopen and update with the how-to.

junzhuimx:
This patch is not friendly when SRCREV="${AUTOREV}", which will break as 
git rev-parse --verify --short AUTOINC 2> /dev/null

otavio:
How would you recommend we tackle this issue, @niziak and @junzhuimx?

dwagenk:
Probably [`SRCPV`](https://docs.yoctoproject.org/singleindex.html#term-SRCPV) is the correct value to use then. I don't know if it is univerally set (and we can just assume it to always be present) or if we need a special case handling for `AUTOREV`.

dwagenk:
[...]

PR #1694, opened 2023-11-28
classes: make localversion classes deterministic

dwagenk:
The localversion string built into the kernel and u-boot should be deterministic. With OpenEmbedded it is common practice to apply patches to the sources. The commit id of HEAD will differ each time the patching step is rerun. With the previous implementation of the fsl-{kernel,u- boot}-localversion classes this produced non-deterministic commit ids to be used. This is undesireable, because the commit-id read from the version string during runtime is not present in the sources. It also caused problems when building out-of-tree kernel-modules if the kernel itself was taken from sstate cache but the kernel-module needed to be rebuild and retriggered the kernels do_patch task (see #961).

Rework to produce a deterministic string that will use the commit id given as SRCREV and count the number of patches applied on top.

With e.g. commit id 11aabbcc and 5 patches this will result in a version string
  +g11aabbcc+p5

Fixes: https://github.com/Freescale/meta-freescale/issues/961

### DIFF
--- a/classes/fsl-kernel-localversion.bbclass
+++ b/classes/fsl-kernel-localversion.bbclass
@@ -25,8 +25,9 @@ do_kernel_localversion() {
 
 	if [ "${SCMVERSION}" = "y" ]; then
 		# Add GIT revision to the local version
-		head=`git --git-dir=${S}/.git rev-parse --verify --short HEAD 2> /dev/null`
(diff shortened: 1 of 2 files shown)


jviguera:
Hi @otavio 

I think this would be great to have in LTS Kirkstone, but being honest, I'm not sure if it would have any bad side effect.

otavio:
Let's way few days and we accept a backport.

PR #1700, opened 2023-12-07
Add support to NXP's iMX93 SoC A1 revision

jmarcoscosta:
Hello,

I'm currently in a project where the target board is the A1 revision of iMX93, and I picked meta-freescale's BSP to work upon.

At first, I built a core-image-minimal and it didn't boot. As it turns out, the layer only supported building firmware for iMX93's A0 revision. Basically, firmware-sentinel only supports A1 starting at [v0.11](https://github.com/nxp-imx/meta-imx/blob/mickledore-6.1.36-2.1.0/meta-bsp/recipes-bsp/firmware-sentinel/firmware-sentinel_0.11.bb).

I'm sending this pull request to update firmware-sentinel according to meta-imx's, and I also added some tweaks to dynamically set the firmware name (e.g, `mx93a0-ahab-container.img` or `mx93a1-ahab-container.img`), because the current implementation doesn't actually use `SECO_FIRMWARE_NAME` (and consequently neither `IMX_SOC_REV`).

Please do not hesitate to point any mistakes, or even to point a specific revision that already fixes (partially or entirely) this issue.

Best regards,
João Marcos Costa


recipes-bsp/imx-mkimage/imx-mkimage_git.inc
```diff
@@ -4,6 +4,7 @@ DEPENDS = "zlib-native openssl-native"
 
 SRC_URI = "git://github.com/nxp-imx/imx-mkimage.git;protocol=https;branch=${SRCBRANCH} \
            file://0001-iMX8M-soc.mak-use-native-mkimage-from-sysroot.patch \
+           file://0001-iMX9-soc.mak-dynamically-set-AHAB_IMG.patch \
```
otavio, line 7: Hey @thochstein, could you please ensure that you apply those changes for the upcoming release? Thank you!

otavio (approved):

recipes-bsp/imx-mkimage/imx-mkimage_git.inc
jmarcoscosta: Hello,

I'm going to test this tomorrow morning, but I was getting the very same error (which I'll try to reproduce and send the logs here) even though I was setting IMX_SOC_REV to "A1" by hand in my local.conf, so I don't really see how setting this elsewhere would give a different result. 

Kind regards,
João

thochstein: Indeed, I did run into the error as well. It seems that imx-mkimage is not up to date either.

thochstein: Hopefully this is fixed now by #1701.

thochstein: Hi @jmarcoscosta,

Glad to hear it works!

The firmware name is not hard-coded anymore [1] and is driven by `IMX_SOC_REV`, which sets `SECO_FIRMWARE_NAME` and gets passed to `imx-mkimage` [2].

[1] https://github.com/Freescale/meta-freescale/commit/0688f79193e697b1bd7444358a7b46c22b83ca2b
[2] https://github.com/nxp-imx/imx-mkimage/blob/lf-6.1.36_2.1.0/iMX9/soc.mak#L20-L21

In the end this error should never have happened had we done the upstreaming of NXP release 6.1.36-2.1.0 properly. I can try to flesh out the new commit messages with some of this information.

jmarcoscosta: I see, thanks!

Could you please give me a heads-up once this is all merged into meta-freescale's master?

otavio:
Replaced by #1701

tprrt:
Hello @otavio,

Can you backport these changes on the Kirkstone (LTS) branch?

Kind regards,
Thomas

otavio:
Please prepare a PR for this.

tprrt:
Hello @otavio,

> Please prepare a PR for this.
I'm preparing a pull-request to backport SoC revision support and sentinel firmware update, otherwise boards with new SoC revisions aren't usable with the Kirkstone (LTS) branch.

But what do you advise about Freescale EULA ?

Because the sentinel firmware v0.11 is using the EULA V48 that isn't available on the Kirkstone branch.

Kind regards,
Thomas Perrot

otavio:
Could it be added? we must confirm it does not break old recipes.

tprrt:
Hello @otavio,

Moreover, it seems that the sentinel API has been modified since the 0.9 and the U-Boot 2022.04 used in the Kirkstone (LTS) branch doesn't use this new API.

So, to be able to use new SoC revision with the Kirkstone (LTS) it should be also necessary to update U-Boot or to backport patches adding the new sentinel API.

- Have you planned to support all SoC revision on the Kirkstone (LTS) branch until 2026?
- Have you planned to quickly release a Scarthgap (the next LTS) branch in April ?

Kind regards,
Thomas

ricardosalveti:
We are using meta-freescale master with kirkstone (currently using based on the previous BSP rev, about to update to the latest), and it works fine with imx93 a1.

ricardosalveti:
I believe not only u-boot, but optee might need to be updated as well, which is why backporting is a pain.

PR #1713, opened 2023-12-27
imx93: add fixes to enable multilib out-of-the-box

jmarcoscosta:
Hello,

I have been trying to build a 32-bit userspace + 64-bit kernel system through multilib. However, this feature doesn't seem to work out-of-the-box, and some fixes were needed so I could have a flawless build. The target I used in this context was `lib32-core-image-minimal`.

The main goal of this PR is to fix erroneous overrides that end up using the userspace architecture. Those overrides are not taken into consideration if the userspace architecture is something other than `aarch64`.

Best regards,
João


jmarcoscosta:
Update: the override I previously used was restricted to `iMX9`, even though `iMX8` is also `aarch64`, and that would break the build for `iMX8`.

I added a new override, namely `imx-aarch64` to represent them both, and updated the following commits to use it instead of `mx9-generic-bsp`.

otavio:
Thank you for your work on this pull request. However, I would prefer to explicitly keep the overrides. Would you mind reworking it to use `mx8-generic-bsp` and `mx9-generic-bsp`?

jmarcoscosta:
Hello, @otavio 

I'm not sure if I understood you correctly. Did you mean having each variable defined twice, using the two overrides?

Like this:

```
foo:mx8-generic-bsp = "bar"
foo:mx9-generic-bsp = "bar"
```

Best regards,
João

otavio:
@jmarcoscosta exactly! 👍🏻

jmarcoscosta:
PR updated.

classes/fsl-dynamic-packagearch.bbclass
```diff
@@ -74,7 +74,7 @@ python __anonymous () {
             d.appendVar("PACKAGE_EXTRA_ARCHS", " %s" % d.getVar("THUMB_EXTRA_SOCARCH"))
 }
 
-MACHINE_SOCARCH = "${TUNE_PKGARCH}${MACHINE_SOCARCH_SUFFIX}"
+MACHINE_SOCARCH = "${TUNE_PKGARCH}"
```
thochstein, line 77: Please see #1717.

Issue #1496, opened 2023-04-13
[qoriq] Error building `u-boot-qoriq` for QorIQ machines - `ModuleNotFoundError: No module named 'pkg_resources'`


otavio:
I encountered a build failure while attempting to build the `u-boot-qoriq` recipe version 2021.04+fslgit-r0. This error seems to be caused by a missing module called `pkg_resources`. The build command involves using `binman` tool that calls this missing module, causing the build to fail.

The build fails on the following machines:

- [ ] `mpc8548cds`
- [ ] `p2041rdb`
- [ ] `p3041ds`
- [ ] `p4080ds`
- [ ] `p5040ds-64b`
- [ ] `t4240rdb-64b`

These machines have been broken for a long time and if no one is interested in maintaining them, it would be preferable to remove them from the repository.

Please advise on how to proceed with fixing the build for these machines or removing them from the repository.

otavio:
This is still reproducible. Are anyone going to look at this or should we remove the machines?

jeffkize:
I fixed it by updating the recipe (adding python3-setuptools-native):
DEPENDS = "libgcc virtual/${TARGET_PREFIX}gcc bison-native bc-native swig-native python3-native python3-setuptools-native"

otavio:
@jeffkize, would you be able to create a pull request that resolves this issue?

MrCry0:
@otavio does the 1741 fix this issue?

otavio:
Indeed; closing.

PR #1741, opened 2024-01-25
u-boot-qoriq: add 'python3-setuptools-native' to depends

mjeanson:
Building this recipe on an Unbuntu 20.04 host results in the following erro:

    Traceback (most recent call last):
      File "u-boot-qoriq/2021.04+fslgit-r0/git/tools/binman/binman", line 39, in <module>
        from binman import control
      File "u-boot-qoriq/2021.04+fslgit-r0/git/tools/binman/../binman/control.py", line 11, in <module>
        import pkg_resources
    ModuleNotFoundError: No module named 'pkg_resources'

With the added dependency the build completes.


mjeanson:
Fixes #1496

PR #1746, opened 2024-02-06
jailhouse-imx: Avoid the build error when using -mcpu and -march

angolini:
It changes the Makefile, so the recipe is the only one controlling the build arguments.

It is related to [1]

[1] https://git.yoctoproject.org/poky/commit/?id=04eac1f2b67eac5f892a9e0f8fcfe54849923af5

---

Now it builds

### DIFF
--- a/recipes-extended/jailhouse/files/arm-arm64-Makefile-Remove-march-option-from-Makefile.patch
+++ b/recipes-extended/jailhouse/files/arm-arm64-Makefile-Remove-march-option-from-Makefile.patch
@@ -1,0 +1,40 @@
+From 7117c1fe979053e4b1071968b444c03fbffe55bb Mon Sep 17 00:00:00 2001
+From: Daiane Angolini <daiane.angolini@foundries.io>
+Date: Tue, 6 Feb 2024 18:01:07 -0300
+Subject: [PATCH] arm:arm64: Makefile: Remove -march option from Makefile
+
+It let the recipe to decide which are the compilation flags to be used.
+
+It acomodates to [1].
+
+[1] https://git.yoctoproject.org/poky/commit/?id=04eac1f2b67eac5f892a9e0f8fcfe54849923af5
+
+Upstream-Status: Inappropriate [See above]
+
+Signed-off-by: Daiane Angolini <daiane.angolini@foundries.io>
+---
+ hypervisor/arch/arm/Makefile   | 2 --
+ hypervisor/arch/arm64/Makefile | 1 -
+ 2 files changed, 3 deletions(-)
+
+diff --git a/hypervisor/arch/arm/Makefile b/hypervisor/arch/arm/Makefile
+index a54370fc..cddb6cd3 100644
+--- a/hypervisor/arch/arm/Makefile
++++ b/hypervisor/arch/arm/Makefile
+@@ -12,5 +12,3 @@
+
+ LINUXINCLUDE += -I$(src)/arch/arm-common/include
+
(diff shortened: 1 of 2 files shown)


thochstein (commented):
Looks good to me except that the patch cannot be upstreamed as-is. Can you see if there is some way to keep the existing config as the default but overridable by our recipe?

MrCry0 (approved):

quaresmajose (approved):
I have tested it locally and it fixes the the build error
```
|   /srv/oe/build/tmp-lmp/work-shared/imx93-11x11-lpddr4x-evk/kernel-source/scripts/check-local-export /srv/oe/build/tmp-lmp/work/imx93_11x11_lpddr4x_evk-lmp-linux/jailhouse-imx/git/git/configs/arm64/foundation-v8-inmate-demo.o
| cc1: error: switch '-mcpu=cortex-a55+crypto' conflicts with '-march=armv8-a+nofp' switch [-Werror]
|   /srv/oe/build/tmp-lmp/work-shared/imx93-11x11-lpddr4x-evk/kernel-source/scripts/check-local-export /srv/oe/build/tmp-lmp/work/imx93_11x11_lpddr4x_evk-lmp-linux/jailhouse-imx/git/git/inmates/lib/arm64/../arm-common/setup.o
|   /srv/oe/build/tmp-lmp/work-shared/imx93-11x11-lpddr4x-evk/kernel-source/scripts/check-local-export /srv/oe/build/tmp-lmp/work/imx93_11x11_lpddr4x_evk-lmp-linux/jailhouse-imx/git/git/configs/arm64/espressobin.o
|   /srv/oe/build/tmp-lmp/work-shared/imx93-11x11-lpddr4x-evk/kernel-source/scripts/check-local-export /srv/oe/build/tmp-lmp/work/imx93_11x11_lpddr4x_evk-lmp-linux/jailhouse-imx/git/git/inmates/lib/arm64/../arm-common/timing.o
|   /srv/oe/build/tmp-lmp/work-shared/imx93-11x11-lpddr4x-evk/kernel-source/scripts/check-local-export /srv/oe/build/tmp-lmp/work/imx93_11x11_lpddr4x_evk-lmp-linux/jailhouse-imx/git/git/inmates/lib/arm64/../arm-common/uart-imx-lpuart.o
```

angolini:
> Looks good to me except that the patch cannot be upstreamed as-is. Can you see if there is some way to keep the existing config as the default but overridable by our recipe?

@thochstein 
I couldn't think of a way to override the default from Makefile. Do you know how?

Another way would be using sed, do you prefer this instead?

```
@@ -60,6 +59,9 @@ do_configure:prepend() {
    then
       cp "${STAGING_DIR_HOST}/${CELLCONF_DIR}/"*.c ${S}/configs/${ARCH}/
    fi
+
+   sed -i "/KBUILD_CFLAGS/d" ./hypervisor/arch/arm64/Makefile
+   sed -i "/KBUILD_CFLAGS/d" ./hypervisor/arch/arm/Makefile
 }
```

thochstein:
> > Looks good to me except that the patch cannot be upstreamed as-is. Can you see if there is some way to keep the existing config as the default but overridable by our recipe?
> I couldn't think of a way to override the default from Makefile. Do you know how?
>  
> Another way would be using sed, do you prefer this instead?

I don't know of a better fix off hand, but I think the patch is preferable to the `sed` operation. Please comment on this in the patch commit message and add an `Upstream-Status: Inappropriate [See above]`.

Issue #1744, opened 2024-01-29
imx-gpu-viv-demos: differences between linux-imx and linux-fslc-imx kernels


blemouzy:
Hello,

I'm working with Yocto Kirkstone on i.MX6 CPU (i.MX6D rev1.3 at 792MHz).

meta-freescale imx-gpu-viv recipe provides imx-gpu-viv-demos package which can be used to test the GPU.
The script `/unit_tests/GPU/gpu.sh` run the following tests:
- `/opt/viv_samples/vdk/tutorial3 -f 100`
- `/opt/viv_samples/vdk/tutorial4_es20 -f 100`
- `/opt/viv_samples/tiger/tiger`

With linux-imx kernel:
- tutorial3 runs fine
- tutorial4_es20 runs fine with the following output:
```
Rendered 100 frames in 659 milliseconds: 151.75 fps
```
- tiger runs fine with the following output:
```
id=23, a,b,g,r=0,8,8,8, d,s=16,0, AA=0,openvgbit=79
frames:100 -- fps:99.206345
```
<img src="https://github.com/Freescale/meta-freescale/assets/124877155/57f84857-dc9b-49be-8eac-322ba3a1fd42" width="640">

But with linux-fslc-imx kernel:
- tutorial3 runs fine
- tutorial4_es20 runs fine with the following output:
```
Rendered 100 frames in 1697 milliseconds: 58.93 fps
```
- tiger doesn't run fine (wrong display) with the following output:
```
id=23, a,b,g,r=0,8,8,8, d,s=16,0, AA=0,openvgbit=79
frames:100 -- fps:58.411217
```

<img src="https://github.com/Freescale/meta-freescale/assets/124877155/ec832d30-7786-4faa-8b46-0a80c4bb04fb" width="640">

Do you know:
- Why FPS are lower on linux-fslc-imx kernel than on linux-imx?
- What tiger test does that can explain wrong display with linux-fslc-imx kernel? (closed source code)

blemouzy:
tiger test doesn't work due to missing `libOpenVG.so` link.

The link is located in the `-dev` package:
```
$ oe-pkgdata-util find-path /usr/lib/libOpenVG.3d.so.1.1.0
libopenvg-imx: /usr/lib/libOpenVG.3d.so.1.1.0

$ oe-pkgdata-util find-path /usr/lib/libOpenVG.so
libopenvg-imx-dev: /usr/lib/libOpenVG.so
```

[...]

PR #1747, opened 2024-02-06
imx-gpu-viv: Fix EGL dependencies

thochstein:
For each of the EGL rendering APIs, the EGL implementation does a dlopen for the unversioned library link to the rendering API implementation. So, a runtime dependency is needed for the -dev package of each such rendering API implementation.

Fixes: #1744


MrCry0 (approved):

angolini (approved):
good catch

PR #1748, opened 2024-02-07
jailhouse-imx: Update the recipe to align to the NXP BSP LF6.1.55_2.2.0

angolini:
Update the recipe to the BSP 6.1.55_2.2.0.


recipes-extended/jailhouse/jailhouse-imx_git.bb
```diff
@@ -31,6 +31,10 @@ DEPENDS = " \
     dtc-native \
 "
 
+RDEPENDS:${PN} += " \
+    pyjailhouse \
+"
```
quaresmajose, line 36: This RDEPENDS is already there at line 97 https://github.com/Freescale/meta-freescale/pull/1748/commits/1120b832e8a1c375a63029e3a528666c9a8e73ef#diff-5efdd6d33c3565dc2267d301eee873ed647c036a9b8a0bd50f21eef07e331673R101

angolini: thanks a lot!

PR #1759, opened 2024-03-02
qtbase: Enable wayland knob when in distro features

kraj:
Mimic the behavior from recipe proper here since we override packageconfig defaults. With QT 6.7, qtbase built without wayland support is not sufficient to build rest of QT components e.g. qtwayland errors out during build/configure

    ERROR: Qt Wayland Client requires QtGui to be build with support for wayland

Therefore add it when wayland is in distro features.


kraj:
ping, qt6.7 fails to build without this

PR #1760, opened 2024-03-05
imx-base.inc: Rename bmap-tools to bmaptool

mschwan-phytec:
bmaptool is now part of the Yocto Project and their recipe renamed it to bmaptool.

### DIFF
--- a/conf/machine/include/imx-base.inc
+++ b/conf/machine/include/imx-base.inc
@@ -637,7 +637,7 @@ WKS_FILE_DEPENDS ?= " \
     virtual/bootloader \
     \
     e2fsprogs-native \
-    bmap-tools-native \
+    bmaptool-native \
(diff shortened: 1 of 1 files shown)


conf/machine/include/imx-base.inc
```diff
@@ -637,7 +637,7 @@ WKS_FILE_DEPENDS ?= " \
     virtual/bootloader \
     \
     e2fsprogs-native \
-    bmap-tools-native \
+    bmaptool-native \
```
thochstein, line 640: This is only compatible with scarthgap.

mschwan-phytec: Correct. I did not see `LAYERSERIES_COMPAT` was set to more branches than `scarthgap` or `master`. Maybe we should create a `nanbield` branch not including this patch?

otavio: @thochstein submitted a patch to OE-Core that added fields for backward compatibility, allowing for reversion.

PR #1767, opened 2024-03-11
conf/layer: add bmaptool compatibility for old versions

quaresmajose:
- Fixup for: cf8b57d9 imx-base.inc: Rename bmap-tools to bmaptool
- This makes it possible to use the master branch with oe-core before the scarthgap release.


MrCry0 (approved):
LGTM

PR #1780, opened 2024-04-04
imx6qdlsabresd: Change device tree paths to linux mainline

rafluan:
Change device tree paths to linux mainline, this was changed to the latest linux version.


angolini:
when you say "latest" what do you mean? Or in other worlds. Would you mind adding the linux version this is related?

rafluan:
@angolini sorry, fixed!

PR #1762, opened 2024-03-08
uuu_bootloader_tag.bbclass: Add UUU_BOOTLOADER_UNTAGGED

hiagofranco:
Add UUU_BOOTLOADER_UNTAGGED flag to add the possibility to deploy a imx-boot binary without the tagged footer
'UUUBURNXXOEUZX7+A-XY5601QQWWZ%sEN' at the end.

The addition of the footer 'UUUBURNXXOEUZX7+A-XY5601QQWWZ%sEN' at the end causes issues when the binary is loaded into RAM memory, this footer may cause issues where the binary size exceeds the typical USB package size used by UUU (1021 bytes).

As example, if the footer is split across two distinct USB packages, such as sending 'UUUBURNXXXOE' in the last packet and 'UZX7+A-XY5601QQWWZ%sEND' in the next one, it can lead to UUU halting while waiting for a response in SDPS mode.

After sending 'UUUBURNXXXOE', iMX8MP's boot room starts SPL, not waiting for the next packet ('UZX7+A-XY5601QQWWZ%sEND'), leading to UUU breaking.

Therefore, deploy both binaries separately (-tagged and -untagged) to allow users to choose the appropriate binary.

(cherry picked from commit 5a906365fd29)
Signed-off-by: Hiago De Franco <hiago.franco@toradex.com>

PR #1791, opened 2024-04-18
Restore default bootloader to untagged version

thochstein:
The introduction of the UUU-tagged bootloader causes problems for the non-wic case, as reported here [1] and here [2].

[1] https://github.com/Freescale/meta-freescale/pull/1762
[2] https://github.com/nxp-imx/mfgtools/issues/416

This PR is an attempt to improve the original feature by limiting the use of the tagged file to the wic case, while restoring the default file back to the untagged version. This does effectively revert [1] as no longer needed.

### DIFF
--- a/conf/machine/include/imx-base.inc
+++ b/conf/machine/include/imx-base.inc
@@ -653,9 +653,9 @@ WKS_FILE_DEPENDS:append:imx-generic-bsp:mx9-generic-bsp = " \
     ${@oe.utils.ifelse(d.getVar('UBOOT_PROVIDES_BOOT_CONTAINER') == '0', 'imx-boot', '')} \
 "
 
-SOC_DEFAULT_WKS_FILE ?= "imx-uboot-bootpart.wks.in"
-SOC_DEFAULT_WKS_FILE:mx8-generic-bsp ?= "imx-imx-boot-bootpart.wks.in"
+SOC_DEFAULT_WKS_FILE                 ?= "imx-uboot-bootpart.wks.in"
 SOC_DEFAULT_WKS_FILE:mxs-generic-bsp ?= "imx-uboot-mxs-bootpart.wks.in"
+SOC_DEFAULT_WKS_FILE:mx8-generic-bsp ?= "imx-imx-boot-bootpart.wks.in"
 SOC_DEFAULT_WKS_FILE:mx9-generic-bsp ?= "imx-imx-boot-bootpart.wks.in"
 
 WKS_FILE ?= "${SOC_DEFAULT_WKS_FILE}"
--- a/classes/imx-boot-container.bbclass
+++ b/classes/imx-boot-container.bbclass
@@ -96,7 +96,7 @@ do_deploy:append() {
                     if [ ! -f "${DEPLOYDIR}/imx-boot" ]; then
                         ln -sf flash.bin-${MACHINE}-${type} flash.bin
                         ln -sf flash.bin-${MACHINE}-${type} imx-boot
-
+                        ln -sf flash.bin.tagged imx-boot.tagged
                     else
                         bbwarn "Use custom wks.in for $UBOOT_CONFIG = $type"
                     fi
--- a/classes/uuu_bootloader_tag.bbclass
+++ b/classes/uuu_bootloader_tag.bbclass
@@ -1,12 +1,20 @@
-# Append a tag to the bootloader image used in the SD card image. The tag
-# contains the size of the bootloader image so UUU can easily find the end of
-# the bootloader in the SD card image.
+# Create a tagged boot partition file for the SD card image file. The tag
+# contains the size of the boot partition image so UUU can easily find
+# the end of it in the SD card image file.
+#
+# IMPORTANT: The tagged boot partition file should never be used directly with
+#            UUU, as it can cause UUU to hang.
+
+UUU_BOOTLOADER                 = "${UBOOT_BINARY}"
+UUU_BOOTLOADER:mx8-generic-bsp = "${@d.getVar('UBOOT_PROVIDES_BOOT_CONTAINER') == '0' and 'imx-boot' or 'flash.bin'}"
+UUU_BOOTLOADER:mx9-generic-bsp = "${@d.getVar('UBOOT_PROVIDES_BOOT_CONTAINER') == '0' and 'imx-boot' or 'flash.bin'}"
+
 do_deploy:append() {
     if [ "${UUU_BOOTLOADER}" != "" ]; then
-        cp ${DEPLOYDIR}/${UUU_BOOTLOADER}   ${DEPLOYDIR}/${UUU_BOOTLOADER_TAGGED}
-        cp ${DEPLOYDIR}/${UUU_BOOTLOADER}   ${DEPLOYDIR}/${UUU_BOOTLOADER_UNTAGGED}
-        ln -sf ${UUU_BOOTLOADER_TAGGED}     ${DEPLOYDIR}/${UUU_BOOTLOADER}
-        stat -L -cUUUBURNXXOEUZX7+A-XY5601QQWWZ%sEND ${DEPLOYDIR}/${UUU_BOOTLOADER_TAGGED} \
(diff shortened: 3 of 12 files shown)


thochstein:
Hi @hiagofranco, this PR does revert the addition of `UUU_BOOTLOADER_UNTAGGED` as no longer needed. Does this break you? Can you adjust?

otavio (approved):

conf/machine/include/imx-base.inc
```diff
@@ -656,11 +656,19 @@ WKS_FILE_DEPENDS:append:imx-generic-bsp:mx9-generic-bsp = " \
...
-SOC_DEFAULT_WKS_FILE ?= "imx-uboot-bootpart.wks.in"
-SOC_DEFAULT_WKS_FILE:mx8-generic-bsp ?= "imx-imx-boot-bootpart.wks.in"
+SOC_DEFAULT_WKS_FILE                 ?= "imx-uboot-bootpart.wks.in"
 SOC_DEFAULT_WKS_FILE:mxs-generic-bsp ?= "imx-uboot-mxs-bootpart.wks.in"
+SOC_DEFAULT_WKS_FILE:mx8-generic-bsp ?= "imx-imx-boot-bootpart.wks.in"
 SOC_DEFAULT_WKS_FILE:mx9-generic-bsp ?= "imx-imx-boot-bootpart.wks.in"
 
+WIC_BOOTLOADER = " \
+    ${@bb.utils.contains('IMX_DEFAULT_BOOTLOADER', 'imx-boot',   'imx-boot-tagged', \
+       bb.utils.contains('IMX_DEFAULT_BOOTLOADER', 'u-boot-imx', '${WIC_BOOTLOADER_32BIT}', \
```
thochstein, line 666: I realize that `PREFERRED_PROVIDER_u-boot` and `PREFERRED_PROVIDER_virtual/bootloader` and be overridden, so should I use one of these here instead of `IMX_DEFAULT_BOOTLOADER`?

otavio: I think it must use an `imx-boot` prefix to decide what to use. Users might use different `IMX_DEFAULT_BOOTLOADER`, and we should respect that.

I am wondering how to do this in a generic way...

hiagofranco:
> Hi @hiagofranco, this PR does revert the addition of `UUU_BOOTLOADER_UNTAGGED` as no longer needed. Does this break you? Can you adjust?

Hi @thochstein, thanks for letting me know, this will break our one of our layers where we are using `${IMX_BOOT}-untagged` binary, but I can fix it on our side to get the `imx-boot` instead one this PR is merged. Please let me know if you want to backport this to kirkstone as well, so I will have to fix our kirsktone branch too.

hiagofranco:
Hi @thochstein and @otavio, sorry for the delay here, I was out last week.

So, the whole issue happens when we are using UUU to download the binary directly into RAM, as we do here at Toradex with Tezi. We put the SoC into recovery mode and then download the boot container and jump to it. UUU sends 40 bytes size packages if I am not wrong, therefore if we are unlucky enough (which happened to us) to have the binary size at an specific value (more info below), the last package will contain only the TAG and, because of that, U-Boot will start before the last package is transmitted, causing UUU to hang forever.

Here is the full issue we described to mfgtool repo: https://github.com/nxp-imx/mfgtools/issues/416

NXP reported they opened a ticket internally, but I am not sure what it is the status here. 

Anyway, if we are going to do changes, we need to make sure the untagged binary is deployed, and we can choose between them, because this tag will cause issues. And about upstream/downstream, actually this tag is only add here at `uuu_bootloader_tag.bbclass` file, so if you compile u-boot manually it should always work with UUU as far as I remember.

Thanks!

thochstein:
> NXP reported they opened a ticket internally, but I am not sure what it is the status here.

This is the result of that ticket ;-)

> Anyway, if we are going to do changes, we need to make sure the untagged binary is deployed, and we can choose between them, because this tag will cause issues.

I believe this PR satisfies this by restoring the default binary to be the untagged one.

Will the user wonder though which version to use and perhaps still make a mistake by selecting the tagged one? I wonder if uuu can detect the trailing UUUBURN tag and handle the potential problem better?

hiagofranco:
> This is the result of that ticket ;-)

Ah cool, thanks =)

> I believe this PR satisfies this by restoring the default binary to be the untagged one.
>

All right, with the untagged binary being deployed, we can simply use it, so no problems for us.
 
> Will the user wonder though which version to use and perhaps still make a mistake by selecting the tagged one? I wonder if uuu can detect the trailing UUUBURN tag and handle the potential problem better?

Hmmm, what happens if there is no tag and you want to flash this binary? IOW, is this tag really needed by UUU? Or is this something the boot ROM needs?

thochstein:
Okay, I clarified with Frank Li, the tag is only need by UUU when it is operating on the wic image file. I guess in the boot ROM it has no effect. I tweaked the bbclass comment and the commit message, but otherwise the content of the PR is as before.

conf/machine/include/imx-base.inc
```diff
@@ -653,11 +653,21 @@ WKS_FILE_DEPENDS:append:imx-generic-bsp:mx9-generic-bsp = " \
...
-SOC_DEFAULT_WKS_FILE ?= "imx-uboot-bootpart.wks.in"
-SOC_DEFAULT_WKS_FILE:mx8-generic-bsp ?= "imx-imx-boot-bootpart.wks.in"
+SOC_DEFAULT_WKS_FILE                 ?= "imx-uboot-bootpart.wks.in"
 SOC_DEFAULT_WKS_FILE:mxs-generic-bsp ?= "imx-uboot-mxs-bootpart.wks.in"
+SOC_DEFAULT_WKS_FILE:mx8-generic-bsp ?= "imx-imx-boot-bootpart.wks.in"
 SOC_DEFAULT_WKS_FILE:mx9-generic-bsp ?= "imx-imx-boot-bootpart.wks.in"
 
+IMX_BOOTPART_FILE_WKS                 = "${IMX_BOOTPART_FILE}"
+IMX_BOOTPART_FILE_WKS:imx-generic-bsp = \
+    "${@bb.utils.contains('PREFERRED_PROVIDER_u-boot', 'u-boot-imx', '${IMX_BOOTPART_FILE_TAGGED}', \
```
otavio, line 663: I believe this is not going to work for vendor, bootloaders and others. I am wondering if we couldn't standardize the file name and use it for all.

otavio (changes requested):

conf/machine/include/imx-base.inc
thochstein: So instead of `u-boot-tagged.imx` and `imx-boot-tagged` you want a common filename, like `imx-bootpart-wic.bin`?

otavio: I'm wondering if this is dependent of we using the downstream bootloader or the upstream bootloader. If this is the case, I believe that we should use a machine variable to decide if we are using one or another.

My understanding is that in the case with the upstream bootloader this is not necessary and work out of box. If this is the case, then we should mark this as U-Boot i.MX specific option and make that conditional to the value of a specific variable.

In the case of downstream, we can then set that accordingly if we are using the downstream or upstream versions. In the case of external vendors, they can set that in the machine accordingly.

How does that sound?

otavio:
I want you guys to take a look at the branch below.

https://github.com/Freescale/meta-freescale/tree/uuu

This is mostly inspired by Tom's work; however, I simplified it a little bit. More importantly, I reduced the places where we need to set the variables and the differences from the current metadata.

Tom, I rewrote your commit and added myself as a co-author to indicate that we worked together on this. I didn't take over the commit because you did most of the hard work of finding the inner details. If you agree with the new approach, please update this pull request with those commits. Otherwise, you can explain to me what you disagree with, and we can try to adjust that accordingly.

thochstein:
Looks good to me, thanks Otavio.

otavio (approved):

otavio:
@hiagofranco Is it possible for you to review the changes?

hiagofranco:
> @hiagofranco Is it possible for you to review the changes?

LGTM, thanks!

PR #1826, opened 2024-05-16
imx-base: follow renamed fw package name

MaxKrummenacher:
Commit 9a522207 ("firmware-imx: Use vpu-amphion and vpu-wave suffix.") renamed the vpu firmware package name from firmware-vpu-imx8 to firmware-imx-vpu-amphion.
Follow that in the setting of MACHINE_FIRMWARE setting.

@thochstein :
meta-imx is still using the package names from before commit 9a522207.

meta-imx-bsp/recipes-bsp/firmware-imx/firmware-imx_8.23.bb creates the old package names, thus the imx8 vpu fw will not be installed assuming the meta-imx recipe wins.
meta-imx-bsp/conf/machine/include/imx-base-extend.inc: installs " firmware-imx-vpu-imx95", thus the imx95 vpu fw will not be installed assuming the meta-freescale recipe wins.

Should we revert commit 9a522207 rather than merging what I propose in the recipe here?


thochstein:
@MaxKrummenacher , sorry I missed this. Thanks for the notice. We are fine with this.

PR #1843, opened 2024-06-12
linux: drop the old style kernels unshared S

quaresmajose:
The default source code location for kernel is defined on the kernel.bbclass and it was 'S = "${STAGING_KERNEL_DIR}"'

In reality, although the kernel has its code in 'S = "${WORKDIR}/git"' it ends up being moved to '${STAGING_KERNEL_DIR}' in the do_symlink_kernsrc kernel.bbclass function.

With this change, the do_symlink_kernsrc kernel.bbclass function will no longer create the symbolic link and the final result will be the same.


quaresmajose:
Found this issue on the oe-core mailing list https://lists.openembedded.org/g/openembedded-core/topic/106531570

PR #1862, opened 2024-07-09
fmc: update patch format

limeng-linux:
The code of fmc package is created in windows editer environment, and the CRLF is different from unix format, so update patch format so that apply patch smoothly.


otavio (changes requested):
Please fix commit log to use `fmc` as prefix.

limeng-linux:
thanks for helping me to improve the patch

Issue #1861, opened 2024-07-02
Upgrade BSP to LF6.6.23_2.0.0


otavio:
The immediate items to work on are:

- [ ]  Upgrade Linux kernel and headers (i.MX variants)
- [ ]  Upgrade Linux kernel (FSLC forks i.MX variants)
- [ ]  Upgrade U-Boot (FSLC forks i.MX variants)
- [ ]  Upgrade OP-TEE
- [ ]  Upgrade ATF (i.MX variants)
- [ ]  Upgrade Multimedia Packages
- [ ]  Upgrade Graphics Packages
- [ ]  Upgrade Jailhouse
- [ ]  Upgrade mcore-demos
- [ ]  Upgrade firmware-imx*
- [ ]  Upgrade imx-mkimage

The Release Notes are [here](https://www.nxp.com/docs/en/release-note/IMX_LINUX_RELEASE_NOTES.pdf) for checking.

ricardosalveti:
@angolini @MrCry0 ^ :-)

hiagofranco:
Hi @otavio, are these statuses updated?
For example, I see linux-fslc was already updated and backported to scarthgap. Thanks!

hiagofranco:
Hi @otavio, now that this issue is closed, can we start the backports to Scarthgap? How does it work?

If that is the case, I am happy to open a PR with the backports, I am just asking first to understand how the developing is done. These backports is something we need into our BSP. Thank you!

otavio:
Sure, this is complete, indeed.

Specifically, regarding the backports, I think the best way to do that is to cherry-pick every commit that should be backported to the stable branch. Still, we can also use this opportunity to backport fixes from the master to the branch.

We must be careful not to import patches that break the branch missing changes from the master from the OpenEmbedded-Core.

PR #1874, opened 2024-07-26
Update gstreamer and imx-opencl-converter from 6.6.3-1.0.0 to 6.6.23-2.0.0

hiagofranco:
#1861


hiagofranco:
A couple of comments from my side:
- `imx-gst1.0-plugin` version 4.9.0 is broken, I was not able to compile the new NXP branch here:  https://github.com/nxp-imx/imx-gst1.0-plugin/tree/MM_04.09.00_2405_L6.6.y. So I decided to keep it out of this PR, maybe I will (or someone else) come up with a fix to the code. For now, 4.8.3 is still being used.
- `gstreamer1.0-plugins-ugly` does not work on master, only on Scarthgap for now. This is because it depends on mpeg2dec which was removed from OE master, and ffmpeg should replace it in the future. As the OE master branch did not fix the `gstreamer1.0-plugins-ugly`  there yet [1], I am also keeping the mpeg2dec. We could integrate the change from master in the future.

[1] https://github.com/openembedded/openembedded-core/blob/master/meta/recipes-multimedia/gstreamer/gstreamer1.0-plugins-ugly_1.24.5.bb#L33

recipes-multimedia/gstreamer/gstreamer1.0_1.24.0.imx.bb
```diff
@@ -92,14 +92,21 @@ LIC_FILES_CHKSUM = " \
 "
 
 # Use i.MX fork of GST for customizations
-SRC_URI:remove = "https://gstreamer.freedesktop.org/src/gstreamer/gstreamer-${PV}.tar.xz"
+SRC_URI:remove = "https://gstreamer.freedesktop.org/src/gstreamer/gstreamer-${PV}.tar.xz \
+                file://0001-tests-respect-the-idententaion-used-in-meson.patch \
+                file://0002-tests-add-support-for-install-the-tests.patch \
+                file://0003-tests-use-a-dictionaries-for-environment.patch;striplevel=3 \
+                file://0004-tests-add-helper-script-to-run-the-installed_tests.patch;striplevel=3 \
+"
```
quaresmajose, line 100: wonder who was this tested? look like all the test patches were inadvertently removed

recipes-multimedia/gstreamer/gstreamer1.0_1.24.0.imx.bb
```diff
@@ -92,14 +92,21 @@ LIC_FILES_CHKSUM = " \
...
 SRC_URI:prepend = "${GST1.0_SRC};branch=${SRCBRANCH} "
 GST1.0_SRC ?= "gitsm://github.com/nxp-imx/gstreamer.git;protocol=https"
-SRCBRANCH = "MM_04.08.03_2312_L6.6.y"
-SRCREV = "1a43c16272a7f4274eb8260e03206a57f317d823"
+SRCBRANCH = "MM_04.09.00_2405_L6.6.y"
+SRCREV = "753c7b4471d14befd1b7c9cd3fa9af9f10ab4a08"
 
 S = "${WORKDIR}/git"
 
+PACKAGECONFIG[tests] = "-Dtests=enabled,-Dtests=disabled"
```
quaresmajose, line 108: This line is not needed because it is already present on line 24
https://github.com/Freescale/meta-freescale/pull/1874/commits/5c05405e1057681c698aafc3ec284500a69e1adf#diff-3a0a7153e6a185fd9204a4b9d6b987c1980cd3a2ee839fd11d72a582dbdd4675R43

quaresmajose (changes requested):
There are some inconsistency in the version used, some recipes uses the `${@get_gst_ver("${PV}")}` and others `1.24.0` and `1.24.5`

hiagofranco:
> There are some inconsistency in the version used, some recipes uses the `${@get_gst_ver("${PV}")}` and others `1.24.0` and `1.24.5`

fixed, thanks!

recipes-multimedia/gstreamer/gstreamer1.0_1.24.0.imx.bb
hiagofranco: Sorry, my bad, I copied the recipe from meta-imx layer and they explicity removed these patches there. But looking at the patch and the MR opened at gstreamer upstream, I have added them back.

The only thing is that I had to rework them, because they were not applying anymore. I also removed the "striplevel=3" from patches 0003 and 0004 because this was making them fail to apply. Please let me know if something else is needed, thank you.

recipes-multimedia/gstreamer/gstreamer1.0_1.24.0.imx.bb
hiagofranco: I am not sure about this, this recipe is copied from OE upstream and then these variables are replaced only at the end of the file, after the `########### i.MX overrides ################`.

The differences are:

```
PACKAGECONFIG[tests] = "-Dtests=enabled -Dinstalled_tests=true,-Dtests=disabled -Dinstalled_tests=false"
PACKAGECONFIG[tests] = "-Dtests=enabled,-Dtests=disabled"
```

So the flags are changing, but I do not know how relevant they are. That being said, should I remove this line anyway?

Thanks.

recipes-multimedia/gstreamer/gstreamer1.0_1.24.0.imx.bb
quaresmajose: I am not aware of the meta-imx but these patches are still available upstream in oe-core. They are so we can run the unit tests with the oe-core ptest infrastructure. 
All of these [patches](https://github.com/openembedded/openembedded-core/blob/662396533177b72cc1d83e95841b27f7e42dcb20/meta/recipes-multimedia/gstreamer/gstreamer1.0_1.24.5.bb#L18-L24) are still in use and mandatory upstream in oe-core otherwise we can't run the unit test on the target using the `ptest`.

recipes-multimedia/gstreamer/gstreamer1.0_1.24.0.imx.bb
quaresmajose: The meson "-Dtests" doesn't make much sense without the mentioned patches above. The [PACKAGECONFIG[tests]]( https://github.com/openembedded/openembedded-core/blob/662396533177b72cc1d83e95841b27f7e42dcb20/meta/recipes-multimedia/gstreamer/gstreamer1.0_1.24.5.bb#L36) is still in use as well upstream in oe-core.

quaresmajose:
> > There are some inconsistency in the version used, some recipes uses the `${@get_gst_ver("${PV}")}` and others `1.24.0` and `1.24.5`
> 
> fixed, thanks!

some still missing:

https://github.com/Freescale/meta-freescale/blob/8b234355c80f338f00e5cd7e5fcdda1009897f98/conf/machine/include/imx-base.inc#L526

https://github.com/Freescale/meta-freescale/blob/8b234355c80f338f00e5cd7e5fcdda1009897f98/conf/machine/include/imx-base.inc#L532

quaresmajose:
The `LICENSE_FLAGS_ACCEPTED` was spreed in some of the recipes and this is not quite correct because with that we are forcing any distro that uses this bsp layer to have this licences.
The `LICENSE_FLAGS_ACCEPTED` is a disto variable and we can't use them on the recipes or other global config in the bsp,
the right place for them is the `conf/local.conf` or chosen by the disto and defined here but not on the bsp layer.
```
LICENSE_FLAGS_ACCEPTED += " commercial "
```

quaresmajose:
Please try to follow the pattern in all recipes:
```
########### End of OE-core copy ###########

########### i.MX overrides ################

DEFAULT_PREFERENCE = "-1"
```
Maybe this seems a little petty but gstreamer was one of my gateway into oe-core and I'm just trying to help :)
Finally, thank you for your contribution and fixes.

recipes-multimedia/gstreamer/gstreamer1.0_1.24.0.imx.bb
hiagofranco: All right, thanks for the explanation. So let's keep them, not sure why meta-imx removed.

hiagofranco:
> > > There are some inconsistency in the version used, some recipes uses the `${@get_gst_ver("${PV}")}` and others `1.24.0` and `1.24.5`
> > 
> > 
> > fixed, thanks!
> 
> some still missing:
> 
> https://github.com/Freescale/meta-freescale/blob/8b234355c80f338f00e5cd7e5fcdda1009897f98/conf/machine/include/imx-base.inc#L526
> 
> https://github.com/Freescale/meta-freescale/blob/8b234355c80f338f00e5cd7e5fcdda1009897f98/conf/machine/include/imx-base.inc#L532

Actually for gstreamer1.0-python I selected 1.24.5 on purpose, because I could not compile version 1.24.0:

``` bash
Called: `/home/frh/tdx/oe/master/build/tmp/work/armv8a-mx8-tdx-linux/gstreamer1.0-python/1.24.0.imx/recipe-sysroot-native/usr/bin/pkg-config --libs python-3.12-embed` -> 0
stdout:
-lpython3.12
-----------
Run-time dependency python found: YES 3.12
Message: python_abi_flags = 

../gst-python-1.24.0/meson.build:84:16: ERROR: Unknown variable "python_opt".
```

So I checked the OE upstream and I aw they did not have any version 1.24.0, instead, they are using 1.24.5 (https://github.com/openembedded/openembedded-core/blob/master/meta/recipes-multimedia/gstreamer/gstreamer1.0-python_1.24.5.bb), therefore I thought we could use the same as the upstream (and this one compile without any issues). What is your opinion? Should we stick to 1.24.0?

hiagofranco:
> The `LICENSE_FLAGS_ACCEPTED` was spreed in some of the recipes and this is not quite correct because with that we are forcing any distro that uses this bsp layer to have this licences. The `LICENSE_FLAGS_ACCEPTED` is a disto variable and we can't use them on the recipes or other global config in the bsp, the right place for them is the `conf/local.conf` or chosen by the disto and defined here but not on the bsp layer.
> 
> ```
> LICENSE_FLAGS_ACCEPTED += " commercial "
> ```

Got it. I tried to compile, and I saw it was failing because the LICENSE_FLAGS_ACCEPTED was missing, so checking the meta-imx from NXP I saw they added it to their recipes (gstreamer1.0-libav, gstreamer1.0-plugins-ugly and ffmpeg), that is why I added them here too. Not sure what to do here, because without these lines we cannot use the recipes.

hiagofranco:
> Please try to follow the pattern in all recipes:
> 
> ```
> ########### End of OE-core copy ###########
> 
> ########### i.MX overrides ################
> 
> DEFAULT_PREFERENCE = "-1"
> ```

So this is also something that confuses me, to be honest. For recipes that are available on meta-imx, we have this "i.MX overrides" comment, which comes from meta-imx itself. For recipes they do not have it, gstreamer1.0-python is one example, we do not use it. So it should be correct. 

> 
> Maybe this seems a little petty but gstreamer was one of my gateway into oe-core and I'm just trying to help :) Finally, thank you for your contribution and fixes.

No worries, I really appreciate your help =) Reviewing now is better than fixing later, just part of the process, thanks for the comments.

recipes-multimedia/gstreamer/gstreamer1.0_1.24.0.imx.bb
hiagofranco: I removed the line 

```bash
PACKAGECONFIG[tests] = "-Dtests=enabled,-Dtests=disabled"
```

thanks for clarifying.

quaresmajose:
> > > > There are some inconsistency in the version used, some recipes uses the `${@get_gst_ver("${PV}")}` and others `1.24.0` and `1.24.5`
> > > 
> > > 
> > > fixed, thanks!
> > 
> > 
> > some still missing:
> > https://github.com/Freescale/meta-freescale/blob/8b234355c80f338f00e5cd7e5fcdda1009897f98/conf/machine/include/imx-base.inc#L526
> > 
> > https://github.com/Freescale/meta-freescale/blob/8b234355c80f338f00e5cd7e5fcdda1009897f98/conf/machine/include/imx-base.inc#L532
> 
> Actually for gstreamer1.0-python I selected 1.24.5 on purpose, because I could not compile version 1.24.0:
> 
> ```shell
> Called: `/home/frh/tdx/oe/master/build/tmp/work/armv8a-mx8-tdx-linux/gstreamer1.0-python/1.24.0.imx/recipe-sysroot-native/usr/bin/pkg-config --libs python-3.12-embed` -> 0
> stdout:
> -lpython3.12
> -----------
> Run-time dependency python found: YES 3.12
> Message: python_abi_flags = 
> 
> ../gst-python-1.24.0/meson.build:84:16: ERROR: Unknown variable "python_opt".
> ```
> 
> So I checked the OE upstream and I aw they did not have any version 1.24.0, instead, they are using 1.24.5 (https://github.com/openembedded/openembedded-core/blob/master/meta/recipes-multimedia/gstreamer/gstreamer1.0-python_1.24.5.bb), therefore I thought we could use the same as the upstream (and this one compile without any issues). What is your opinion? Should we stick to 1.24.0?

There are some sets of gstreamer recipes on this layer :

1. GStreamer forked recipes
these are the ones modified for the bsp 
2. GStreamer copied recipes
these are pure copies of the oe-core upstream versions
3. GStreamer downgrade ffmpeg
these look like a requirement for the current version `1.22.0.imx` that is not available in oe-core https://github.com/Freescale/meta-freescale/commit/1c6c0f6523c0a62c15ac973aa782e108a652a61e. 
It may probably no longer be necessary but you need to check


One of the reasons is because the versions have to be the same and we cannot mix versions of different gstreamer plugins because this is not supported upstream by the gstreamer project and as far as I know it is not tested or used by anyone.

# GStreamer forked recipes

https://github.com/Freescale/meta-freescale/blob/5a46df3c737c14c1d421d64e72a101827fcb34a1/conf/machine/include/imx-base.inc#L512-L520

# GStreamer copied recipes

https://github.com/Freescale/meta-freescale/blob/5a46df3c737c14c1d421d64e72a101827fcb34a1/conf/machine/include/imx-base.inc#L522-L534

# GStreamer downgrade ffmpeg

https://github.com/Freescale/meta-freescale/blob/5a46df3c737c14c1d421d64e72a101827fcb34a1/conf/machine/include/imx-base.inc#L536-L538

You can look on the last update to better understand the process
[gstreamer: Upgrade 1.20.3.imx -> 1.22.0.imx](https://github.com/Freescale/meta-freescale/pull/1619)

hiagofranco:
> > > > > There are some inconsistency in the version used, some recipes uses the `${@get_gst_ver("${PV}")}` and others `1.24.0` and `1.24.5`
> > > > 
> > > > 
> > > > fixed, thanks!
> > > 
> > > 
> > > some still missing:
> > > https://github.com/Freescale/meta-freescale/blob/8b234355c80f338f00e5cd7e5fcdda1009897f98/conf/machine/include/imx-base.inc#L526
> > > 
> > > https://github.com/Freescale/meta-freescale/blob/8b234355c80f338f00e5cd7e5fcdda1009897f98/conf/machine/include/imx-base.inc#L532
> > 
> > 
> > Actually for gstreamer1.0-python I selected 1.24.5 on purpose, because I could not compile version 1.24.0:
> > ```shell
> > Called: `/home/frh/tdx/oe/master/build/tmp/work/armv8a-mx8-tdx-linux/gstreamer1.0-python/1.24.0.imx/recipe-sysroot-native/usr/bin/pkg-config --libs python-3.12-embed` -> 0
> > stdout:
> > -lpython3.12
> > -----------
> > Run-time dependency python found: YES 3.12
> > Message: python_abi_flags = 
> > 
> > ../gst-python-1.24.0/meson.build:84:16: ERROR: Unknown variable "python_opt".
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > So I checked the OE upstream and I aw they did not have any version 1.24.0, instead, they are using 1.24.5 (https://github.com/openembedded/openembedded-core/blob/master/meta/recipes-multimedia/gstreamer/gstreamer1.0-python_1.24.5.bb), therefore I thought we could use the same as the upstream (and this one compile without any issues). What is your opinion? Should we stick to 1.24.0?
> 
> There are some sets of gstreamer recipes on this layer :
> 
>     1. GStreamer forked recipes
>        these are the ones modified for the bsp
> 
>     2. GStreamer copied recipes
>        these are pure copies of the oe-core upstream versions
> 
>     3. GStreamer downgrade ffmpeg
>        these look like a requirement for the current version `1.22.0.imx` that is not available in oe-core [1c6c0f6](https://github.com/Freescale/meta-freescale/commit/1c6c0f6523c0a62c15ac973aa782e108a652a61e).
>        It may probably no longer be necessary but you need to check
> 
> 
> One of the reasons is because the versions have to be the same and we cannot mix versions of different gstreamer plugins because this is not supported upstream by the gstreamer project and as far as I know it is not tested or used by anyone.
> # GStreamer forked recipes
> 
> https://github.com/Freescale/meta-freescale/blob/5a46df3c737c14c1d421d64e72a101827fcb34a1/conf/machine/include/imx-base.inc#L512-L520
> 
> # GStreamer copied recipes
> 
> https://github.com/Freescale/meta-freescale/blob/5a46df3c737c14c1d421d64e72a101827fcb34a1/conf/machine/include/imx-base.inc#L522-L534
> 
> # GStreamer downgrade ffmpeg
> 
> https://github.com/Freescale/meta-freescale/blob/5a46df3c737c14c1d421d64e72a101827fcb34a1/conf/machine/include/imx-base.inc#L536-L538
> 
> You can look on the last update to better understand the process [gstreamer: Upgrade 1.20.3.imx -> 1.22.0.imx](https://github.com/Freescale/meta-freescale/pull/1619)

Thanks for the nice explanation, @quaresmajose. I was able to apply a patch to gstreamer1.0-python and use the 1.24.0 version. Now it is compiling without errors.

Please take a look, thank you!

hiagofranco:
Update: imx-gst1.0-plugin works fine now, I had to update imx-opencl-converter in order to fix the build, as one is dependent on the other. Therefore, I pushed one more commit to this PR.

0.2.0 -> 0.4.0

recipes-multimedia/ffmpeg/ffmpeg_4.4.1.bb
```diff
@@ -183,3 +183,5 @@ INSANE_SKIP:${MLPREFIX}libpostproc = "textrel"
 # Downgrade for NXP BSP
 DEFAULT_PREFERENCE = "-1"
 COMPATIBLE_MACHINE = "(imx-nxp-bsp)"
+
+LICENSE_FLAGS_ACCEPTED += " commercial "
```
thochstein, line 187: This should be removed, it was a mistake

recipes-multimedia/gstreamer/gstreamer1.0-libav_1.24.0.imx.bb
```diff
@@ -24,5 +24,13 @@ EXTRA_OEMESON += " \
...
+# Drop .imx from PV
+def get_gst_ver(v):
+    return oe.utils.trim_version(v, 3)
+
 FILES:${PN} += "${libdir}/gstreamer-1.0/*.so"
 FILES:${PN}-staticdev += "${libdir}/gstreamer-1.0/*.a"
+
+COMPATIBLE_MACHINE = "(imx-nxp-bsp)"
+
+LICENSE_FLAGS_ACCEPTED += " commercial "
```
thochstein, line 36: Remove

recipes-multimedia/gstreamer/gstreamer1.0-plugins-base_1.24.0.imx.bb
```diff
@@ -116,18 +116,36 @@ SRC_URI:remove = " \
...
+# does not depend on the viv-fb feature. It used to, but that was actually a bug
+# which was fixed in GStreamer 1.22.5. Since then, the direct texture support is
+# detected by Meson by checking for direct texture symbols like "glTexDirectVIV".)
+PACKAGECONFIG_GL:imxgpu2d = \
+    "${@bb.utils.contains('DISTRO_FEATURES', 'opengl x11', 'opengl', '', d)}"
+PACKAGECONFIG_GL:imxgpu3d = \
+    "${@bb.utils.contains('DISTRO_FEATURES', 'opengl', 'gles2 egl', '', d)}"
+PACKAGECONFIG_GL:use-mainline-bsp = \
+    "${@bb.utils.contains('DISTRO_FEATURES', 'opengl', 'gles2 egl gbm', '', d)}"
+
```
thochstein, line 142: This addition and the next addition are not needed, as they are in the `.bbappend`

recipes-multimedia/gstreamer/gstreamer1.0-plugins-base_1.24.0.imx.bb
```diff
@@ -116,18 +116,36 @@ SRC_URI:remove = " \
...
 PACKAGECONFIG:remove = "${PACKAGECONFIG_REMOVE}"
 PACKAGECONFIG_REMOVE ?= "jpeg"
 
+PACKAGECONFIG_GL:imxgpu2d:append:mx6-nxp-bsp = " viv-fb"
+PACKAGECONFIG_GL:imxgpu2d:append:mx7-nxp-bsp = " viv-fb"
```
thochstein, line 147: Actually, you could take this line over to the bbappend, making the override generic for all i.MX 7

recipes-multimedia/ffmpeg/ffmpeg_4.4.1.bb
hiagofranco: Removed. Should I also remove it from `gstreamer1.0-plugins-ugly`?

recipes-multimedia/gstreamer/gstreamer1.0-libav_1.24.0.imx.bb
hiagofranco: Removed.

recipes-multimedia/gstreamer/gstreamer1.0-plugins-base_1.24.0.imx.bb
hiagofranco: All right, removed it.

recipes-multimedia/gstreamer/gstreamer1.0-plugins-base_1.24.0.imx.bb
hiagofranco: I moved it, but without the `imxgpu2d` override because the last one, version 1.22, did not have it. Should I keep it?

recipes-multimedia/ffmpeg/ffmpeg_4.4.1.bb
thochstein: Yes, I didn't notice that one.

recipes-multimedia/gstreamer/gstreamer1.0-plugins-base_1.24.0.imx.bb
thochstein: Hmm, I think it needs the `imxgpu` override.

hiagofranco: All right, I added them, thanks.

recipes-multimedia/ffmpeg/ffmpeg_4.4.1.bb
hiagofranco: Ok, removed, thank you

thochstein (approved):
LGTM, thanks @hiagofranco and @quaresmajose!

recipes-multimedia/gstreamer/gstreamer1.0-plugins-base_1.24.%.bbappend
```diff
@@ -13,5 +13,5 @@ PACKAGECONFIG_GL:use-mainline-bsp = \
 # does not depend on the viv-fb feature. It used to, but that was actually a bug
 # which was fixed in GStreamer 1.22.5. Since then, the direct texture support is
 # detected by Meson by checking for direct texture symbols like "glTexDirectVIV".)
-PACKAGECONFIG_GL:append:mx6-nxp-bsp = " viv-fb "
-PACKAGECONFIG_GL:append:mx7ulp-nxp-bsp = " viv-fb "
+PACKAGECONFIG_GL:imxgpu2d:append:mx6-nxp-bsp = " viv-fb "
+PACKAGECONFIG_GL:imxgpu2d:append:mx7-nxp-bsp = " viv-fb "
```
quaresmajose, line 17: I think this not works, the `append` needs to be the first override in the list. Can you please verify the `PACKAGECONFIG_GL` on any of the referred machines ?

thochstein: It is necessary, these are appending to the assignments on lines 1-4.

Perhaps it could be clearer by dropping the `append` and adding a variable?

```
PACKAGECONFIG_GL:imxgpu2d = " \
    ${@bb.utils.contains('DISTRO_FEATURES', 'opengl x11', 'opengl', '', d)} \
    ${PACKAGECONFIG_GL_VIV} \
"
PACKAGECONFIG_GL:imxgpu3d = " \
    ${@bb.utils.contains('DISTRO_FEATURES', 'opengl', 'gles2 egl', '', d)} \
    ${PACKAGECONFIG_GL_VIV} \
"
PACKAGECONFIG_GL:use-mainline-bsp = \
    "${@bb.utils.contains('DISTRO_FEATURES', 'opengl', 'gles2 egl gbm', '', d)}"

# The i.MX8 uses KMS instead of the Vivante specific framebuffer API.
# The i.MX7 does not have a GPU, except for ULP.
# This leaves the i.MX6 - with the vendor BSP - as the remaining use case for viv-fb.
#
# (Note that viv-fb is about the _windowing system_. Vivante direct texture support
# does not depend on the viv-fb feature. It used to, but that was actually a bug
# which was fixed in GStreamer 1.22.5. Since then, the direct texture support is
# detected by Meson by checking for direct texture symbols like "glTexDirectVIV".)
PACKAGECONFIG_GL_VIV             = ""
PACKAGECONFIG_GL_VIV:mx6-nxp-bsp = " viv-fb "
PACKAGECONFIG_GL_VIV:mx7-nxp-bsp = " viv-fb "
```

quaresmajose: My concern is just about the `append` positions, I think it should be
```patch
-PACKAGECONFIG_GL:imxgpu2d:append:mx6-nxp-bsp = " viv-fb "
-PACKAGECONFIG_GL:imxgpu2d:append:mx7-nxp-bsp = " viv-fb "
+PACKAGECONFIG_GL:append:imxgpu2d:mx6-nxp-bsp = " viv-fb "
+PACKAGECONFIG_GL:append:imxgpu2d:mx7-nxp-bsp = " viv-fb "
```
but have tested locally and it works in the two cases. Sorry for the noise

PR #1878, opened 2024-07-26
imx-mkimage: Update 6.6.3-1.0.0 to 6.6.23-2.0.0

hiagofranco:
Update recipes to NXP BSP version 6.6.23-2.0.0.
This commit introduces the 'imx95' target.

PR #1880, opened 2024-07-27
recipes-bsp: introduce imx-oei recipe

hiagofranco:
OEI stands for "Optional Executable Image", it was relased for NXP BSP 6.6.23-2.0.0. The OEI Firmware is required for the i.MX 95 SoC.


hiagofranco:
Relates to #1878 and #1861.

PR #1898, opened 2024-08-19
weston-init: Fix RDP support

thochstein:
The modification of the weston.ini command line necessary for RDP support is missing. The problem is the weston.ini template contains `@bindir@`, while the sed operation for the modification contains `${bindir}`.

Fix the problem by moving the replacement of `@bindir@` to the top of the function `do_install:append()`.


thochstein:
Thanks, @otavio. Can you backport to scarthgap?

otavio:
It seems it does not apply cleanly. Could you prepare the backport manually?

PR #1904, opened 2024-08-22
firmware-ele-imx: Bump to version 0.1.2

hiagofranco:
This aligns with the current version available on meta-imx layer.

Related-to: #1861


hiagofranco:
@otavio I believe with this change the "Upgrade firmware-imx*" task from 1861 should be done.

PR #1910, opened 2024-08-28
imx-oei: move to dynamic-layers and add arm-toolchain

hiagofranco:
The recipe depends on meta-arm-toolchain, therefore it should be on
dynamic-layers/arm-toolchain folder. Create it and make the
necessary changes for conf/layer.conf

Fixes: 61beaf5861f2 ("recipes-bsp: introduce imx-oei recipe")


otavio (changes requested):
I believe we can make this recipe conditional to the toolchain layer instead of the full arm layer. Could you give it a try?

https://git.yoctoproject.org/meta-arm/tree/meta-arm/conf/layer.conf

Either way, I marked the recipe to do the backport automatically, so once we merge this pull request, it's going to create automatically the backport pull request for us.

otavio:
@thochstein, could you ask internally why we need to use the arm toolchain instead of the general one? I'd prefer to drop that dependency and make the recipe general.

That would be done in a future release, but it would be nice to consider this improvement.

hiagofranco:
> I believe we can make this recipe conditional to the toolchain layer instead of the full arm layer. Could you give it a try?
> 
> https://git.yoctoproject.org/meta-arm/tree/meta-arm/conf/layer.conf
> 
> Either way, I marked the recipe to do the backport automatically, so once we merge this pull request, it's going to create automatically the backport pull request for us.

I am not sure if I got it correct, but I pushed the changes. Please have a look.

thochstein:
> @thochstein, could you ask internally why we need to use the arm toolchain instead of the general one? I'd prefer to drop that dependency and make the recipe general.
> 
> That would be done in a future release, but it would be nice to consider this improvement.

We need the generic compiler because the output runs on an m-core, not the main core.

conf/layer.conf
```diff
@@ -52,6 +52,8 @@ BBFILES_DYNAMIC += " \
     \
     meta-arm:${LAYERDIR}/dynamic-layers/meta-arm/*/*/*.bbappend \
     \
+    meta-arm-toolchain:${LAYERDIR}/dynamic-layers/meta-arm-toolchain/*/*/*.bb \
```
otavio, line 55: I believe the layer collection uses a different name. Have you tested that?

```suggestion
    arm-toolchain:${LAYERDIR}/dynamic-layers/arm-toolchain/*/*/*.bb \
```

My understanding is that you should be able to build the recipe just with the `meta-arm-toolchain` layer included without the `meta-arm` layer being listed in the `bblayers.conf`.

hiagofranco: Thanks @otavio. I pushed these changes now, I tested with 

```
MACHINE=imx95-19x19-verdin bitbake imx-oei
```

from meta-imx-bsp, using this patch inside meta-freescale and the build passes.

otavio: I believe that the file is in the wrong place, my understanding is that it should be located here:

```
├── arm-toolchain
│   └── recipes-bsp
│       └── imx-oei
│           └── imx-oei_1.0.0.bb
```

This is because of the collection of the layer. See:

https://git.yoctoproject.org/meta-arm/tree/meta-arm-toolchain/conf/layer.conf#n7

I suspect you have the previous file in some location that could be found. Otherwise, the build would fail.

Could you double check that? In worst case, we could jump in a short call.

hiagofranco: Hi @otavio, sorry for the delay. Now I moved to the correct folder and did a quick build again, and it looks like it is working (I did a "cleanall" this time). Can you please take a look? Thank you!

PR #1919, opened 2024-09-04
imx-secure-enclave: add the lf-6.6.3-1.0.0 recipe

tprrt:
This library allows the use of the EdgeLock secure Enclave (ELE) from Linux userspace.

Moreover, this library is a build dependency of i.MX Security Middleware Library.

_Note. This recipe has been also tested on Scarthgap and can be cherry pick on the Scarthgap branch to fix the missing build dependency with SMW._


otavio:
Sure, I will apply this, but is it possible for you to send the newer version as well?

PR #1938, opened 2024-09-12
gstreamer1.0-plugins-bad: Add imxvideoconvert_g2d to autovideoconvert

hiagofranco:
Add patch to enable imxvideoconvert on autovideoconvert. By default, imxvideoconvert_g2d is not included as a potential filter, resulting in poor video playback performance when autovideoconvert is used.

Add imxvideoconvert_g2d to improve performance by utilizing hardware acceleration.


hiagofranco:
Any feedbacks are welcome =D

hiagofranco:
And this we saw happening with Scarthgap, so a backport to scarthgap would be required

thochstein:
> Any feedbacks are welcome =D

The multimedia team says:

> We have supported this in Q3 release.
> autovideoconvert: select the i.MX hardware based video convert plugins or other video convert plugins according to the input and output format.

PR #1950, opened 2024-09-20
kernel-module-isp-vvcam: Update 6.6.3-1.0.0 to 6.6.23-2.0.0

hiagofranco:
Update to the new NXP BSP 6.6.23-2.0.0.


hiagofranco:
This would also go into Scarthgap, thanks!

PR #1954, opened 2024-09-27
[master] linux-fslc: 6.6.50 -> 6.11.0

Livius90:
- Update linux-fslc: 6.6.50 -> 6.11.0

### DIFF
--- a/recipes-kernel/linux/linux-fslc-lts_6.1.bb
+++ b/recipes-kernel/linux/linux-fslc-lts_6.1.bb
@@ -1,0 +1,41 @@
+# Copyright (C) 2012-2019 O.S. Systems Software LTDA.
+# Released under the MIT license (see COPYING.MIT for the terms)
+#
+# SPDX-License-Identifier: MIT
+#
+
+SUMMARY = "FSL Community BSP Linux LTS based kernel with backported features and fixes"
+DESCRIPTION = "Linux kernel based on LTS kernel used by FSL Community BSP in order to \
+provide support for some backported features and fixes, or because it was applied in upstream \
+and will not become part of a LTS version, or because it is not applicable for \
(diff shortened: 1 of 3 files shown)


otavio:
Thank you for looking at this. Please keep the 6.6 kernel but rename it to LTS, as we did for the 6.1.

The 6.11 kernel should be a new recipe without the LTS suffix, so the flavor they choose is transparent to the user.

While you are on it, if you can update the 6.1 and 6.6 recipes to the latest version, that would be great. We can do the same when doing the backports. I have updated the respective branches in our repository.

Livius90:
@otavio 
Do you like to keep `linux-fslc-lts_6.1.bb` also?

otavio:
I think that for now it's important to keep as it is maintained and there might be users of it. Maybe when the next LTS is released, then we can drop it. What do you think?

Livius90:
In master branch `linux-fslc-lts_6.1.bb` is useless for future, because now the latest official LTS is v6.6.x which available as an LTS, `linux-fslc-lts_6.6.bb`. To keep `linux-fslc-lts_6.1.bb` can be important only in `scarthgap`, because it is possible that somebody still use it in a released `scarthgap` layer.

https://www.kernel.org/

otavio:
You have a valid point. So, move to 6.1 to 6.6 LTS in the master branch, and keep the 6.1 version when preparing the backport.

Livius90:
> You have a valid point. So, move to 6.1 to 6.6 LTS in the master branch, and keep the 6.1 version when preparing the backport.

Done

otavio:
Please prepare the pull request for backporting and then we can remove this 6.1 here in master branch.

Livius90:
I think you can use the Backport CI action and it will generate this PR automatically.

PR #1955, opened 2024-09-27
cryptodev-module: Add patch fix for prepare kernel v6.11.0 in scarthgap

Livius90:
Fix build for Linux 6.7-rc1: https://github.com/cryptodev-linux/cryptodev-linux/commit/5e7121e45ff283d30097da381fd7e97c4bb61364
Since Linux 6.7-rc1, no ahash algorithms set a nonzero alignmask, and therefore `crypto_ahash_alignmask` has been removed.

See also:
- https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=0f8660c82b79af595b056f6b9f4f227edeb88574
- https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=c626910f3f1bbce6ad18bc613d895d2a089ed95e

### DIFF
--- a/recipes-kernel/cryptodev/cryptodev-module_%.bbappend
+++ b/recipes-kernel/cryptodev/cryptodev-module_%.bbappend
@@ -1,0 +1,12 @@
+FILESEXTRAPATHS:prepend := "${THISDIR}/files:"
+
+# Fix build for Linux 6.7-rc1: https://github.com/cryptodev-linux/cryptodev-linux/commit/5e7121e45ff283d30097da381fd7e97c4bb61364
+# Since Linux 6.7-rc1, no ahash algorithms set a nonzero alignmask,
+# and therefore `crypto_ahash_alignmask` has been removed.
+#
+# See also: https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=0f8660c82b79af595b056f6b9f4f227edeb88574
+#           https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=c626910f3f1bbce6ad18bc613d895d2a089ed95e
+
+SRC_URI += " \
+    file://fix-build-for-Linux-6.7-rc1.patch \
+"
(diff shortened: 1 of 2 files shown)


recipes-kernel/cryptodev/cryptodev-module_%.bbappend
```diff
@@ -0,0 +1,12 @@
...
+
+# Fix build for Linux 6.7-rc1: https://github.com/cryptodev-linux/cryptodev-linux/commit/5e7121e45ff283d30097da381fd7e97c4bb61364
+# Since Linux 6.7-rc1, no ahash algorithms set a nonzero alignmask,
+# and therefore `crypto_ahash_alignmask` has been removed.
+#
+# See also: https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=0f8660c82b79af595b056f6b9f4f227edeb88574
+#           https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=c626910f3f1bbce6ad18bc613d895d2a089ed95e
+
+SRC_URI += " \
+    file://fix-build-for-Linux-6.7-rc1.patch \
```
otavio, line 11: Is it possible for you to prepare this improvement to open-embedded core?

The reason is this is not something that we should carry on, mostly because if they do any change on their side, eventually we are going to have a conflict.

In the time being, one thing that we could do is to apply this change only for the IMX generic BSP override.

otavio (changes requested):

recipes-kernel/cryptodev/cryptodev-module_%.bbappend
Livius90: I started to use IMX generic BSP override because scarthgap is designed to use kernel v6.6.X originaly. It means oe maintainers will ignore to use totally this patch in their layer, sure.

Livius90:
This patch need only, for cryptodev-module v1.13 if someone like to use it for any new Linux kernel, v6.7.x or higher.
It seems, it was already applied in an other i.MX BSP meta-layers (Phytec) also: https://git.phytec.de/meta-ampliphy/tree/recipes-kernel/cryptodev/cryptodev-module_1.13.bbappend?h=scarthgap

PR #1970, opened 2024-10-11
cryptodev-module: Drop redundant patch

thochstein:
The recipe is updated to 1.14 so the backported patch is no longer needed.


Livius90:
https://layers.openembedded.org/layerindex/recipe/397437/

Yes, it seems cryptodev-module has been updated to v1.14 in scarthgap also. Thanks that info, not need this patch anymore.

otavio:
Could someone prepare a backport?

thochstein:
This PR was for `scarthgap`. I don't see the problem on `master`.

PR #2001, opened 2024-10-25
Upgrade NXP Graphics BSP to LF6.6.36_2.1.0

valijec:
This fixes #1953

Issue #1953, opened 2024-09-26
Upgrade BSP to LF6.6.36_2.1.0


otavio:
The immediate items to work on are:

- [x]  Upgrade Linux kernel and headers (i.MX variants)
  - [x] #1958 
  - [x] #1971
  - [x] #1991 
- [x]  Upgrade Linux kernel (FSLC forks i.MX variants)
- [x]  Upgrade U-Boot (FSLC forks i.MX variants)
- [x]  Upgrade OP-TEE (i.MX variants)
  - [x] #1981  
- [x]  Upgrade ATF (i.MX variants)
  - [x] #1982 
- [ ]  Upgrade Multimedia Packages
  - [x] #2002 
- [x]  Upgrade Graphics Packages
  - [x] #2001 
- [x]  Upgrade Jailhouse 
  - [ ] #2003 
- [x]  Upgrade mcore-demos 
  - [ ] #2003 
- [x]  Upgrade firmware-imx*
- [x]  Upgrade imx-mkimage
  - [ ] #2003 

The Release Notes are [here](https://www.nxp.com/docs/en/release-note/IMX_LINUX_RELEASE_NOTES.pdf) for checking.

PR #2002, opened 2024-10-27
Upgrade recipes multimedia and imx-lib to BSP LF6.6.36_2.1.0

hiagofranco:
Related-to: #1953

Needs #2001 in order to build imx-gst1.0-plugin without errors.


hiagofranco:
@thochstein are you the author of the patch I am removing from imx-gst1.0-plugin? If so, can you confirm this is indeed not needed anymore? From my understanding this should be removed now.

Reference: https://github.com/nxp-imx/imx-gst1.0-plugin/commit/74a6afa6cceb69fe4b8cf09690392093cc7c5b37

Thanks

thochstein:
> @thochstein are you the author of the patch I am removing from imx-gst1.0-plugin? If so, can you confirm this is indeed not needed anymore? From my understanding this should be removed now.

Yes, the patch should be removed.

thochstein (approved):

PR #2009, opened 2024-10-29
imx-boot: Fix 8M non multi-config build problem

esben:
The fix made in commit 2db7047ba40e
("imx-boot: Fix 8M multi-config build problems")
broke builds not using U-Boot multi-config, as the link created ends up being a simple recursive link when UBOOT_DTB_NAME_EXTRA is the same as UBOOT_DTB_NAME.

It fails with something like this:
```
| ./../scripts/dtb_check.sh imx8mq-evk.dtb evk.dtb imx8mq-var-dart-dt8mcustomboard.dtb
|  Can't find u-boot DTB file, please copy from u-boot
```

caused by a symlink like this:
```
lrwxrwxrwx 1 1000 1000 35 Oct 29 21:32 imx8mq-var-dart-dt8mcustomboard.dtb -> imx8mq-var-dart-dt8mcustomboard.dtb
```

Fixes: 2db7047ba40e ("imx-boot: Fix 8M multi-config build problems")


esben:
Backport to scarthgap branch is needed.

PR #2029, opened 2024-11-12
Update imx8mm-ddr4-evk.conf

vlaarco:
When changing LPDDR4 to DDR4 in imx8mmevk.conf the build fails.

For example:

MACHINEOVERRIDES =. "imx8mm-lpddr4-evk:"
require conf/machine/imx8mm-lpddr4-evk.conf

to

MACHINEOVERRIDES =. "imx8mm-ddr4-evk:"
require conf/machine/imx8mm-ddr4-evk.conf

### DIFF
--- a/conf/machine/imx8mm-ddr4-evk.conf
+++ b/conf/machine/imx8mm-ddr4-evk.conf
@@ -6,7 +6,7 @@
 
 require include/imx8mm-evk.inc
 
-KERNEL_DEVICETREE_BASENAME = "${MACHINE}"
+KERNEL_DEVICETREE_BASENAME = "imx8mm-ddr4-evk"
 KERNEL_DEVICETREE:append:use-nxp-bsp = " \
     freescale/${KERNEL_DEVICETREE_BASENAME}-revb.dtb \
     freescale/${KERNEL_DEVICETREE_BASENAME}-revb-rm67191.dtb \


thochstein:
Hi @vlaarco. I suspect that you should just use `imx8mm-lpddr4-evk` and `imx8mm-ddr4-evk` directly. Changing `imx8mmevk.conf` as you describe is not something we do or recommend.

https://github.com/nxp-imx/meta-imx/blob/scarthgap-6.6.36-2.1.0/meta-imx-bsp/conf/machine/imx8mmevk.conf

conf/machine/imx8mm-ddr4-evk.conf
```diff
@@ -6,7 +6,7 @@
 
 require include/imx8mm-evk.inc
 
-KERNEL_DEVICETREE_BASENAME = "${MACHINE}"
+KERNEL_DEVICETREE_BASENAME = "imx8mm-ddr4-evk"
```
otavio, line 9: Please adjust the commit lock to use the machine name as prefix and use the message that you sent and the pull request as the commit message for more detail

otavio:
> Hi @vlaarco. I suspect that you should just use `imx8mm-lpddr4-evk` and `imx8mm-ddr4-evk` directly. Changing `imx8mmevk.conf` as you describe is not something we do or recommend.
> 
> https://github.com/nxp-imx/meta-imx/blob/scarthgap-6.6.36-2.1.0/meta-imx-bsp/conf/machine/imx8mmevk.conf

Is my understanding correct that this issue is not to be fixed here? @thochstein

PR #2044, opened 2024-12-05
Add workflow to automatically update LICENSE file with recipe licenses

hellhound94:
Originally from meta-freescale-3rdparty.

### DIFF
--- a/.github/workflows/update-license-file.yml
+++ b/.github/workflows/update-license-file.yml
@@ -1,0 +1,35 @@
+name: Update LICENSE file
+
+on:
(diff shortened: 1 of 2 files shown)


.github/workflows/update-license-file.yml
```diff
@@ -0,0 +1,35 @@
+name: Update LICENSE file
```
otavio, line 1: Please update the commit message because I think that you made a small mistake. Something like:

Add workflow to automatically update LICENSE file with recipe licenses

PR #2046, opened 2024-12-09
linux-fslc: 6.12.1 -> 6.12.4

Livius90:
- Update linux-fslc to v6.12.4 kernel


Livius90:
@otavio 
Can you backport it to `scarthgap` also?

PR #2056, opened 2025-01-07
mcore-demos: Upgrade to 6.6.52_2.2.0

zelan-nxp:
imx95 mocre: 2.16.001 -> 24.12.00
Install elf binary to /usr/lib/firmware/

### DIFF
--- a/recipes-fsl/mcore-demos/imx-m4-demos_1.0.1.bb
+++ b/recipes-fsl/mcore-demos/imx-m4-demos_1.0.1.bb
@@ -1,27 +1,13 @@
-SUMMARY = "i.MX M4 core Demo images"
-SECTION = "app"
-LICENSE = "Proprietary"
+# Copyright 2017-2021 NXP
+# Released under the MIT license (see COPYING.MIT for the terms)
+
+require imx-mcore-demos.inc
+
 LIC_FILES_CHKSUM:mx7d-nxp-bsp = "file://COPYING;md5=8cf95184c220e247b9917e7244124c5a"
 
-inherit deploy fsl-eula-unpack
-
-M4_SOC ?= "INVALID"
-M4_SOC:mx7d-nxp-bsp = "imx7d-sabresd"
-
-SRC_URI = "${FSL_MIRROR}/${M4_SOC}-m4-freertos-${PV}.bin;fsl-eula=true"
-S = "${WORKDIR}/${M4_SOC}-m4-freertos-${PV}"
+SRC_URI = "${FSL_MIRROR}/${SOC}-m4-freertos-${PV}.bin;fsl-eula=true"
+S = "${WORKDIR}/${SOC}-m4-freertos-${PV}"
 
 SRC_URI[sha256sum] = "cc00d3b936d49b2794a2a99e10129437e70caba3fd26b8379b8c50dd22f73254"
 
-do_deploy () {
-   # Install the demo binaries
-   install -d ${DEPLOYDIR}
-   cp ${S}/*.bin ${DEPLOYDIR}/
-   ls ${DEPLOYDIR}/
-}
-
-addtask deploy before do_build after do_compile
-
-PACKAGE_ARCH = "${MACHINE_SOCARCH}"
 COMPATIBLE_MACHINE = "(mx7d-nxp-bsp)"
-
--- a/recipes-fsl/mcore-demos/imx-mcore-demos.inc
+++ b/recipes-fsl/mcore-demos/imx-mcore-demos.inc
@@ -9,6 +9,7 @@ inherit deploy fsl-eula-unpack
 
 SOC                ?= "INVALID"
 SOC:mx7ulp-nxp-bsp  = "imx7ulp"
+SOC:mx7d-nxp-bsp    = "imx7d-sabresd"
 SOC:mx8dxl-nxp-bsp  = "imx8dxl"
 SOC:mx8mm-nxp-bsp   = "imx8mm"
 SOC:mx8mn-nxp-bsp   = "imx8mn"
@@ -32,26 +33,25 @@ MCORE_TYPE:mx8ulp-nxp-bsp  = "m33"
 MCORE_TYPE:mx93-nxp-bsp  = "m33"
 MCORE_TYPE:mx95-nxp-bsp  = "m7"
 
-SRC_URI = "${FSL_MIRROR}${SOC}-${MCORE_TYPE}-demo-${PV}.bin;name=${SOC};fsl-eula=true"
+SRC_URI = "${FSL_MIRROR}/${SOC}-${MCORE_TYPE}-demo-${PV}.bin;name=${SOC};fsl-eula=true"
 
 S = "${WORKDIR}/${SOC}-${MCORE_TYPE}-demo-${PV}"
-
-#SRC_URI="https://www.nxp.com/lgfiles/NMG/MAD/YOCTO//imx93-m33-demo-2.14.0.bin;fsl-eula=true;name=imx93"
-
 
 SCR = "SCR-${SOC}-${MCORE_TYPE}-demo.txt"
 
 do_install () {
     install -d ${D}${nonarch_base_libdir}/firmware
     install -m 0644 ${S}/*.elf ${D}${nonarch_base_libdir}/firmware
+    install -m 0644 ${S}/*.bin ${D}${nonarch_base_libdir}/firmware
 }
 
 DEPLOY_FILE_EXT       ?= "bin"
 DEPLOY_FILE_EXT:mx7ulp-nxp-bsp = "img"
 
 do_deploy () {
-   # Install the demo binaries
-   install -m 0644 ${S}/*.${DEPLOY_FILE_EXT} ${DEPLOYDIR}/
+    # Install the demo binaries
+    install -d ${DEPLOYDIR}/mcore-demos
+    install -m 0644 ${S}/*.${DEPLOY_FILE_EXT} ${DEPLOYDIR}/mcore-demos/
 }
 
 addtask deploy after do_install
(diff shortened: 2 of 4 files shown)


otavio (changes requested):
The changes look good, but please improve the commit log.

otavio (changes requested):

recipes-fsl/mcore-demos/imx-mcore-demos.inc
```diff
@@ -5,7 +5,7 @@ SUMMARY = "i.MX M4/M7/M33 core Demo images"
 SECTION = "app"
 LICENSE = "Proprietary"
 
-inherit deploy fsl-eula-unpack
+inherit deploy fsl-eula2-unpack2
```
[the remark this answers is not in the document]
thochstein: @zelan-nxp, please remove this change, it's not meant to be upstreamed.

I do need to clean this up on our side.

otavio:
Please rebase this pull request because I made the change to remove the checksum across all the files to a single commit. So you will need also to remove this from this and other pull requests.

recipes-fsl/mcore-demos/imx-mcore-demos.inc
```diff
@@ -50,14 +48,11 @@ DEPLOY_FILE_EXT       ?= "bin"
...
+    install -m 0644 ${S}/*.${DEPLOY_FILE_EXT} ${DEPLOYDIR}/mcore-demos/
 }
 
 addtask deploy after do_install
 
 PACKAGE_ARCH = "${MACHINE_SOCARCH}"
-
-FILES:${PN} = "${nonarch_base_libdir}/firmware"
-
-INSANE_SKIP:${PN} = "arch"
```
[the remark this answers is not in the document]
zelan-nxp: Yes, this change need to upstream fsl-eula2-unpack2.bbclasee. I removed this change.

otavio (approved):

thochstein (approved):
Changes are fine, but it does seem odd that the 1.0.1 recipe doesn't use the .inc file.

zelan-nxp:
> Changes are fine, but it does seem odd that the 1.0.1 recipe doesn't use the .inc file.
Good catch. imx7d m4 demos binary name is `imx7d-sabresd-m4-freertos-1.0.1.bin`, no elf in package, Which special case than others. I  update 1.0.1 recipe to use inc file and overwrite SRC_URI and let don't do install task.

recipes-fsl/mcore-demos/imx-m4-demos_1.0.1.bb
```diff
@@ -1,27 +1,15 @@
...
+require imx-mcore-demos.inc
 
-M4_SOC ?= "INVALID"
-M4_SOC:mx7d-nxp-bsp = "imx7d-sabresd"
+LIC_FILES_CHKSUM:mx7d-nxp-bsp = "file://COPYING;md5=8cf95184c220e247b9917e7244124c5a"
 
-SRC_URI = "${FSL_MIRROR}/${M4_SOC}-m4-freertos-${PV}.bin;fsl-eula=true"
-S = "${WORKDIR}/${M4_SOC}-m4-freertos-${PV}"
+SRC_URI = "${FSL_MIRROR}/${SOC}-m4-freertos-${PV}.bin;fsl-eula=true"
+S = "${WORKDIR}/${SOC}-m4-freertos-${PV}.bin"
```
otavio, line 9: Have you tried that? Because this change seems completely wrong.

Also, there is no package being generated, so it is just not generating the package to allow it to be deployed.

zelan-nxp: Sorry, The `S `directory was typo to `.bin` Updated.
Yes, imx7d mcore package no elf，So no need install them to /lib/firmware

otavio: Why not? How would I load the firmware during the runtime of the system?

zelan-nxp: These firmware was defined `WKS_FILE_DEPENDS`, they are installed in` /run/media/boot-mmcblk1p1/mcore-demos` when bootup. Although imx7d mcore no elf package, the bin package also generated and put in DEPLOYDIR.

otavio: I believe it is wrong. We should allow it to be installed as the user may want to load it manually.

thochstein: Hi @otavio. I'm not an expert, but maybe since these are images to run on the mcore cores in the SOC, it doesn't make sense to load them manually like regular firmware? Note also that the files that are installed are not the same as the files that are deployed. The former is always `.elf`, while the latter is `.bin` or `.img`. As Zelan said, there are no `.elf` files in the archive for 1.0.1.

angolini: as far as I remember (I might be wrong) imx7D does not provide a way to boot only the M4 core. That's why maybe the "one would not load it manually" would not be the default.

However, "load it manually" might mean configure the u-boot to load it during a boot script, so i  don't see how not having  it installed in a partition. I mean, after SPL set everything, there is no reason not boot the M4 core

angolini: @zelan-nxp could you, please, double check what were the files provided by the previous version of this package?
Also, is there a list of which SOC should use which mcore-demos version?

otavio: @zelan-nxp @thochstein Have you seen the comments provided by @angolini?

I see no reason why the files wouldn't be loaded by the rootfs, either using the u-boot bootloader or even from the Linux.

Is there any real reason why the files are not copied into the package to be installed inside the rootfs?

thochstein: I'm not sure how it works. Most of the archives install a `.elf` to the rootfs and a `.bin` to the .wic. For this recipe, there is no `.elf`. Should the `.bin` go in the rootfs? For all SOCs?

zelan-nxp: > @zelan-nxp could you, please, double check what were the files provided by the previous version of this package? Also, is there a list of which SOC should use which mcore-demos version?

Yes, imx7d mcore keep 1.0.0 version for a long time, the previous release we also install them like this ( written in boot partition). Other SOC mcore version you can check[ ([https://github.com/Freescale/meta-freescale/blob/master/recipes-fsl/mcore-demos/README])](https://github.com/Freescale/meta-freescale/blob/master/recipes-fsl/mcore-demos/README)

zelan-nxp: > I'm not sure how it works. Most of the archives install a `.elf` to the rootfs and a `.bin` to the `.wic`. For this recipe, there is no `.elf`. Should the `.bin` go in the rootfs? For all SOCs?

Mcore `.elf` is only used for developer debug, Generally, `.elf` shouldn't be install into rootfs for all SOC ,but developer require to debug their demo packages so we install them in the later release. Mcore pakcage is some feature(like audio)  simple validation, they should write in correct core address, I think its's non-sense to install` .bin` package into standalone rootfs.

zelan-nxp: I agree. Install main package is reasonable even though no .elf supply. I will pull it in layer.conf

thochstein: Thanks, @zelan-nxp, for adding it for 7d, but it seems to me that this is a general feature request, and the firmware should be installed for the other SOCs as well.

Please confirm, @otavio.

recipes-fsl/mcore-demos/imx-mcore-demos.inc
```diff
@@ -32,26 +33,26 @@ MCORE_TYPE:mx8ulp-nxp-bsp  = "m33"
...
 
 S = "${WORKDIR}/${SOC}-${MCORE_TYPE}-demo-${PV}"
 
-#SRC_URI="https://www.nxp.com/lgfiles/NMG/MAD/YOCTO//imx93-m33-demo-2.14.0.bin;fsl-eula=true;name=imx93"
-
-
 SCR = "SCR-${SOC}-${MCORE_TYPE}-demo.txt"
 
 do_install () {
+    # install mcores binary to /lib/firmware
```
thochstein, line 43: This comment is actually wrong, as it installs to /usr/lib/firmware. It's also obvious, so better just to drop the comment

conf/machine/include/imx-base.inc
```diff
@@ -404,6 +404,8 @@ SIGGEN_EXCLUDE_SAFE_RECIPE_DEPS:append = " \
     imx-codec->virtual/imxvpu \
     imx-test->virtual/imxvpu \
 "
+# Mcore installed
+MACHINE_EXTRA_RRECOMMENDS:append:mx7d-nxp-bsp  = " imx-m4-demos"
```
otavio, line 408: I don't think that we need to install that by default, as many users do not use the M4 MCU together with the image. So, my suggestion is just to drop this line.

zelan-nxp: Okay, It's fine that users can customize install "imx-m4-demos" main package if they wanted.

recipes-fsl/mcore-demos/imx-mcore-demos.inc
```diff
@@ -32,26 +33,25 @@ MCORE_TYPE:mx8ulp-nxp-bsp  = "m33"
...
 
-#SRC_URI="https://www.nxp.com/lgfiles/NMG/MAD/YOCTO//imx93-m33-demo-2.14.0.bin;fsl-eula=true;name=imx93"
-
-
 SCR = "SCR-${SOC}-${MCORE_TYPE}-demo.txt"
 
 do_install () {
     install -d ${D}${nonarch_base_libdir}/firmware
     install -m 0644 ${S}/*.elf ${D}${nonarch_base_libdir}/firmware
+    install -m 0644 ${S}/*.bin ${D}${nonarch_base_libdir}/firmware
```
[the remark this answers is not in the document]
thochstein: Ugh, I made that comment yesterday, but guess it was in a review that I still had to mark complete

PR #2061, opened 2025-01-09
Upstream from NXP-6.6.52-2.0.0

flora2086:


angolini:
I would also like to see more information on the commit logs

angolini:
I would add a description on the PR. Something like:

"Update the recipes restool, qoriq-cst, dce, qe-ucode to match the  NXP-6.6.52-2.0.0 release for xxxxxx machine. Tested by building yyyyyyyy"

angolini (approved):
We usually add the changelog from one commit id to the other, when it's a recipe bump. But I don't know those recipes, I don't know if this makes sense.

Other than my other comment, I think there is nothing else to comment.

junzhuimx (approved):

PR #2062, opened 2025-01-09
linux-fslc-imx: bump to add one bugfix

angolini:


otavio:
I believe that you used the wrong revision, because you used the original one before the pull request is measured. Below is the final revision that I would prefer you to use: 
`657504d5644f24a41822960ed31b883062ca30c9`.

angolini:
yes, i did! Thanks

PR #2077, opened 2025-01-22
imx-atf: Adjust for compiler virtual renames in oe-core

kraj:


recipes-bsp/imx-atf/imx-atf_2.10.bb
```diff
@@ -34,7 +34,7 @@ LD[unexport] = "1"
 
 # Baremetal, just need a compiler
 INHIBIT_DEFAULT_DEPS = "1"
-DEPENDS = "virtual/${HOST_PREFIX}gcc"
+DEPENDS = "virtual/cross-cc"
```
thochstein, line 37: I believe that this is a breaking change that will require the creation of a `styhead` branch before the change, right?

kraj: yes it will, if not already done so.

PR #2079, opened 2025-01-24
firmware-ele-imx: Use generic-bsp compatible

tq-steina:
There is no need to limit this to NXP BSPs. This should also be backported to scartgap


angolini:
Can you, please, list what generic/mainline linux kernel version have you tested this?

PR #2080, opened 2025-01-24
Update libimxvpuapi2 to 90d3a52 version to incorporate skipped-frame-info fix

AstarRepo:
Updated libimxvpuapi2_2.3.0.bb to includ fix from:
https://github.com/Freescale/libimxvpuapi/pull/62
into meta-freescale.


angolini (changes requested):
Please elaborate the PR description and use the proper standard to create the commit log

dv1:
Hmm. The fixes are important, but a versioned release would be better. I'll do a version 2.3.1 . It is about time anyway.

dv1:
I pushed an update to version 2.3.1. See here: https://github.com/Freescale/meta-freescale/pull/2114

This replaces this PR:

PR #2085, opened 2025-01-29
[styhead] drop walnascar compatiblity

Villemoes:
This branch is not, in fact, compatible with walnascar due to oe-core commit 4ccc3bc8266c3.


angolini (approved):
Nice

and also nice to have the license bump commit on the branch as well

Villemoes:
> and also nice to have the license bump commit on the branch as well

Well, as I pointed out in #2087 , the license bump is actually useless. Look  at the changes it does, it merely shuffles the lines around because whatever working the bot did its thing on had `find` returning the files in another order than the previous time the bot ran... so I'd actually prefer the bot's commit to be excluded.

Villemoes:
> > and also nice to have the license bump commit on the branch as well
> 

I'm actually also rather puzzled why that commit was even added in the first place. "github workflows" is a new thing to me, but from what I can read from .github/workflows/update-license-file.yml, the logic should trigger on any changes to a .inc or .bb file, or daily at midnight UTC. My commit only changes a .conf file, and we certainly haven't passed UTC midnight since I created that branch this morning (CET time).

angolini:
My guess is a new branch, but it's the first time I'm seeing this, so I was only gladly surprised

PR #2107, opened 2025-02-03
linux-fslc-imx: Forward port patches into 6.6-2.2.x

ernestvh:
Port some patches from 6.6-2.1.x to 6.6-2.2.x and update the recipe documentation to include them.


angolini (approved):
Thanks a lot!

otavio:
Your change looks pretty good. Do you mind to update it to include the latest commit that I merged in the same repository? That would avoid us to send a new update one after another.

ernestvh:
> Your change looks pretty good. Do you mind to update it to include the latest commit that I merged in the same repository? That would avoid us to send a new update one after another.

Done!

PR #2116, opened 2025-02-08
linux-fslc: 6.12.6 -> 6.12.13

Livius90:
- Update linux-fslc to v6.12.13 kernel


otavio:
Please update the comment inside of the recipe which mentions the changes that have been included besides the one included in the stable release.

Livius90:
> Please update the comment inside of the recipe which mentions the changes that have been included besides the one included in the stable release.

Whre can i find this comment in `linux-fslc_6.12.bb`? What new text should be here?

otavio:
Ignore my comment, I mix up that with the downstream fork which we maintain and where we have a complete log of the changes that we include. Sorry about the confusion.

PR #2135, opened 2025-02-18
firmware-nxp-wifi: create symbolic links to firmware/mrvl folder

hiagofranco:
The upstream kernel Wi-Fi driver expects the sdiouart8997_combo_v4 and sdiouartiw416_combo_v0 binaries to be located under firmware/mrvl [1], different from the out-of-tree driver, which places them under firmware/nxp. Create symbolic links to deploy these binaries in the correct location for the upstream driver.

Due to this change, bump the recipe version to 1.1.

[1] https://github.com/torvalds/linux/blob/master/drivers/net/wireless/marvell/mwifiex/sdio.h#L31

### DIFF
--- a/recipes-bsp/firmware-imx/firmware-nxp-wifi_1.1.bb
+++ b/recipes-bsp/firmware-imx/firmware-nxp-wifi_1.1.bb
@@ -44,7 +44,10 @@ do_install() {
 
     oe_runmake install INSTALLDIR=${D}${nonarch_base_libdir}/firmware/nxp
 
-
+    # Upstream SDIO8997 and IW416 driver firmwares are located on mrvl folder
+    install -d ${D}${nonarch_base_libdir}/firmware/mrvl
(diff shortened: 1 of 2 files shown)


hiagofranco:
Hi, this would also be needed for Scarthgap. Thanks!

recipes-bsp/firmware-imx/firmware-nxp-wifi_1.1.bb
```diff
@@ -168,5 +173,3 @@ RDEPENDS:${PN}-nxpiw612-sdio += "${PN}-nxp-common"
 RPROVIDES:${PN}-nxpiw612-sdio = "linux-firmware-nxpiw612-sdio"
 RREPLACES:${PN}-nxpiw612-sdio = "linux-firmware-nxpiw612-sdio"
 RCONFLICTS:${PN}-nxpiw612-sdio = "linux-firmware-nxpiw612-sdio"
-
-COMPATIBLE_MACHINE = "(imx-generic-bsp)"
```
hiagofranco, line 172: I think this should be removed since this recipe should not be restricted to NXP machines, as explained on the commit message.

PR #2146, opened 2025-03-03
u-boot: Add CVE_PRODUCT definition

Jasper-Ben:
Adding a correct CVE_PRODUCT definition helps to keep track of upstream CVEs


recipes-bsp/u-boot/u-boot-imx-common_2024.04.inc
```diff
@@ -7,6 +7,7 @@ SRC_URI = "git://github.com/nxp-imx/uboot-imx.git;protocol=https;branch=${SRCBRA
 SRCBRANCH = "lf_v2024.04"
 LOCALVERSION ?= "-imx_v2024.04_6.6.52-2.2.0"
 SRCREV = "6c4545203d123c246c5d7995f2893959506d28e0"
+CVE_PRODUCT = "nxp:uboot_secondary_program_loader denx:u-boot"
```
Jasper-Ben, line 10: Not sure whether this should be overridden when using the fslc bootloader

linkjumper: `CVE_PRODUCT = "denx:u-boot"` -> for u-boot-fslc
`CVE_PRODUCT = "nxp:uboot_secondary_program_loader denx:u-boot"` -> for u-boot-imx
IMO, just add those variables to both inc files.

Jasper-Ben: > CVE_PRODUCT = "denx:u-boot" -> for u-boot-fslc

fslc doesn't have a CVE vendor identifier, does it? 🤔

Jasper-Ben:
I noticed an issue while testing this: since NXP uses a different versioning from the upstream u-boot recipe, denx:u-boot will not return any CVEs, since we are missing the upstream version of u-boot on which the nxp flavored u-boot is based on. I will propose a change in poky to add the possibility of overriding the version in a CVE_PRODUCT entry: https://github.com/yoctoproject/poky/blob/08c435c1bc2c79137a0c221bec626b93a1e3332a/meta/lib/oe/cve_check.py#L207

PR #2157, opened 2025-03-26
Optee

thochstein:
I built only, no runtime test. The clang patches failed to apply with the update, and I didn't know how to fix them, so I disabled them. Note that clang is currently disabled anyway.


recipes-security/optee-imx/optee-os_4.4.0.imx.bb
```diff
@@ -2,9 +2,12 @@
...
-    file://0001-core-Define-section-attributes-for-clang.patch \
-    file://0002-optee-enable-clang-support.patch \
-"
+# FIXME - Clang patches cannot be applied and need to be updated.
+#         Note that Clang is disabled at the moment anyway. See
+#         optee-fslc.inc.
+#SRC_URI += " \
+#    file://0001-core-Define-section-attributes-for-clang.patch \
+#    file://0002-optee-enable-clang-support.patch \
+#"
```
otavio, line 11: I would prefer if you could remove the files together.

angolini (approved):
I don't have what to comment on this

kraj (commented):
meta-arm does not disable clang see

https://git.yoctoproject.org/meta-arm/tree/meta-arm/recipes-security/optee/optee.inc#n19

it only enables it if meta-clang is part of bblayers and TOOLCHAIN is set which by default is clang but user can set it to clang as well. I think you should reword this and also I see here you are explicitly setting TOOLCHAIN to gcc. It would be good to explain why do we need that while upstream from meta-arm does not.

thochstein:
> meta-arm does not disable clang see

The release that I'm upstreaming here was tested with an older version of meta-arm where clang is disabled.

https://git.yoctoproject.org/meta-arm/commit/meta-arm/recipes-security/optee/optee.inc?h=styhead&id=0d0b8ffac2ae7ad6cb50f4ac2b09349d89267edb

recipes-security/optee-imx/optee-fslc.inc
```diff
@@ -1,26 +1,43 @@
...
-B = "${WORKDIR}/build"
-
-OPTEE_ARCH:arm     = "arm32"
+OPTEE_ARCH = "null"
+OPTEE_ARCH:arm = "arm32"
 OPTEE_ARCH:aarch64 = "arm64"
+OPTEE_CORE = "${@d.getVar('OPTEE_ARCH').upper()}"
+
+# FIXME - breaks with Clang 18.  See https://github.com/OP-TEE/optee_os/issues/6754
+TOOLCHAIN = "gcc"
```
kraj: this still hardcodes it ?

thochstein: Please hit Refresh

kraj (approved):
look ok

PR #2163, opened 2025-03-28
linux-fslc: 6.12.13 -> 6.12.20

Livius90:
- Update linux-fslc to v6.12.20 kernel


Livius90:
@otavio 
Can you trigger a backport to `scarthgap`

PR #2194, opened 2025-04-06
isp-imx: fix build against updated tinyxml2

MaxKrummenacher:
Latest master meta-oe updated tinyxml2 to version 11.0.0. isp-imx packages a prebuilt library dynamically linked against tinyxml2.so.10 which is no longer provided, thus:

| ERROR: isp-imx-4.2.2.24.4-r0 do_package_qa: QA Issue: /usr/lib/libcam_device.so contained in package 
| isp-imx requires libtinyxml2.so.10()(64bit), but no providers found in RDEPENDS:isp-imx? [file-rdeps]

Fix the build by updating the dynamic section to require tinyxml2.so.11.

Note that I don't have a system to runtime test if that still works. The tinyxml2 API doesn't look like it changed from 10 to 11.

Note that the tinyxml2 version bump is only in meta-openembedded master. Styhead is providing tinxml2 10.0.0.


otavio:
@thochstein Is it possible to not provide the binary and instead to provide the source code so this can be properly built?

dolcini:
@otavio : should we take this in the meantime we figure out if a better option is available?

PR #2195, opened 2025-04-07
alsa-state: Use i:MX-specific configuration to support QorIQ platforms

tq-steina:
This asound.conf can be used for both i.MX and Layerscale platforms. But the latter don't have `imx-<default>-bsp` in their overrides. Use the common `use-<default>-bsp` instead.

### DIFF
--- a/recipes-bsp/alsa-state/alsa-state.bbappend
+++ b/recipes-bsp/alsa-state/alsa-state.bbappend
@@ -2,3 +2,4 @@
 FILESEXTRAPATHS:prepend := "${THISDIR}/${PN}:"
 
 PACKAGE_ARCH:imx-generic-bsp = "${MACHINE_ARCH}"
+PACKAGE_ARCH:qoriq-generic-bsp = "${MACHINE_ARCH}"
--- a/conf/machine/include/qoriq-base.inc
+++ b/conf/machine/include/qoriq-base.inc
@@ -43,6 +43,8 @@ EXTRA_IMAGEDEPENDS += "u-boot qoriq-cst-native"
 
 MACHINEOVERRIDES =. "qoriq:"
 
+INHERIT += "machine-overrides-extender"
+
 # Machines or distros can define which BSP it should use by default. We are
 # intending to default for nxp BSP by default and specific machines or
 # DISTROs might change it if need.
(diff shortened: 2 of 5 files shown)


otavio:
Please, instead of making this change, introduce the extra overwrites. So we use this consistently and not use a common place for everything.

tq-steina:
Just to be clear, you mean adding `qoriq-mainline-bsp` and `qoriq-nxp-bsp`. Additionally adding a symlink for alsa-state, right?

otavio:
Exactly, we should not overuse those overrides that are intended to properly select the BSP to be used and not to be used across the recipes.

otavio:
You don't need the extra symbolic links if you are using the generic override as a base because both BSP uses the generic as a common override. So please drop the extra symbolic links, and we are good to merge.

tq-steina:
> You don't need the extra symbolic links if you are using the generic override as a base because both BSP uses the generic as a common override. So please drop the extra symbolic links, and we are good to merge.

Without the symlinks QorIQ platfroms don't have `MACHINEOVERRIDES` which matches `imx-generic-bsp` or `imx-mainline-bsp`. So the default (empty) `asound.conf` is picked up. That's why I initially went for using `use-generic-bsp` & friends.

otavio:
So this is exactly what we need to adjust so those are injected automatically and then you can avoid the extra symbolic link

otavio:
The symbolic link is not using the override that you created; this is the only thing that is missing. You can use the generic one as, up to now, we are not making it to vary depending on the SoC we are using, just about the family.

otavio:
Yes, it has been changed since you sent it, so it now needs to be rebased, so it can be merged again.

recipes-bsp/alsa-state/alsa-state.bbappend
```diff
@@ -2,3 +2,4 @@
 FILESEXTRAPATHS:prepend := "${THISDIR}/${PN}:"
 
 PACKAGE_ARCH:imx-generic-bsp = "${MACHINE_ARCH}"
+PACKAGE_ARCH:qoriq-generic-bsp = "${MACHINE_ARCH}"
```
otavio, line 5: Please use this override for the directory name for the files.

tq-steina: I'm confused, sorry. `imx-generic-bsp` doesn't work on QorIQ-platforms, and vice versa. Using this for the directory name needs either:
* symlink from `qoriq-generic-bsp` -> `imx-generic-bsp`  (for the `nxp` and `mainline` directories as well)
* Create new directory and copy already existing files

I'm not fond of both that's why I went for using `use-<foo>-bsp` instead, which matches both imx and qoriq platforms

otavio: Currently, they are exactly the same. I think that you create a symbolic link. But it is important to make clear that both families are sharing those files. It is critical for long-term maintenance that we make this explicit. Otherwise, it can be confusing.

PR #2204, opened 2025-04-17
kernel-module-nxp-wlan: remove COMPATIBLE_MACHINE

ivitro:
The current COMPATIBLE_MACHINE restricts the recipe to i.MX SoCs with both BSP variants. However, the kernel-module-nxp-wlan provides the MXM drivers for the NXP Wi-Fi module (e.g., MAYA-W260), which may also be used with other SoCs like the Toradex Verdin-AM62P.

To enable deployment of the MXM drivers on non-i.MX platforms, remove the COMPATIBLE_MACHINE restriction.


ivitro:
@otavio, could you please backport to scarthgap?

Issue #2208, opened 2025-04-24
imx-vpu-hantro-daemon out of sync with the kernel


landswellsong:
I'm investigating the errors I get while trying to encode video on IMX8MP with V4L2 layer and I get the encoder params all messed up (width=100, height=0, unsupported format which I added a print for and it's way out of range):
```
# cat /home/vsi/daemon.log
src/vsi_dec.c:176:vsi_dec_get_codec_format() INFO core_num 2, ReadAsicCoreCount 2
src/vsi_main.c:365:main() INFO main thread id = ffffbe525720
src/vsi_main.c:585:inst_init() DEBUG inst_init begin
src/vsi_main.c:626:inst_init() DEBUG inst_init ffffbe0df1a0 end
src/vsi_main.c:515:worker_threads() LOG worker_threads inst[2800000001]: codec_fmt 1, cmd id 1
src/vsi_main.c:515:worker_threads() LOG worker_threads inst[2800000001]: codec_fmt 1, cmd id 1
src/vsi_main.c:515:worker_threads() LOG worker_threads inst[2800000001]: codec_fmt 1, cmd id 1
src/vsi_main.c:515:worker_threads() LOG worker_threads inst[2800000001]: codec_fmt 1, cmd id 1
src/vsi_main.c:515:worker_threads() LOG worker_threads inst[2800000001]: codec_fmt 1, cmd id 1
src/vsi_main.c:515:worker_threads() LOG worker_threads inst[2800000001]: codec_fmt 1, cmd id 1
src/vsi_main.c:515:worker_threads() LOG worker_threads inst[2800000001]: codec_fmt 1, cmd id 1
src/vsi_main.c:515:worker_threads() LOG worker_threads inst[2800000001]: codec_fmt 1, cmd id 1
src/vsi_main.c:515:worker_threads() LOG worker_threads inst[2800000001]: codec_fmt 1, cmd id 0
src/vsi_enc.c:261:enc_cmd_processor() INFO Has Param when streamon.
src/vsi_enc_video_h2.c:1241:getAlignedPicSizebyFormat() ERROR not support this format
src/vsi_enc_video_h2.c:1275:get_config_from_cmd_h2() ERROR Invalid parameters width=100, height=0
src/vsi_enc.c:378:enc_state_processor() ERROR set_parameter() failed, ret=-101
src/vsi_main.c:515:worker_threads() LOG worker_threads inst[2800000001]: codec_fmt 1, cmd id 1
[...]

PR #2211, opened 2025-04-25
bump imx-vpu-hantro to latest of rel_6.6.52_2.2.0

JPSGoncalves:
Hello,

The video encoding wasn’t working on i.MX8MM. The change from https://github.com/nxp-imx/linux-imx/commit/08d3e59a62a7dec30a6401e050b88a260f085580 broke compatibility between the kernel driver and the current version of  userspace VPU libraries on `meta-freescale` . While this was fixed on the userspace libraries in `meta-imx`, we still need to bump `meta-freescale`.

I bumped to the latest version of rel_6.6.52_2.2.0 since we'll need the backport for scarthgap too. The i.MX8MP doesn't seem to be affected (at least by this same issue), but it was more practical to bump `imx-vpu-hantro-vc` as well.

The bump was tested on both i.MX8MM and i.MX8MP using latest Toradex BSP . 

Maybe, this also fix https://github.com/Freescale/meta-freescale/issues/2208.

### DIFF
--- a/recipes-bsp/imx-vpu-hantro-vc/imx-vpu-hantro-vc_1.10.1.bb
+++ b/recipes-bsp/imx-vpu-hantro-vc/imx-vpu-hantro-vc_1.10.1.bb
@@ -2,15 +2,16 @@
 
 DESCRIPTION = "i.MX VC8000E Encoder library"
 LICENSE = "Proprietary"
-LIC_FILES_CHKSUM = "file://COPYING;md5=db4762b09b6bda63da103963e6e081de"
+LIC_FILES_CHKSUM = "file://COPYING;md5=ca53281cc0caa7e320d4945a896fb837"
 
 inherit fsl-eula-unpack
 
-SRC_URI = "${FSL_MIRROR}/${BP}.bin;fsl-eula=true"
+SRC_URI = "${FSL_MIRROR}/${BP}-${IMX_SRCREV_ABBREV}.bin;fsl-eula=true"
+IMX_SRCREV_ABBREV = "c0244a1"
 
-S = "${WORKDIR}/${BPN}-${PV}"
+SRC_URI[sha256sum] = "713ba375f25490727fcc62bab5d5508f74de03204b4c153464b696b652c5c7df"
 
-SRC_URI[sha256sum] = "84fcefa0619def2f009ca6651c5cffcda57fed29cd7060ef68be48c5d0d7814b"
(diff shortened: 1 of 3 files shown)


landswellsong:
I'll try running this against Olimex IMX8MP and report back here.

otavio:
Is it possible for you to prepare the same update to the latest BSP? I won't do the backport for the LTS branch, but we can do that later, once we have the new kernel updates merged as well.

JPSGoncalves:
@landswellsong changing the vpu encoder plugin from `v4l2h264enc` to `vpuenc_h264`, make the pipeline to run successfully for me. 

`gst-launch-1.0 -e videotestsrc num-buffers=100 ! video/x-raw,format=NV12,width=1280,height=720 ! queue ! vpuenc_h264 ! fakesink`

Not sure if  `v4l2h264enc` was working before.

landswellsong:
@JPSGoncalves thank you! I'll try that tomorrow on the real hardware.

I'm a little confused however. As per RM00294:
>The Hantro VPU on the i.MX 8M series includes both decoder (G1/G2) and encoder (H1/VC8000E). The early
VPU wrapper interface (deprecated) is built based on Hantro user library, which communicates with the Hantro
kernel driver directly. To support the V4L2 interface, one V4L2-based kernel driver is implemented and one
daemon application is designed accordingly to link the V4L2 driver and Hantro user library.

It was my understanding that `vpuenc_*` were deprecated in favor of exposing the V4L2 interface. I assumed that the daemon was thus only needed for `v4l2*enc` since `vpuenc_*` would just call the library directly.

I don't really mind any of the plugins as long as it works but just wanted to make sure the daemon pipeline also works. I'll think of a way to test it. 

Many thanks again for getting this fixed!

JPSGoncalves:
> It was my understanding that vpuenc_* were deprecated in favor of exposing the V4L2 interface. I assumed that the daemon was thus only needed for v4l2*enc since vpuenc_* would just call the library directly.

@landswellsong you're right. The fix we needed was just to make `vpuenc_h264` work, as this is what we use for testing on the downstream kernel (and customers still use it), so we needed the VPU wrapper lib upgrade. Also, we're still using v6.6 kernel and I'm not sure if `v4l2h264enc` works on this version (though I think from a previous version of RM00294, the VPU wrapper lib was already deprecated on rel_imx_6.6.52_2.2.0). Anyway, I'll check if I can get `v4l2h264enc` to  work as this seems to be supported on scarthgap too.

PR #2231, opened 2025-05-05
kernel-module-nxp-wlan: add patch to lower PRINTM_MMSG() log level

ivitro:
Add a patch to change the log level used in the mwifiex driver's PRINTM_MMSG macro from KERN_ALERT to KERN_INFO.

KERN_ALERT is meant for critical errors; using it for debug messages clutters logs. KERN_INFO is more appropriate for non-critical informational output.


ivitro:
@otavio, could you please backport to scarthgap?

PR #2249, opened 2025-05-12
Fixing issues now present with GCC 15

MaxKrummenacher:


recipes-kernel/linux/linux-imx-headers_6.6.bb
```diff
@@ -8,6 +8,7 @@ LICENSE = "GPL-2.0-only"
 LIC_FILES_CHKSUM = "file://COPYING;md5=6bc538ed5bd9a7fc9398086aedcd7e46"
 
 SRC_URI = "git://github.com/nxp-imx/linux-imx.git;protocol=https;branch=${SRCBRANCH}"
+SRC_URI:append = " file://0001-video-fbdev-mxc-make-uapi-ipu.h-pxp_dma.h-compatible.patch"
```
otavio, line 11: Please use a single definition using multiple lines so we make it clear that we are adding that to the same definition.

otavio:
@MaxKrummenacher Thank you very much for taking a look at this. This is very, very welcome, and I think those should be taken internally. @thochstein, please forward those internally because this is critical to be ended for the next release.

thochstein:
> @MaxKrummenacher Thank you very much for taking a look at this. This is very, very welcome, and I think those should be taken internally. @thochstein, please forward those internally because this is critical to be ended for the next release.

Thanks @MaxKrummenacher and @otavio, will forward.

recipes-kernel/linux/linux-imx-headers_6.6.bb
MaxKrummenacher: @otavio
Done, thanks.

PR #2252, opened 2025-05-12
imx-secure-enclave: Update to lf-6.12.3-1.0.0

mdrodrigo:


thochstein:
Conceptually it seems

recipes-bsp/imx-secure-enclave/imx-secure-enclave_git.bb
```diff
@@ -1,31 +1,59 @@
...
 SYSTEMD_AUTO_ENABLE = "disable"
 SYSTEMD_SERVICE:${PN} = "nvm_daemon.service"
+SYSTEMD_SERVICE:${PN}:mx95-nxp-bsp = ""
+
+do_install() {
+	oe_runmake -C ${S} DESTDIR=${D} install_tests
+}
+
+do_install:append:mx95-nxp-bsp() {
+    # Remove common content that is to be installed by imx-secure-enclave
```
thochstein, line 48: This doesn't make sense since the recipes are now combined. Combining the recipes suggests that you can just drop the append.

thochstein (commented):
I don't think it makes sense. For 95, the imx-secure-enclave recipe did depend on the imx-secure-enclave-seco recipe. So, two builds were being done, one for PLAT=ele, and one for PLAT=seco.

recipes-bsp/imx-secure-enclave/imx-secure-enclave_git.bb
otavio: We can certainly put both together. However, we will need to have a variable to control whether we need to remove or not those files.

otavio:
Ok, I understand what you are explaining, that you have two different builds, but does it make sense to have it build twice? Could you provide more information about why do we need to have both builds and why this is specific of this SOC type?

thochstein:
I don't understand the details, but what I do understand is that they requested both PLAT flavors to be installed for certain SOCs. Building both PLAT flavors in a single recipe seems a good idea. To verify the results, please check that the packages created with the combined recipe do accurately contain the full union of the packages created with the separate recipes.

recipes-bsp/imx-secure-enclave/imx-secure-enclave_git.bb
thochstein: With two recipes, the source install duplicated headers, a README, and a script, which caused a build break due to installing the same file multiple times to the rootfs. With one combined recipe, this specific problem doesn't exist.

otavio:
Okay, but I think that we need more details to understand why this needs to be done. You know that sometimes they envision some fancy architecture for the recipes that are not required, or it's just a misunderstanding between what needs to be done and the proper way of doing that. So could you inquire that internally to understand why this needs to be done that way? And then we see how we can try to support that properly.

thochstein:
Okay, will do.

thochstein:
I think it's relevant, but unfinished. The commit did an update and a rework, but the rework looks to have problems as I pointed out.

PR #2253, opened 2025-05-12
alsa-lib: Add new GLIB_64BIT_TIME_FLAGS options

mdrodrigo:


recipes-multimedia/alsa/imx-alsa-plugins_git.bb
```diff
@@ -20,8 +20,8 @@ inherit autotools pkgconfig use-imx-headers
 PV = "1.0.26+${SRCPV}"
 
 SRC_URI = "git://github.com/nxp-imx/imx-alsa-plugins.git;protocol=https;branch=${SRCBRANCH}"
-SRCBRANCH = "MM_04.09.02_2410_L6.6.y"
-SRCREV = "b2ba082e70333f187972ee4e85f63f9d2f608331"
+SRCBRANCH = "MM_04.09.03_2412_L6.12.y"
+SRCREV = "eaf0900662d6381bcc433165219280cca1752890"
```
otavio, line 24: Include the list of changes between both revisions.

recipes-multimedia/alsa/alsa-lib_%.bbappend
```diff
@@ -12,4 +12,6 @@ SRC_URI:append:imx-nxp-bsp = "${IMX_PATCH}"
 PACKAGE_ARCH:imx-nxp-bsp = "${MACHINE_SOCARCH}"
 
 GLIBC_64BIT_TIME_FLAGS = ""
+GLIBC_64BIT_TIME_FLAGS:arm:imx-nxp-bsp = " -D_TIME_BITS=64 -D_FILE_OFFSET_BITS=64"
```
otavio, line 15: If you're not mistaken, there is a feature that is used to include or not include this support. If that's the case, then we should make this conditional to this distro feature and not architecture-specific.

recipes-multimedia/alsa/imx-alsa-plugins_git.bb
mdrodrigo: Fixed

PR #2272, opened 2025-05-13
Instead of adding patch file directly to SRC_URI, using SRC_URI:append

sathishscripts:
SRC_URI:append will add the patch file even there is complete override of SRC_URI in other layers.


MaxKrummenacher:
We updated meta-toradex-nxp yesterday. Could you try to pull that and rebuild.

sathishscripts:
@MaxKrummenacher It works. So closing this pull request.

PR #2305, opened 2025-06-13
Remove kernel-fitimage image type

MaXinjian:
kernel-fitimage has been removed in following commit of poky:
 | 1d8c78c8cd6 kernel-fitimage.bbclass: remove it

Not sure if is is ok to remove it directly. Or we need to set it to other values?

### DIFF
--- a/conf/machine/ls1012afrwy.conf
+++ b/conf/machine/ls1012afrwy.conf
@@ -12,8 +12,7 @@ require conf/machine/include/arm/arch-arm64.inc
 
 MACHINE_FEATURES += "optee"
 
-KERNEL_CLASSES  = " kernel-fitimage "
-KERNEL_IMAGETYPES = "fitImage"
+KERNEL_CLASSES += "kernel-fit-extra-artifacts"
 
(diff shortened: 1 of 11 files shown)


tq-niebelm:
Should the usage of the class be replaced by `kernel-fit-image.bbclass` As far as I understand if a machine builds the fitImage the bootflow usually somehow depends on it. Simply remove the class may break booting the kernel.

MaXinjian:
> Should the usage of the class be replaced by `kernel-fit-image.bbclass` As far as I understand if a machine builds the fitImage the bootflow usually somehow depends on it. Simply remove the class may break booting the kernel.

Yes, you are right
Now we need to replace KERNEL_CLASSES  = " kernel-fitimage " with KERNEL_CLASSES += "kernel-fit-extra-artifacts"

And I have tested, if we need to build fitimage kernel, now we need to create a linux-xxx-fitimage.bbclass recipe like ./poky/meta/recipes-kernel/linux/linux-yocto-fitimage.bb do

otavio:
We made a branch before this commit.

PR #2308, opened 2025-06-26
Bump linux recipes

angolini:
I only built locally `linux-fslc-imx `and `linux-fslc` for imx8mm-lpddr4-evk


fabioestevam:
> > Why do you bump 6.1 to 6.1.111 instead of 6.1.141?
> 
> Because I used the latest that is in the branch

It is better to update the 6.1 branch to the latest 6.1.141.

recipes-kernel/linux/linux-fslc-imx_6.12.bb
```diff
@@ -43,6 +43,9 @@ Latest stable Kernel patchlevel is applied and maintained by Community."
 # Additional commits may exist to better acommodate yocto builds.
 #
 # $ git log --oneline  --no-merges v6.12.3.. ^mainline/linux-6.12.y ^NXP/lf-6.12.y
```
MrCry0, line 45: `s/6.12.3/6.12.34/`

PR #2311, opened 2025-07-01
pipewire: Simplify gstreamer package config override

thochstein:
Allow the user to more easily override the removal of the gstreamer PACKAGECONFIG.


thochstein:
Hi @otavio, can you please have a look at this? We need this fix in order to override it from our layer.

thochstein:
Thanks, @otavio. And can you backport this to `walnascar`?

PR #2331, opened 2025-08-10
classes: Add support image_types_uuu (tarball)

alimon:
Often is needed to generate a tarball with proper image and uuu binary, and lst files to flash boards via USB.

Can be enabled at bsp or distro level with:

```
IMAGE_CLASSES += "image_types_uuu"
```


otavio:
Take a look on the class below, I think that it is quite aligned with what you are looking for, however it has been done in a more generic way.

https://github.com/Freescale/meta-freescale/blob/master/classes/image_populate_mfgtool.bbclass

alimon:
@otavio Yes, haven't seen it. Thanks!,

PR #2334, opened 2025-08-12
Upgrade MM package to LF6.12.20-2.0.0

zelan-nxp:
imx-alsa-plugins: Bump to 1.0.26
gstreamer: Upgrade to 1.26.0
imx-codec: Upgrade to 4.10.0
imx-dspc-asrc: Upgrade to 1.0.2
imx-opencl-converter: Upgrade to 0.8.0
imx-parser" Upgrade to 4.10.0:
imx-vpuwrap: Upgrade to LF6.12.20-2.0.0
nxp-afe: Initial recipe
nxp-afe-voiceseeker: Initial recipe


recipes-multimedia/gstreamer/gstreamer1.0-plugins-ugly_1.26.0.bb
```diff
@@ -12,11 +12,12 @@ LICENSE = "LGPL-2.1-or-later & GPL-2.0-or-later"
 LICENSE_FLAGS = "commercial"
 
 SRC_URI = " \
-            https://gstreamer.freedesktop.org/src/gst-plugins-ugly/gst-plugins-ugly-${@get_gst_ver("${PV}")}.tar.xz \
+            https://gstreamer.freedesktop.org/src/gst-plugins-ugly/gst-plugins-ugly-${PV}.tar.xz \
```
otavio, line 15: You have dropped the .imx suffix and that could not be done because we could not possibly conflict with existing recipe, so please keep the suffix when updating.

recipes-multimedia/gstreamer/gstreamer1.0-python_1.26.0.bbappend
```diff
@@ -0,0 +1 @@
+DEPENDS:append = " gstreamer1.0-plugins-bad"
```
otavio, line 1: You need to create the copy of the recipe here with the suffix, so we have a consistent version, even if doing the backport in the future.

otavio: We have been adding the suffix to ensure proper identification of the recipe, which is not sourced from the project's official repositories. That is why we have been using this approach.

recipes-multimedia/gstreamer/gstreamer1.0-plugins-ugly_1.26.0.bb
zelan-nxp: Okay, Updated

thochstein:
> gstreamer upgrade depends on mesa and wayland-protocols upgrade to 1.41.imx. @thochstein

See #2342

Issue #2326, opened 2025-08-05
[SCARTHGAP] Upgrade BSP to LF6.6.52_2.2.1


angolini:
The immediate items to work on are:

- [x] Upgrade Linux kernel and headers (i.MX variants)
    - [x] https://github.com/Freescale/meta-freescale/pull/2329
- [x] Upgrade Linux kernel (FSLC forks i.MX variants)
    - [x]  merge the tag - https://github.com/Freescale/linux-fslc/pull/736
    - [x] https://github.com/Freescale/meta-freescale/pull/2328
- [x] Upgrade Linux kernel Modules
    - [x] #2336 
- [ ] Upgrade U-Boot (FSLC forks i.MX variants)
    - [x] imx-boot-imx #2349 
- [x] Upgrade OP-TEE (i.MX variants)
    - [x] #2348 
- [x] Upgrade ATF (i.MX variants)
    - [x] #2327  
- [x] Upgrade Multimedia Packages
   - [x] #2349 
- [x] Upgrade Graphics Packages
   - [x] #2341  
- [x] Upgrade Jailhouse
    - [x] #2280 
- [ ] Upgrade mcore-demos
- [x] Upgrade firmware-imx*
    - [x] #2327  
- [x] Upgrade imx-mkimage
    - [x] #2327 
- [ ] Upgrade NXP Wi-Fi
- [ ] Upgrade imx-lib
- [ ] Upgrade imx-uuc
- [ ] Upgrade imx-test
- [ ] Upgrade isp-imx
- [x] Upgrade imx-system-manager - https://github.com/Freescale/meta-freescale/pull/2379

Release Notes: **link missing**

ernestvh:
I'm working on imx-mkimage, firmware-imx* (especially firmware-ele-imx), imx-atf and imx-oei

angolini:
I"m working with kernel, linux-imx/linux-fslc-imx

jviguera:
@eghidoli, please review my comments in the merged optee pull request. There are a couple of minor things that need to be fixed.

angolini:
We have very few packages still pending.

* `u-boot-fslc` could be updated, if there is anyone available. (but it's not part of the upstream)
* imx-lib
* imx-uuc
* imx-test
* isp-imx
* mcore-demos

Is there anything else I'm missing?

jviguera:
[...]

PR #2335, opened 2025-08-12
[SCARTHGAP] Update imx-oei to NXP BSP LF6.6.52-2.2.1, support B0 silicon

ernestvh:
This is part of issue #2326

Took the changes from `meta-imx` to update to BSP 6.6.52-2.2.1, but with one notable difference in `imx95-a1-19x19-verdin.conf`:

Add the "imx95-19x19-verdin" machine override such that all
conditional overrides for the base machine are also executed for this A1
machine.
    
In practice this override was only used in `imx-boot_1.0.bb` for
MKIMAGE_EXTRA_ARGS. meta-imx added a `imx95-a1-19x19-verdin` override
there that does the same as the `imx95-19x19-verdin` override, this
commit solves it differently by adding the `imx95-19x19-verdin` override
to the A1 machine.

### DIFF
--- a/conf/machine/imx95-19x19-verdin.conf
+++ b/conf/machine/imx95-19x19-verdin.conf
@@ -9,6 +9,12 @@ MACHINEOVERRIDES =. "mx95:"
 require conf/machine/include/imx95-evk.inc
 IMX_DEFAULT_BSP = "nxp"
 
+# iMX95 Verdin EVK V1.1E is the first revision with a B0 SoC,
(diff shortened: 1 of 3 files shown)


otavio (approved):

otavio:
I agree with Daiane here. Let's choose the default value that we are going to use and let the user change it using the `local.conf` file or change the machine themselves, and not duplicate the machine definition.

PR #2351, opened 2025-09-04
classes: add classes for uuu bundle generation

dwagenk:
> **NOTE** This is meant for discussion. This implementation evolved over the past few years and I've used it in multiple projects. The image_populate_mfgtool class did not exist when I first used this class. Are you interested in integrating this into meta-freescale?

The uuu_bundle.bbclass can be used to create uuu zip files containing arbitrary files collected from DEPLOY_DIR_IMAGE and a uuu.auto script. A user provided uuu.auto.in script is processed to replace file placeholders and text placeholders.

The uuu_image_bundle.bbclass does the same for image recipes when inheriting the class and specifying a basetype with `uuu-zip` extension in IMAGE_FSTYPE, e.g. `ext4.gz.uuu-zip`. This can be done for multiple basetypes, so the same image recipe can provide different uuu-bundles, e.g. for booting the image from ramdisk for test purposes or flashing it to internal storage.

The provided classes differ from the image_populate_mfgtool.bbclass by:
- not being limited to image recipes
- allowing creation of uuu bundles for multiple types from IMAGE_FSTYPES
- not bundling the uuu binaries
- creating a name.uuu.zip file that can be directly used with the uuu tool, e.g. uuu path/to/name.uuu.zip
- more flexible variable replacement, e.g. not assuming `UBOOT_BINARY` is used as bootloader in uuu.auto script
- properly integrating with the Openembedded deploy and image deploy mechanisms
- different semantics

### DIFF
--- a/classes/uuu_bundle.bbclass
+++ b/classes/uuu_bundle.bbclass
@@ -1,0 +1,118 @@
+# Class to create a zip file suitable for flashing to NXP processors via the
+# uuu tool.
+#
+# Usage:
+#   Provide a `uuu.auto.in` flashing script via SRC_URI.
+#   In the recipe set the Variable `UUU_FILES` to a list of identifiers
+#     UUU_FILES = "bootloader whatever"
+#   and matching variable flags containing the file names or paths like
+#     UUU_FILE[bootloader] = "u-boot-${MACHINE}.${UBOOT_SUFFIX}-uuu"
+#     UUU_FILE[whatever] = "some-file.bin"
+#   Reference these files via `@@file_var@@`, e.g.
+#     @@file_bootloader@@
+#     @@file_whatever@@
+#   placeholders in your uuu script.
+#   The corresponding files will be searched in `${WORKDIR}/uuu` and
+#   DEPLOY_DIR_IMAGE (in that order) and added to the uuu bundle.
+#
+#   Specify needed dependencies via the UUU_DEPS variable, e.g.
+#     UUU_DEPS = "virtual/kernel:do_deploy"
+#
+#   In the recipe set the Variable `UUU_TEXTS` to a list of identifiers
+#     UUU_TEXTS = "bootargs whatever"
+#   and matching variable flags containing strings like
+#     UUU_TEXT[bootargs] = "init=/bin/sh"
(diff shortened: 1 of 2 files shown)


dwagenk:
I'm still interested in integrating this class in meta-freescale.

otavio:
@dwagenk I don't see a reason why we couldn't apply this. That said, I would like to understand why you are creating a new set of classes instead of using the ones that already exist. So, could you verify what the limitations that the current version doesn't offer for you?

dwagenk:
@otavio Thanks for the heads up!

Like noted above part of the reason is that I already had an earlier version of this when the `image_populate_mfgtool` class did not exist yet. I know that reason is not very relevant for upstreaming.

Besides that:
- The mfgtool class makes more assumptions, e.g.
   - the set of replacements is hardcoded here https://github.com/Freescale/meta-freescale/blob/master/classes/image_populate_mfgtool.bbclass#L111-L114 which do not really fit when using e.g. barebox as bootloader
   - I usually do not add the uuu binaries to the bundles. This is possible with the introduced classes but not the default
- The mfgtool class writes to `DEPLOY_DIR_IMAGE` directly, which is wrong according to https://docs.yoctoproject.org/singleindex.html#term-DEPLOY_DIR_IMAGE . I think it does not break sstate since the class is only used for images and they usually have a `nostamp` task and are thus rebuild everytime
- The introduced classes allow for non-image bundles, e.g. I have a `reset-u-boot-env-bundle.bb` that does not flash anything but just usb-boots a u-boot and erases the env partition on the device
- The introduced uuu_image_bundle class works well with `IMAGE_FSTYPES`, e.g. I often have `IMAGE_FSTYPES += "cpio.gz.uuu-zip wic.gz.uuu-zip"` with the `cpio` variant just using usb-booting to boot into the image as a live system for testing purposes (except partitioning and persistence), while the `wic` version flashes the emmc.

I'll collect some usage examples and add note them here to illustrate the usage.

dwagenk:
Closing this for now. I observed occasional [pseudo abort](https://wiki.yoctoproject.org/wiki/Pseudo_Abort) problems when using this class in combination with meta-freescale. I wasn't able to reproduce and debug them yet.

PR #2373, opened 2025-09-17
[master] uuu: upgrade 1.5.165 -> 1.5.233

eghidoli:
Upgrade uuu to latest release, which introduces support for container format v2 required on newer SoCs such as i.MX943 and i.MX95 B0.

This version adds support for macOS on ARM. As a result, the former executable uuu_mac has been renamed to uuu_mac_x86, and a new uuu_mac_arm binary is provided.


eghidoli:
This must be backported to walnascar and scarthgap but, considering NXP documentation, they are using older versions:
- UUU v1.5.182 in scarthgap
- UUU v1.5.201 in walnascar

Can we just bump to the latest version also on these branches?

PR #2402, opened 2025-10-30
imx-oei: split recipe

tq-steina:
Similar the recipe into version configuration and build instructions, similar to imx-system-manager.

This makes it easier to select board-specific branches while keeping the build instructions at a common location


otavio (approved):

thochstein (approved):
LGTM

otavio:
Can you prepare, please?

PR #2418, opened 2025-12-17
qtwayland: extend xcomposite disabling to all i.MX 9 family

ivitro:
Broadens the qtwayland xcomposite feature removal from `mx95-nxp-bsp` to `mx9-nxp-bsp` to cover the entire i.MX 9 family.

Without this change, qtwayland builds fail on i.MX 93 SoCs with the following error:

```
  ERROR: Feature 'xcomposite-glx' was enabled, but the pre-condition 'features.wayland-client && features.opengl && !features.opengles2 && features.xlib && libs.xcomposite && libs.glx' failed.
```


ivitro:
@otavio could you please backport to scarthgap?  Apologies, I should have mentioned that in PR description

PR #2443, opened 2026-02-16
imx-boot: add mx943 support

BenjaminMiss:
Add the support for imx943 inspired from meta-imx.
Based on 6.12.49-2.2.0 NXP BSP release


angolini:
#2423

PR #2444, opened 2026-02-17
linux-imx: Add '.git' suffix to SRC_URI

kamildziezyk:
This patch fixes the need to clone the same repo twice. The 'github.com/nxp-imx/linux-imx.git' repo is used in:
 - linux-imx-headers
 - linux-imx
 - kernel-module-imx-gpu-viv


thochstein (approved):
Good catch!

PR #2487, opened 2026-03-30
mkimage: Update to lf-6.18.2-1.0.0

zelan-nxp:
mkimage: Update to lf-6.18.2-1.0.0
- Add crrm boot type
- Add mcore image deploy

### DIFF
--- a/recipes-bsp/imx-mkimage/imx-boot_1.0.bb
+++ b/recipes-bsp/imx-mkimage/imx-boot_1.0.bb
@@ -1,4 +1,4 @@
-# Copyright (C) 2017-2024 NXP
+# Copyright (C) 2017-2024,2026 NXP
 
 require imx-mkimage_git.inc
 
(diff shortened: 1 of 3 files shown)


zelan-nxp:
Can we merge this PR ?

ernestvh:
Hi @zelan-nxp, thanks for all your work on updating to the 6.18 BSP. Could you have a look at the imx-boot recipe and see how we best support the generic BSP usecase and NXP BSP usecase without issues?

zelan-nxp:
> Hi @zelan-nxp, thanks for all your work on updating to the 6.18 BSP. Could you have a look at the imx-boot recipe and see how we best support the generic BSP usecase and NXP BSP usecase without issues?

Hi @ernestvh , which issues did you met? I've built with imx8mp, no issues found.

ernestvh:
> Hi @ernestvh , which issues did you met? I've built with imx8mp, no issues found.

The issues happen when `M4_DEFAULT_IMAGE` is not set. I built for the imx95. I'm assuming you built for the NXP BSP in which this variable is set. Right now with this patch, the do_deploy and do_compile tasks seem to assume that the `M4_DEFAULT_IMAGE` always exists. See also my comments above on line 240 and 429. Thanks.

dolcini:
@zelan-nxp : is this change forcing the deploy of some m4 demo firmware for lot of machine? If yes, I think this should be changed, there are use case in which this is not ok, and this is forcing this to everyone, out of the blue, hidden in a commit with as title just `mkimage: Update to lf-6.18.2-1.0.0`.

Am I missing something? @otavio ?

otavio:
@dolcini You are right; however, this is what NXP has on its side, so I think we should start cleaning this up to improve it for the next version.

@zelan-nxp @thochstein Ideally, we should only install the firmware file we are going to use, which can be done by filtering the features we are enabling or, in the worst case, selecting this SoC family.

zelan-nxp:
> @dolcini You are right; however, this is what NXP has on its side, so I think we should start cleaning this up to improve it for the next version.
> 
> @zelan-nxp @thochstein Ideally, we should only install the firmware file we are going to use, which can be done by filtering the features we are enabling or, in the worst case, selecting this SoC family.

Yes, It's unreasonable to force install m4 image each machine. as so far, they are only supported at i.MX SOCs. Let us drop this change firstly.  See https://github.com/Freescale/meta-freescale/pull/2515

dolcini:
@zelan-nxp @otavio 

> Yes, It's unreasonable to force install m4 image each machine. as so far, they are only supported at i.MX SOCs. Let us drop this change firstly. See #2515

I would also add that the assumption that the firmware are generic for each SoC is not correct. meta-freescale is a generic BSP that enable also non-NXP boards, and the firmware running on the cortex-M on the EVK might just be not compatible.

IF anything (it should stay optional), it must be done per board, not per SoC

otavio:
@dolcini, I want to ask you to review the proposed PR #2515  and centralize the discussion there so we keep it in the proper place to track progress.

PR #2488, opened 2026-03-30
Bsp: upgrade to 6.18.2

zelan-nxp:
firmware-ele: Bump 2.0.2 - > 2.0.5

- Don't deploy the extra/runtime firmware, it's not needed by imx-boot
- Don't create separate packages for the firmware, one package is sufficient

imx-oei: Update to lf-6.18.2-1.0.0

- Drop OEI_CONFIGS and use PACKAGECONFIG directly
- Add the UBOOT_CONFIG sd-ecc so a bootloader with DDR or ECC can be selected

imx-sm: Bump 2025q4 -> 2026q1

firmware-imx: Bump 8.28 -> 8.31
- install VPU firmware


zelan-nxp:
Can we merge this PR？

PR #2492, opened 2026-03-30
graphics: piglit and x11 update to lf-6.18.2

zelan-nxp:
piglit: Remove opencl for 6 and 7
xinput-calibrator: Remove xinput_calibrator.desktop from xdg/autostart to avoid xinput_calibrator runs when Desktop shows
xserver-xf86-config: Add xorg.conf for imxdrm and imxfbdev

### DIFF
--- a/recipes-graphics/xinput-calibrator/xinput-calibrator_%.bbappend
+++ b/recipes-graphics/xinput-calibrator/xinput-calibrator_%.bbappend
@@ -1,0 +1,7 @@
+# FIXME: Here it can avoid xinput_calibrator run under background
+#        when Desktop shows
+do_install:append() {
+     # Remove the xinput_calibrator.desktop from xdg/autostart
+     rm -rf ${D}${sysconfdir}/xdg/autostart
+}
+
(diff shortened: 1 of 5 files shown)


recipes-graphics/xinput-calibrator/xinput-calibrator_%.bbappend
```diff
@@ -0,0 +1,7 @@
+# FIXME: Here it can avoid xinput_calibrator run under background
+#        when Desktop shows
+do_install:append() {
+     # Remove the xinput_calibrator.desktop from xdg/autostart
+     rm -rf ${D}${sysconfdir}/xdg/autostart
+}
```
otavio, line 6: This change impacts everyone, and I want to understand the reason behind it. The auto start feature is crucial because it allows the device to begin automatically when calibration is needed. I believe there are alternative methods to prevent calibration from occurring, such as adding persistent calibration data. In that case, the automatic start could check if calibration exists and avoid triggering it unnecessarily. It would be helpful to explore if any mechanism already exists for this purpose.

otavio (changes requested):

PR #2508, opened 2026-04-22
isp-imx: explicit coreutils dependency

landswellsong:
The startup script does `realpath` / `head` with options that busybox doesn't support. 
image

I assume a lot of builds include `coreutils` elsewhere.

### DIFF
--- a/recipes-bsp/isp-imx/isp-imx_4.2.2.26.1.bb
+++ b/recipes-bsp/isp-imx/isp-imx_4.2.2.26.1.bb
@@ -82,6 +82,6 @@ FILES_SOLIBS_VERSIONED = " \
 
 INSANE_SKIP:${PN} = "already-stripped"
 
-RDEPENDS:${PN} = "libdrm"
+RDEPENDS:${PN} = "libdrm coreutils"
 
 COMPATIBLE_MACHINE = "(mx8mp-nxp-bsp)"


recipes-bsp/isp-imx/isp-imx_4.2.2.26.1.bb
```diff
@@ -82,6 +82,6 @@ FILES_SOLIBS_VERSIONED = " \
 
 INSANE_SKIP:${PN} = "already-stripped"
 
-RDEPENDS:${PN} = "libdrm"
+RDEPENDS:${PN} = "libdrm coreutils"
```
otavio, line 85: I would rather change the script to avoid relying on those features, so we don't introduce a huge dependency.

otavio (changes requested):

recipes-bsp/isp-imx/isp-imx_4.2.2.26.1.bb
landswellsong: This way we'd be patching the OEM code so it can require maintaining should the new versions of the script be different enough for the patch to fail, but I could do that. It would take some experiementation though, so I'll repush the branch when I'm done, ok?

otavio: That's not a big problem because we can ask them to apply this fix for their next release. cc @thochstein

landswellsong: Ok will do

PR #2511, opened 2026-04-28
imx-base.inc: set SECOEXT_FIRMWARE_NAME to empty by default

BenjaminMiss:
SECOEXT_FIRMWARE_NAME is set to 'UNDEFINED' by default. Since commit d7930c6e ("firmware-ele-imx: Bump 2.0.2 -> 2.0.5"), the firmware-ele-imx recipe fails during do_install if this variable is not overridden.

The install step tries to access a non-existent path named 'UNDEFINED', resulting in:
  install: cannot stat '.../UNDEFINED': No such file or directory

Set SECOEXT_FIRMWARE_NAME to an empty value to prevent build failures when the parameter is not explicitly defined.

### DIFF
--- a/conf/machine/include/imx-base.inc
+++ b/conf/machine/include/imx-base.inc
@@ -504,7 +504,7 @@ MACHINE_EXTRA_RRECOMMENDS += " \
 # ELE extra Firmware,to be installed and accessed from userspace.
 # Note that multiple names can be given in order to install multiple versions
 # to a single rootfs.
-SECOEXT_FIRMWARE_NAME ?= "UNDEFINED"
+SECOEXT_FIRMWARE_NAME ?= ""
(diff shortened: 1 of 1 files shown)


ernestvh:
Hi @BenjaminMiss,

Thanks for this, I was just looking into the same because we hit this problem as well. Personally I was inclined to partially revert https://github.com/Freescale/meta-freescale/commit/d7930c6e70564c5f9e6acde0afc4d12cf5e38d32 because I do not understand the benefit. I also feel like we are now installing more than we might need:

```SECOEXT_FIRMWARE_NAME:mx95-generic-bsp   ?= "mx95a0runtime-ahab-container.img mx95b0runtime-ahab-container.img"```

For the this solution would also be good if others agree.

@zelan-nxp: Could you explain the need of having both of these in SECOEXT_FIRMWARE_NAME at the same time? My instinct is that this should be a conditional thing coming from a machine conf.

PR #2515, opened 2026-04-29
imx-boot_1.0.bb: Drop mcore image in imx-boot image

zelan-nxp:
As Mcore image only applied for i.MX socs
fixup coommit:6cdff9f7


otavio:
I really like the idea of reducing the number of situations where we install these binaries. However, have you been able to figure out which cases the binaries will still be installed in? If you verify that all the binaries in these specific situations are installed and have been properly tested, have you documented this? Could you share some information about the tests you have conducted?

dolcini:
I did not test it, but it looks good to me and it solves the concerns raised in https://github.com/Freescale/meta-freescale/pull/2487

@ernestvh : any comment from your side?

ernestvh:
> I did not test it, but it looks good to me and it solves the concerns raised in #2487
> 
> @ernestvh : any comment from your side?

I like it. I guess we can figure out where and how to install the binaries when relevant later. LGTM

Thank you @zelan-nxp for the quick resolution.

dolcini:
@otavio : what's the next step to get this merged? Our CI/CD and test are broken since a while because of this issue and this seems also an improvement going in the right direction

PR #2524, opened 2026-05-21
kernel-module-nxp-wlan: fix device wakeup capability of mwifiex driver

fschnyder:
The device is always advertised as being wakeup capable, even if no interrupt line has been allocated for it.

Add a patch to Initialize the device only as wakeup capable if an interrupt has been registered.


otavio:
@thochstein Please review to apply this patch internally.

thochstein:
Will do.

PR #2527, opened 2026-05-27
[RFC] Distro to fragments

angolini:
Phase 1 of the distro-to-fragments migration (RFC: meta-freescale-distro deprecation)
  
  This series adds BitBake configuration fragments for BSP selection and the X11 display backend,
  registers the conf/fragments/ tree with bitbake-config-build, and updates the default-registry
  JSON to match.

  What changes:

  - conf/fragments/graphics/x11.conf: new fragment for the X11 display backend (wayland, xwayland,
  framebuffer already existed).
  - conf/fragments/bsp/fslc.conf: selects the FSLC mainline BSP (IMX_DEFAULT_BSP = "mainline",
  linux-fslc, u-boot-fslc).
  - conf/fragments/bsp/nxp.conf: selects the NXP downstream BSP (linux-imx, u-boot-imx) and carries
   the provider overrides and GStreamer version pins previously found in fsl-base.inc in
  meta-freescale-distro.
  - default-registry: replaces the old distro-bundled presets (which did not express a BSP choice)
  with fslc-framebuffer, nxp-framebuffer, and fslc-eula-framebuffer; replaces the phantom
  distro/fsl and distro/fslc one-of options with the real upstream fragments yocto/distro/poky and
  core/distro/nodistro.

  No breaking changes. Existing users of DISTRO = "fslc-wayland" (or any other distro from
  meta-freescale-distro) are not affected. A companion patch to meta-freescale-distro wires those
  distros through these fragments so they are validated by existing builds before the distros are
  deprecated.

---

* meta-freescale-distro/conf/distro/include/fsl-base.inc remains untouched and not linked
* provides 2 distro options `poky` and `nodistro`.  (bitbake configuration template requires a DISTRO)
* X11 fragment is there, but I'm not really sure we want that
* https://github.com/Freescale/meta-freescale-distro/pull/138 depends on this PR
* another point to remember is that if we decide to deprecate meta-freescale-distro and only use fragments, sstate-cache/sdk are impacted

### DIFF
--- a/conf/fragments/bsp/fslc.conf
+++ b/conf/fragments/bsp/fslc.conf
@@ -1,0 +1,8 @@
+BB_CONF_FRAGMENT_SUMMARY = "Use FSLC BSP components (linux-fslc, u-boot-fslc)"
+BB_CONF_FRAGMENT_DESCRIPTION = "Sets IMX_DEFAULT_BSP = 'mainline' and selects the FSLC \
+linux-fslc kernel and u-boot-fslc bootloader as defaults for i.MX builds. \
+Note: DISTRO = 'poky' is used as-is; TARGET_VENDOR and DISTROOVERRIDES from \
+the legacy fslc-base.inc are not carried over."
+
+IMX_DEFAULT_BSP = "mainline"
+IMX_DEFAULT_BOOTLOADER = "u-boot-fslc"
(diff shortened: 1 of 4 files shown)


angolini:
tested with   

```
./sources/bitbake/bin/bitbake-setup init \
      -L meta-freescale ./sources/meta-freescale \
      https://raw.githubusercontent.com/angolini/meta-freescale/refs/heads/distro-to-fragments/default-registry/configurations/fsl-community-bsp-master.conf.json
```

3,24,1

PR #2528, opened 2026-05-29
imx-atf: Fix build failure with GCC 16.1.0

fschnyder:
The scmi_imx9 source declares a tmp pointer that is never read. The newer GCC version 16.1.0 treats the unused pointer as -Wunused-but-set-variable. With -Werror, this breaks the build.

Add a patch to remove the unused tmp pointer.


otavio:
@thochstein Please apply this internally.

thochstein:
Forwarded, thanks!

PR #2529, opened 2026-06-03
vulkan-wsi-layer: Fix build issue with GCC 16

fschnyder:
'__STDC_VERSION__' is predefined by the compiler in C++ mode. Redefining it triggers '-Wbuiltin-macro-redefined', which with '-Werror' fails the build with GCC 16.

error: '__STDC_VERSION__' defined [-Werror]

As suggested in the Vulkan WSI Layer issue tracker, add a C++ guard before the __STDC_VERSION__ define.


otavio:
@thochstein Apply this internally.

PR #2532, opened 2026-06-14
Consolidate recipe content from meta-freescale-distro

otavio:
## Consolidate recipe content from meta-freescale-distro

This is the first step of consolidating all recipe content into a single
layer. Every recipe, packagegroup and demo image under
`meta-freescale-distro/recipes-*/` is moved into `meta-freescale`, one
commit per recipe.

Going forward `meta-freescale` becomes the single home for recipe content.
`meta-freescale-distro` shrinks into a thin backward-compatibility layer
that only keeps the distro definitions (`conf/distro/fsl-*`, `fslc-*`) for
the wrynose LTS lifetime and is no longer used for recipe development.

### What moves (25 recipes)
- **recipes-devtools:** half, stb
- **recipes-fsl:** fsl-rc-local, 4 demo images (fsl-image-*), 9 packagegroups
- **recipes-graphics:** devil, gli, glm (bbappend), imx-gpu-sdk (+ libxdg-shell), rapidopencl, rapidopenvx, rapidvulkan, assimp
- **recipes-multimedia:** gst-variable-rtsp-server

### Notes
- The distro `.conf`/`.inc` files intentionally stay in meta-freescale-distro.
- No filename collisions with existing meta-freescale recipes; all
  destination directories already exist.
- The distro configs only `require conf/fragments/...`, all of which already
  live in meta-freescale, so nothing breaks for the layers that consume them.
- Plain move (no git-history replay); each commit notes the origin.

Paired with the removal PR: Freescale/meta-freescale-distro#140

### DIFF
--- a/recipes-devtools/half/half_2.1.0.bb
+++ b/recipes-devtools/half/half_2.1.0.bb
@@ -1,0 +1,19 @@
+SUMMARY = "C++ library for half precision floating point arithmetics"
+DESCRIPTION = "half is a C++ header-only library to provide an IEEE-754 conformant \ 
+half-precision floating point type along with corresponding arithmetic operators, \
+type conversions and common mathematical functions."
+
+LICENSE = "MIT"
+LIC_FILES_CHKSUM = "file://LICENSE.txt;md5=813a6278831975d26c115ed6f9c21831"
+
+SRC_URI = "https://sourceforge.net/projects/half/files/half/${PV}/${BP}.zip"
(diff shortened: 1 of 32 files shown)


angolini (approved):
I couldn't find the LICENSE list we had in meta-freescale-distro. Are we going to bring that as well? Or it will be covered by the github actions after the merge?

otavio:
I think that it is going to be automatically generated.

PR #2533, opened 2026-06-15
Add ethos-u recipes to scarthgap

fschnyder:
Similar to how it was done here for walnascar, add recipes for the Ethos-U firmware 
and driver stack that is used by the iMX93 NPU. 
This matches the Ethos-U support provided in the NXP lf-6.6.52_2.2.0 BSP release.


fschnyder:
@otavio I know that there is meta-freescale-ml for this kind of purpose but it has no scarthgap branch.
I'm also not sure if you are fine adding new recipes to this scarthgap branch

Let me know your opinion so I'm also aware for similar stuff in the future.

otavio:
@fschnyder, It seems that this is the same as we have in the master branch, so no problem doing the backport.

PR #2538, opened 2026-06-23
README, SECURITY.md, merge-files, glm: Fix layer compatibility checks

lucianogdittgen:
Update the layer metadata and recipe organization to address failures found
  while validating meta-freescale against current OpenEmbedded-Core tooling.

  - Add the meta-freescale mailing list to `README.md` so the layer provides
    explicit maintainer information.
  - Add `SECURITY.md` to document where potential vulnerabilities should be
    reported and which branches receive security fixes.
  - Remove `debug-files` from `INSANE_SKIP` in `merge-files` because the
  recipe
    does not install `.debug` directories outside the debug package and no
    longer requires that QA exception.
  - Move the `glm` bbappend to the `openembedded-layer` dynamic layer because
    the matching recipe is provided by meta-openembedded. This preserves the
    existing packaging adjustment without causing a dangling bbappend when
    that layer is absent.


README.md
```diff
@@ -4,7 +4,11 @@ Welcome to `meta-freescale`. This document outlines our commitment to providing
 
 This layer provides support for NXP's platforms for use with OpenEmbedded and/or Yocto Project.
 
-### Dependencies
+## Maintainer
+
+Maintainer: meta-freescale maintainers <meta-freescale@lists.yoctoproject.org>
```
otavio: You can put myself as maintainer, explicitly; I don't want to use the mailing list because this is deprecated and not getting much attention lately.

otavio (changes requested):

PR #2544, opened 2026-07-02
u-boot-imx: handle efitools dependency for i.mx9 platforms

jmarcoscosta:
A missing dependency is causing the error below in u-boot-imx, when the target is an i.MX9* platform:

|   cert-to-efi-sig-list /src/build/tmp/work/(...)/u-boot-imx/2025.04/git/CRT.crt arch/arm/dts/capsule_esl_file
| /bin/sh: 1: cert-to-efi-sig-list: not found
| make[3]: *** [scripts/Makefile.lib:392: arch/arm/dts/capsule_esl_file] Error 127

cert-to-efi-sig-list is provided by efitools-native, and this recipe is provided by meta-efi-secure-boot.

Declare this dependency on two levels: u-boot-imx's recipe and meta-freescale's layer configuration. It is conditional to i.MX9* in both places, so this shouldn't affect other platforms.

### DIFF
--- a/conf/layer.conf
+++ b/conf/layer.conf
@@ -10,6 +10,7 @@ BBFILE_PATTERN_freescale-layer := "^${LAYERDIR}/"
 BBFILE_PRIORITY_freescale-layer = "5"
 LAYERSERIES_COMPAT_freescale-layer = "whinlatter wrynose"
 LAYERDEPENDS_freescale-layer = "core"
+LAYERDEPENDS_freescale-layer:append:mx9-nxp-bsp = " efi-secure-boot"
 
 # Add the Freescale-specific licenses into the metadata
 LICENSE_PATH += "${LAYERDIR}/custom-licenses"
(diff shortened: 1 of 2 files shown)


jmarcoscosta:
FYI, this change is related to what was suggested here: https://github.com/Freescale/meta-freescale/pull/2485#issuecomment-4151325900

Best regards,

otavio:
I think that we need to verify if we could or package this recipe over here, or set this as a dynamic layer restriction. In fact, I prefer to just package it over here for now.

jmarcoscosta:
I took a look in the dependencies, and some of them are not provided by oe-core ([libfile-slurp-perl](https://layers.openembedded.org/layerindex/recipe/67091/), [sbsigntool](https://layers.openembedded.org/layerindex/recipe/87353/)). Same for some one of the classes which the original recipe inherits from ([user-key-store](https://github.com/Wind-River/meta-secure-core/blob/master/meta-signing-key/classes/user-key-store.bbclass)). IMHO packaging it in a reliable way seems too tricky if compared to just adding a dependency whenever this family of boards is used.

That said, how would the dynamic layer restriction work in that case? Usually, the condition to activate a dynamic layer is the presence of another layer, but the condition here is more related to the hardware used.

otavio:
I think that we will probably need to include those recipes in our layer for now.

jmarcoscosta:
In that case, I will close this PR. I hope the work will move forward in #2485 . Thanks!

PR #2567, opened 2026-07-15
mcore: Upgrade to lf-6.18.20-2.0.0

flora2086:
i.MX 7ULP: Bump 2.16.000 -> 26.06.00
i.MX 952: Bump 25.12.00 -> 26.06.00
i.MX 8MM, 8MN, 8MNULite, 8MP, 8MQ, 8ULP, 93, 943, 95: Bump 26.03.00 -> 26.06.00


recipes-fsl/mcore-demos/imx-m7-demos_26.06.00.bb
```diff
@@ -0,0 +1,25 @@
+# Copyright 2026 NXP
+# Released under the MIT license (see COPYING.MIT for the terms)
+
+require imx-mcore-demos.inc
+
+LIC_FILES_CHKSUM:mx8mn-nxp-bsp = "file://COPYING;md5=bc649096ad3928ec06a8713b8d787eac"
```
otavio: The md5 seems replicated across the full source, so verify this so we reduce the duplication.

otavio (changes requested):

recipes-fsl/mcore-demos/imx-m7-demos_26.06.00.bb
flora2086: Good idea, have done the deduplication.

PR #2553, opened 2026-07-13
treewide: Sort DEPENDS/RDEPENDS and drop double blank lines

otavio:
Part **3 of 5** of the stacked oelint-adv remediation series.
Based on #2552 (PR 2).

## What

Per-recipe mechanical cleanups (one commit per recipe / PN):

- Sort DEPENDS / RDEPENDS entries alphabetically and consolidate
- Order DEPENDS and package variables per oelint
- Drop double blank lines
- gstreamer1.0-plugins-bad: restore CVE_PRODUCT from OE-core

45 commits, each scoped to one recipe so any single one reverts in isolation.
The diffs are small and homogeneous (dependency reordering / blank-line
removal); they change ordering only, not the dependency sets themselves.

## Why

Mechanical but per-recipe, so grouped separately from the treewide sweeps
and from the semantic metadata work.

## Testing

Parse-level; dependency ordering does not change build output. Spot-checking
a few gstreamer / vulkan / optee recipes is enough.

## Stack

1. #2551 — treewide: Normalize whitespace and add .editorconfig
2. #2552 — treewide: Set HOMEPAGE and lowercase SECTION
3. **treewide: Sort DEPENDS/RDEPENDS and drop double blank lines** ← this PR
4. oelint: Suppress layer-inapplicable rules
5. treewide: Set mandatory SUMMARY/DESCRIPTION/HOMEPAGE and SECTION

PR #2571, opened 2026-07-16
opencv: fix 4.13.0.imx build regressions

lucianogdittgen:
## Summary

  Fix the OpenCV 4.13.0.imx build for i.MX targets by:

  - Pointing CMake to the Kleidicv source already fetched by the recipe on
    AArch64.
  - Guarding the optional `/usr/bin/cpp` rename during `do_install`.
  - Selecting OpenCV 4.13.0.imx for the mx8 and mx9 NXP BSP overrides.

  ## Problem

  On AArch64, CMake attempted to download Kleidicv during `do_configure`
  despite the source already being fetched by the recipe. Since network
  downloads are disabled during configuration, this caused the task to fail.

  After correcting that, `do_install` failed because it unconditionally tried
  to rename `${D}${bindir}/cpp`, which is not produced by OpenCV 4.13.0 in
  the reported configuration.

  The mx8 and mx9 preferred-version overrides also continued to request
  OpenCV 4.6.0.imx, producing version-selection warnings.

  ## Solution

  Pass `KLEIDICV_SOURCE_PATH` to CMake for AArch64 builds so it uses the
  pre-fetched source tree. Only rename the `cpp` path when it exists, and
  update the BSP preferences to 4.13.0.imx.

### DIFF
--- a/dynamic-layers/openembedded-layer/recipes-support/opencv/opencv_4.13.0.imx.bb
+++ b/dynamic-layers/openembedded-layer/recipes-support/opencv/opencv_4.13.0.imx.bb
@@ -84,18 +84,16 @@ EXTRA_OECMAKE = "-DOPENCV_EXTRA_MODULES_PATH=${S}/contrib/modules \
                  -DOPENCV_GENERATE_PKGCONFIG=ON \
                  -DOPENCV_DOWNLOAD_PATH=${OPENCV_DLDIR} \
                  -DOPENCV_ALLOW_DOWNLOADS=OFF \
-    ${@bb.utils.contains("TARGET_CC_ARCH", "-msse3", "-DENABLE_SSE=1 -DENABLE_SSE2=1 -DENABLE_SSE3=1 -DENABLE_SSSE3=1", "", d)} \
-    ${@bb.utils.contains("TARGET_CC_ARCH", "-msse4.1", "-DENABLE_SSE=1 -DENABLE_SSE2=1 -DENABLE_SSE3=1 -DENABLE_SSSE3=1 -DENABLE_SSE41=1", "", d)} \
-    ${@bb.utils.contains("TARGET_CC_ARCH", "-msse4.2", "-DENABLE_SSE=1 -DENABLE_SSE2=1 -DENABLE_SSE3=1 -DENABLE_SSSE3=1 -DENABLE_SSE41=1 -DENABLE_SSE42=1", "", d)} \
+    ${@bb.utils.contains("TARGET_CC_ARCH", "-msse3", "-DCPU_DISPATCH=SSE,SSE2,SSE3,SSSE3", "", d)} \
+    ${@bb.utils.contains("TARGET_CC_ARCH", "-msse4.1", "-DCPU_DISPATCH=SSE,SSE2,SSE3,SSSE3,SSE41", "", d)} \
+    ${@bb.utils.contains("TARGET_CC_ARCH", "-msse4.2", "-DCPU_DISPATCH=SSE,SSE2,SSE3,SSSE3,SSE41,SSE42", "", d)} \
 "
 
 LDFLAGS:append:mips = " -Wl,--no-as-needed -latomic -Wl,--as-needed"
 LDFLAGS:append:riscv32 = " -Wl,--no-as-needed -latomic -Wl,--as-needed"
 
 EXTRA_OECMAKE:append:x86 = " -DX86=ON"
-# disable sse4.1 and sse4.2 to fix 32bit build failure
-# https://github.com/opencv/opencv/issues/21597
-EXTRA_OECMAKE:remove:x86 = "-DENABLE_SSE41=1 -DENABLE_SSE42=1"
+EXTRA_OECMAKE:append:aarch64 = " -DKLEIDICV_SOURCE_PATH=${S}/3rdparty/kleidicv"
 
 PACKAGECONFIG ??= "python3 eigen jpeg png tiff v4l libv4l samples tbb \
     ${@bb.utils.contains_any('DISTRO_FEATURES', '${GTK3DISTROFEATURES}', 'gtk', '', d)}"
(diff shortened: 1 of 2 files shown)


dynamic-layers/openembedded-layer/recipes-support/opencv/opencv_4.13.0.imx.bb
```diff
@@ -314,8 +315,10 @@ do_install:append() {
     if ${@bb.utils.contains('PACKAGECONFIG', 'tests-imx', 'true', 'false', d)}; then
         cp -r share/opencv4/testdata/cv/face/* ${D}${datadir}/opencv4/testdata/cv/face/
     fi
-    # Rename cpp folder to avoid collision with GCC /usr/bin/cpp.
-    mv ${D}${bindir}/cpp ${D}${bindir}/opencv_cpp
+    if [ -e ${D}${bindir}/cpp ]; then
```
otavio: If this is failing, then possibly we can just drop it.

lucianogdittgen: Agreed. This appears to be obsolete for OpenCV 4.13.0, and the upstream recipe does not perform this rename. I’ll drop the block instead of guarding it.

thochstein: We are not seeing any problem with this logic, so I think we need to determine what is different for your build.

thochstein: The recipe in meta-imx is updated with changes from meta-openembedded. These are missing in meta-freescale.

Can you sync the meta-freescale recipe with the meta-imx recipe?

thochstein: I see that opencv is already supposed to be in sync, but seems something went wrong with that. See #2568.

Note that the changes for oelint do affect the copied recipe contents, see https://github.com/Freescale/meta-freescale/pull/2553/changes#diff-a3fdc95c6344cacf056cb8944fe28ab6d9c6df8163dadc038404e7f90b279b88.

thochstein (changes requested):
See comments regarding faulty update #2568.

otavio:
> See comments regarding faulty update #2568.

My understanding is that this pull request fixes those. Are we missing something?

thochstein:
Here are the diffs I see. Note that there are un-necessary whitespace changes as well. Perhaps also due to oelint?

```
nxa21648@lsv080035:~/upstream/sources$ diff -w meta-imx/meta-imx-bsp/recipes-support/opencv/opencv_4.13.0.imx.bb meta-freescale/dynamic-layers/openembedded-layer/recipes-support/openc
v/opencv_4.13.0.imx.bb
19c19
< DEPENDS = "libtool swig-native bzip2 zlib glib-2.0 libwebp"
---
> DEPENDS = "bzip2 glib-2.0 libtool libwebp swig-native zlib"
49d48
< 
88,90c87,89
<     ${@bb.utils.contains("TARGET_CC_ARCH", "-msse3", "-DCPU_DISPATCH=SSE,SSE2,SSE3,SSSE3", "", d)} \
<     ${@bb.utils.contains("TARGET_CC_ARCH", "-msse4.1", "-DCPU_DISPATCH=SSE,SSE2,SSE3,SSSE3,SSE41", "", d)} \
<     ${@bb.utils.contains("TARGET_CC_ARCH", "-msse4.2", "-DCPU_DISPATCH=SSE,SSE2,SSE3,SSSE3,SSE41,SSE42", "", d)} \
---
>     ${@bb.utils.contains("TARGET_CC_ARCH", "-msse3", "-DENABLE_SSE=1 -DENABLE_SSE2=1 -DENABLE_SSE3=1 -DENABLE_SSSE3=1", "", d)} \
>     ${@bb.utils.contains("TARGET_CC_ARCH", "-msse4.1", "-DENABLE_SSE=1 -DENABLE_SSE2=1 -DENABLE_SSE3=1 -DENABLE_SSSE3=1 -DENABLE_SSE41=1", "", d)} \
>     ${@bb.utils.contains("TARGET_CC_ARCH", "-msse4.2", "-DENABLE_SSE=1 -DENABLE_SSE2=1 -DENABLE_SSE3=1 -DENABLE_SSSE3=1 -DENABLE_SSE41=1 -DENABLE_SSE42=1", "", d)} \
97c96,98
< EXTRA_OECMAKE:append:aarch64 = " -DKLEIDICV_SOURCE_PATH=${S}/3rdparty/kleidicv"
---
> # disable sse4.1 and sse4.2 to fix 32bit build failure
> # https://github.com/opencv/opencv/issues/21597
> EXTRA_OECMAKE:remove:x86 = "-DENABLE_SSE41=1 -DENABLE_SSE42=1"
```

otavio:
@thochstein, this is because of the linter. We are trying to automate as much as we can to keep up with a standard format for the recipes, and even though, for now, there's going to be a massive number of changes, if we all adopt this, the updates should be much easier because everything will be on the same format and style.

thochstein:
Sure, no problem. Will OE-Core be updating for the linter?

To be clear, there are still non-whitespace changes that are missing from meta-freescale. Those need to be fixed, which will presumably restore the need for the rename of the `cpp` folder.

otavio:
I don't think that they will adopt, at least not at this moment. However, given our limited number of contributors for the layer, the more we can facilitate the process, the better.

otavio:
> Sure, no problem. Will OE-Core be updating for the linter?
> 
> To be clear, there are still non-whitespace changes that are missing from meta-freescale. Those need to be fixed, which will presumably restore the need for the rename of the `cpp` folder.

@lucianogdittgen, can you synchronize the recipe according to Tom's suggestion? We will reapply the linter fixes. 

@thochstein, once we update here, it would be nice if you could synchronize it internally so we are aligned next time.

thochstein (approved):

PR #2584, opened 2026-07-22
imx-imx-boot-bootpart.wks.in: Use variable to choose ptable type

junzhuimx:
In NXP i.MX BSP, i.MX 8 & 9 is preferred to use gpt ptable:
- Boot ROM Compatibility: i.MX 8/9 Boot ROM can work with both MBR and GPT
- U-Boot Configuration: NXP U-Boot has configured to recognize GPT (CONFIG_EFI_PARTITION=y)
- Partition Alignment: GPT first usable LBA in GPT is in sector 34 (17KB), and i.MX 8&9 has IMX_BOOT_SEEK offset (32-33KB), so no conflict
- Advantages of GPT
    - Support for >2TB storage (not typically relevant for SD cards)
    - More robust with CRC32 checksums and backup partition table
    - Up to 128 partitions vs 4 primary partitions in MBR
    - Better partition naming and GUID identification

### DIFF
--- a/conf/machine/include/imx-base.inc
+++ b/conf/machine/include/imx-base.inc
@@ -682,6 +682,8 @@ SOC_DEFAULT_WKS_FILE:mx8-generic-bsp ?= "imx-imx-boot-bootpart.wks.in"
 SOC_DEFAULT_WKS_FILE:mx9-generic-bsp ?= "imx-imx-boot-bootpart.wks.in"
 
 WKS_FILE ?= "${SOC_DEFAULT_WKS_FILE}"
+IMX_BOOTLOADER_PTABLE             ?= "msdos"
+IMX_BOOTLOADER_PTABLE:use-nxp-bsp ?= "gpt"
 
 SERIAL_CONSOLES = "115200;ttymxc0"
 SERIAL_CONSOLES:mxs-generic-bsp = "115200;ttyAMA0"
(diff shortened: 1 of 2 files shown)


conf/machine/include/imx-base.inc
```diff
@@ -682,6 +682,8 @@ SOC_DEFAULT_WKS_FILE:mx8-generic-bsp ?= "imx-imx-boot-bootpart.wks.in"
 SOC_DEFAULT_WKS_FILE:mx9-generic-bsp ?= "imx-imx-boot-bootpart.wks.in"
 
 WKS_FILE ?= "${SOC_DEFAULT_WKS_FILE}"
+IMX_BOOTLOADER_PTABLE             ?= "msdos"
+IMX_BOOTLOADER_PTABLE:use-nxp-bsp ?= "gpt"
```
thochstein, line 686: The commit message mentions 8 and 9 but then this sets it for 6 and 7 as well.

Also, I thought we would limit this change to our distro, not automatically switch all user boards.

otavio: Personally, I would prefer to limit it to 8 and 9 only if it breaks 6 and 7. Users can reset it if they want to keep the old behavior in their project, especially on the master branch. On the LTS, then we could follow your suggestion, which is to maintain backward compatibility and allow overrides.

junzhuimx:
@otavio please backport this change to branch **wrynose**

conf/machine/include/imx-base.inc
thochstein: Oh, I realize I was mistaken. Since this variable is used only in `imx-imx-boot-bootpart.wks.in`, this commit does have no effect on 6 and 7.

That could be confusing for the user, right?

junzhuimx: let me set it to mx8-nxp-bsp and mx9-nxp-bsp

PR #2595, opened 2026-07-23
fixup! iproute2: Fix hardcoded include path and do_install copies per…

thochstein:
… oelint

The recent commit accidentally moved header files from /usr/include to /usr/include/include.

### DIFF
--- a/recipes-connectivity/iproute2/iproute2_%.bbappend
+++ b/recipes-connectivity/iproute2/iproute2_%.bbappend
@@ -1,11 +1,11 @@
 do_install:append:imx-generic-bsp () {
     install -d ${D}${includedir}/tc
-    cp -R ${B}/include ${D}${includedir}
+    cp -R ${B}/include/* ${D}${includedir}
     install -m 0644 ${B}/tc/*.h ${D}${includedir}/tc
 }
 
 do_install:append:qoriq-generic-bsp () {
     install -d ${D}${includedir}/tc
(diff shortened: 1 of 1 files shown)


otavio (approved):

thochstein:
We're having another build break on Wrynose.

```
Errors were encountered while processing:
 /tmp/apt-dpkg-install-swecwu/293-iproute2-dev_6.19.0-r0_arm64.deb
W: chown to _apt:root of directory /.../yocto/build/tmp/work/imx8mqevk-poky-linux/test-internal-qt6/1.0/recipe-sysroot-native/var/cache/apt/archives/partial failed - SetupAPTPartialDirectory (1: Operation not permitted)
W: chown to _apt:root of directory /.../yocto/build/tmp/work/imx8mqevk-poky-linux/test-internal-qt6/1.0/apt/lists/auxfiles failed - SetupAPTPartialDirectory (1: Operation not permitted)
W: chown to root:adm of file /.../yocto/build/tmp/work/imx8mqevk-poky-linux/test-internal-qt6/1.0/recipe-sysroot-native/var/log/apt/term.log failed - OpenLog (1: Operation not permitted)
E: Sub-process dpkg returned an error code (1)
```

I'm trying a bbappend to add `--no-preserve=ownership` and hoping that that fixes it.

Perhaps it would have been better to not try backporting all these changes to Wrynose so quickly (or even at all?).

PR #2618, opened 2026-07-31
imx-base.inc: Set IMX_BOOTLOADER_PTABLE to gpt for i.MX 8/9 only

junzhuimx:
Even variable IMX_BOOTLOADER_PTABLE is used in imx-imx-boot-bootpart.wks.in only, add thus SoC specific setting to avoid confusing.


thochstein (commented):
It's fine with me, but it does beg the question whether the variable should be added to the other .wic files.

otavio:
Yes, I think that would be better to have a standard here, so it would be good if this was made the same way in all files.

PR #2619, opened 2026-08-03
vulkan-loader: Refresh LF-11869 patch

ernestvh:
The Vulkan Loader source changed add_data_files() to consume a string list for manifest search paths, so the LF-11869 patch no longer applies to the current oe-core recipe.

Refresh 0001-LF-11869-change-mali-wsi-layer-activating-order.patch against the current loader code while keeping the downstream ordering change intact.


otavio:
This cannot be backported, right?

ernestvh:
> This cannot be backported, right?

Indeed, it only applies to master. Thanks!

PR #2620, opened 2026-08-04
image_populate_mfgtool: use oe.utils.base_set_filespath

wardru:
Commit `4e04ca516c6b` ("utils: Move functions from utils.bbclass to utils.py") from openembedded-core moved `base_set_filespath` into oe.utils.

This change is not part of wrynose yet. No backport is needed for now.


wardru:
@otavio I believe it will break wrynose now with the backport. The changes from OE haven't been backported yet

PR #2626, opened 2026-08-07
oelint cleanup: stop the layer changing machines it does not own

lucianogdittgen:
meta-freescale changes machines it does not own. Some bbappend lines carry no machine override, so if the layer is in BBLAYERS they apply to every machine in the build, i.MX or not.

  oelint.vars.noncoreoverride flagged 76 of these. Most are harmless. The layer's normal pattern is a helper variable that is empty by default and only gets a value under an i.MX override, which does nothing off-target. Four were real leaks.

  I checked every one with bitbake -e on qemuarm64, with meta-freescale in and out of BBLAYERS.

  The four real leaks

  - weston-init added no-idle-timeout to every machine, and its do_install:append ran everywhere — about double oe-core's. Now scoped to i.MX.
  - piglit added virtual/libgbm, glslang-native and vulkan-loader to DEPENDS on every machine and turned Vulkan tests on. Now scoped to i.MX.
  - weston-init (IVI) replaced oe-core's weston.ini with NXP's IVI one on every machine, in any build that also has meta-ivi. Now scoped to i.MX.
  - imx-base put the i.MX SDK flashing tools into every SDK. Moved to TOOLCHAIN_HOST_TASK.

  The SDK one needs a word. The obvious fix is to scope the packagegroup's RDEPENDS by machine, and that would be wrong: nativesdk-packagegroup-sdk-host has one fixed PACKAGE_ARCH, so varying its contents by machine would build different packages under the same architecture, and two SDKs from the same TMPDIR would overwrite each other in the feed.
  TOOLCHAIN_HOST_TASK is read by the SDK recipe, which is already per-machine, so it is safe there.

  Everything else

  - u-boot-qoriq overwrote DEPENDS after require u-boot.inc and lost kern-tools-native. It appends now.
  - qemu-qoriq, spc and fmc: removed class-native overrides that can never run.
  - qemu-qoriq: removed an old SRC_URI that is never used. Most of the files it lists do not exist.
  - Added SECTION to the optee includes, fixed dpdk's HOMEPAGE, and made four ordering fixes.
  - Told oelint about six variable names it wrongly reports as typos.
  - The other 63 flagged lines are the harmless pattern. Each is suppressed with a one-line reason.

  Findings: 578 → 482. noncoreoverride: 76 → 9.

  What to check before merging

  1. QorIQ machines change. They lose no-idle-timeout, the weston-init do_install body, and three of the four SDK tools. That is intended, but QorIQ is a family this layer owns, so please confirm it.
  2. The IVI fix is not tested. meta-ivi is not in this tree, so that file is never parsed here. Please confirm it on a tree that has meta-ivi.

  Testing

  - u-boot-qoriq: built before and after on ls1046ardb and compared buildhistory. Packages come out identical.
  - qemu-qoriq SRC_URI: forces a rebuild, and the unpacked source tree is identical before and after.
  - The scoping fixes and the removals: bitbake -e before and after.
  - The comment-only commits: parse plus an oelint rescan.
  - All 22 commits parse: 1166 recipes, 0 errors.


otavio (approved):

thochstein (approved):
LGTM

PR #2634, opened 2026-08-13
Resolve oelint.var.filesoverride findings across i.MX/QorIQ recipes

lucianogdittgen:
## Summary

  This series clears every `oelint.var.filesoverride` finding in the layer. The
  rule flags plain `FILES:<pkg> = "..."` assignments because a hard `=` on a
  package variable silently discards anything a prior `.bbclass`, `.inc`, or
  distro layer set for that package. Each finding is resolved in its own
  per-recipe commit; no functional change is intended in any of them.

  ## Two remediation patterns

  The findings fall into two cases, each handled with the value-safe fix:

  - **Recipe-specific split packages** → `=` becomes `+=`. These `FILES`
    assignments target packages the recipe itself creates (e.g. `libegl-imx`,
    `libcamera` split packages, the optee `-ta` package, `smw` tests,
    `pyjailhouse`), which nothing else ever sets. `+=` is behaviourally
    identical here and no longer trips the check.

  - **Intentional main-package overrides** → keep the assignment, add a
    documented inline `# nooelint: oelint.var.filesoverride`. Recipes such as
    `imx-boot`, `uuu-bin`, `basler-camera`, `imx-vpu-hantro-vc`, and the optee
    QorIQ/tadevkit recipes deliberately repack the main package (e.g. deploying
    the bootloader under `/boot`). The override is correct, so it is accepted
    with a comment explaining why rather than mechanically rewritten.

  Several recipes (`opencv`, `weston`, `mali-imx`, `optee-os-fslc`) mix both
  patterns and are handled accordingly.

  ## Validation

  Each recipe change was validated with a buildhistory before/after comparison
  on a machine that activates the recipe; buildhistory was unchanged in every
  case (machines used include imx8mp-lpddr4-evk, imx95-19x19-verdin,
  ls1028ardb, and imx6qdlsabresd). The per-commit `Tested with ...` trailers
  record the exact target and machine.

  ## Notes

  - One logical change per commit; no unrelated recipes or layers are mixed.
  - Recipes touched: opencv, imx-boot, imx-vpu-hantro-vc, basler-camera,
    uuu-bin, jailhouse-imx, imx-mcore-demos, imx-gpu-viv, mali-imx, weston,
    libcamera, optee-os-fslc, optee-os-tadevkit-fslc-imx, optee-os-qoriq,
    optee-os-qoriq-tadevkit, smw.

### DIFF
--- a/dynamic-layers/openembedded-layer/recipes-support/opencv/opencv_4.13.0.imx.bb
+++ b/dynamic-layers/openembedded-layer/recipes-support/opencv/opencv_4.13.0.imx.bb
@@ -183,12 +183,24 @@ python populate_packages:prepend () {
 
 PACKAGES_DYNAMIC += "^libopencv-.*"
 
+# Main package is intentionally empty (see ALLOW_EMPTY below); everything is
+# split into the sub-packages and the PACKAGES_DYNAMIC libopencv-* packages.
+# nooelint: oelint.var.filesoverride
 FILES:${PN} = ""
 FILES:${PN}-dbg += "${datadir}/OpenCV/java/.debug/* ${datadir}/OpenCV/samples/bin/.debug/*"
+# -dev is curated to headers/pkgconfig/cmake; the per-library dev symlinks are
+# packaged by the PACKAGES_DYNAMIC libopencv-* packages, not the main -dev.
+# nooelint: oelint.var.filesoverride
 FILES:${PN}-dev = "${includedir} ${libdir}/pkgconfig  ${libdir}/cmake/opencv4/*.cmake"
 FILES:${PN}-staticdev += "${libdir}/opencv4/3rdparty/*.a"
+# -apps/-java/-samples are recipe-specific split packages (no default FILES), so
+# the '=' is a complete definition. Kept byte-identical to meta-oe
+# opencv_4.13.0.bb to avoid fork drift; suppress the append-preference warning.
+# nooelint: oelint.var.filesoverride
 FILES:${PN}-apps = "${bindir}/* ${datadir}/opencv4 ${datadir}/licenses"
+# nooelint: oelint.var.filesoverride
 FILES:${PN}-java = "${datadir}/OpenCV/java"
+# nooelint: oelint.var.filesoverride
 FILES:${PN}-samples = "${datadir}/opencv4/samples/"
 
 INSANE_SKIP:${PN}-java = "libdir"
@@ -197,6 +209,9 @@ INSANE_SKIP:${PN}-dbg = "libdir"
 ALLOW_EMPTY:${PN} = "1"
 
 SUMMARY:python3-opencv = "Python bindings to opencv"
+# Recipe-specific split package (no default FILES); '=' is a complete definition.
+# Kept byte-identical to meta-oe opencv_4.13.0.bb to avoid fork drift.
+# nooelint: oelint.var.filesoverride
 FILES:python3-opencv = "${PYTHON_SITEPACKAGES_DIR}/*"
 RDEPENDS:python3-opencv = "python3-core python3-numpy"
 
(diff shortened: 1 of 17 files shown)


otavio (approved):

dynamic-layers/openembedded-layer/recipes-support/opencv/opencv_4.13.0.imx.bb
```diff
@@ -183,21 +183,27 @@ python populate_packages:prepend () {
...
 FILES:${PN}-dbg += "${datadir}/OpenCV/java/.debug/* ${datadir}/OpenCV/samples/bin/.debug/*"
+# -dev is curated to headers/pkgconfig/cmake; the per-library dev symlinks are
+# packaged by the PACKAGES_DYNAMIC libopencv-* packages, not the main -dev.
+# nooelint: oelint.var.filesoverride
 FILES:${PN}-dev = "${includedir} ${libdir}/pkgconfig  ${libdir}/cmake/opencv4/*.cmake"
 FILES:${PN}-staticdev += "${libdir}/opencv4/3rdparty/*.a"
-FILES:${PN}-apps = "${bindir}/* ${datadir}/opencv4 ${datadir}/licenses"
-FILES:${PN}-java = "${datadir}/OpenCV/java"
-FILES:${PN}-samples = "${datadir}/opencv4/samples/"
+FILES:${PN}-apps += "${bindir}/* ${datadir}/opencv4 ${datadir}/licenses"
```
thochstein: This section is copied from upstream. Perhaps better not to change it. Unless you change the upstream as well?

recipes-extended/jailhouse/jailhouse-imx_git.bb
```diff
@@ -91,7 +91,7 @@ PACKAGE_BEFORE_PN += "pyjailhouse"
 FILES:${PN} += "${nonarch_base_libdir}/firmware ${libexecdir} ${sbindir} ${JH_DATADIR}"
 # Remove libdir/* appended by setuptools3-base.bbclass for module split to work correctly
 FILES:${PN}:remove = "${libdir}/*"
-FILES:pyjailhouse = "${PYTHON_SITEPACKAGES_DIR}"
+FILES:pyjailhouse += "${PYTHON_SITEPACKAGES_DIR}"
```
thochstein, line 94: Does this not seem strange? I thought base recipes generally used `=`, not `+=`.

recipes-graphics/wayland/weston_14.0.2.imx.bb
```diff
@@ -132,14 +132,17 @@ PACKAGES += "${@bb.utils.contains('PACKAGECONFIG', 'xwayland', '${PN}-xwayland',
              libweston-${WESTON_MAJOR_VERSION} ${PN}-examples"
 
 FILES:${PN}-dev += "${libdir}/${BPN}/libexec_weston.so"
+# Main package is curated to weston's own binaries, config and plugins; the
+# shared libraries are split out into the libweston-${WESTON_MAJOR_VERSION} package.
+# nooelint: oelint.var.filesoverride
```
thochstein, line 137: Same with this file, these contents are copied. Maybe not hard to deal with, now that I think about it.

dynamic-layers/openembedded-layer/recipes-support/opencv/opencv_4.13.0.imx.bb
otavio: @lucianogdittgen Tom has a point here, and in this case we could change it by manually adding an override to skip the link on this specific line.

recipes-extended/jailhouse/jailhouse-imx_git.bb
otavio: It depends. Normally, we just use `=` to reset it. So normally, we do that in the base package or the `-dev` package. 

In this specific case, it's not a base package; it is a split package. Adding those files won't restrict the files we pack here, since they were already absorbed by the previous packages.

recipes-graphics/wayland/weston_14.0.2.imx.bb
otavio: Yeah, it is not that difficult to handle that because, in one way or another, we will need to adapt to that. At the same time, I feel that the benefit of having an explanation for why we have the override is more meaningful than not needing it at all.

dynamic-layers/openembedded-layer/recipes-support/opencv/opencv_4.13.0.imx.bb
lucianogdittgen: Done in the latest push. Reverted to the upstream = and kept byte-identical to meta-oe, just added # nooelint: per line. That's the per-line override you suggested, @otavio. Same for weston/libcamera.

thochstein (approved):

PR #2642, opened 2026-08-24
oelint: Clear oelint.vars.multilineident across the layer

lucianogdittgen:
Fix every `oelint.vars.multilineident` finding in meta-freescale (23 → 0),
  worked one recipe at a time so each commit stays small and independently
  revertible. No functional change to any recipe.

  ## What the rule flags

  `oelint.vars.multilineident` wants continuation lines of a multi-line
  assignment aligned to the value column. On this layer the findings fall
  into three cases, handled differently:

  - **Genuine misalignment** — the continuation is simply short of the
    value column. Re-indented; the values are whitespace-split, so the
    content is unchanged.
  - **`:append` / `:prepend` leading space** — the space after the opening
    quote is load-bearing (BitBake joins these with no separator). The
    linter reads it as a removable indent; re-indenting would corrupt the
    value. Kept and suppressed with a JUSTIFIED `# nooelint:` directive.
  - **Verbatim upstream copy** — re-indenting would make the file diverge
    from the source it must track. Kept byte-identical to upstream and
    suppressed with an UPSTREAM-PARITY directive; provenance verified by
    diff, not assumed.

  ## Commits (one per recipe)

  - opencv: Restore verbatim-copy indent to upstream, keep parity
  - libdrm: Align LIC_FILES_CHKSUM continuation, keep DESCRIPTION prose
  - imx-gpu-viv: Align FILES:libopencl-imx continuation indent
  - libsdl2: Align CFLAGS continuation, keep :append leading space
  - qtbase: Suppress multilineident on the ' gl' PACKAGECONFIG token
  - qt4-imx-support: Keep :append leading space in QT_CONFIG_FLAGS
  - testfloat: Keep SRC_URI:append leading space
  - wireplumber: Keep SRC_URI:append leading space

  ## Notable

  - **opencv** is a verbatim copy of the meta-openembedded recipe
    (upstream hash `9b77eae6`). `EXTRA_OECMAKE` had drifted to a 17-space
    indent and `PACKAGES` to 13-space; both are restored to the upstream
    4-space so the copied section is byte-identical to the source again,
    then the multi-line assignments are marked UPSTREAM-PARITY.
  - **libdrm** `DESCRIPTION` is prose kept flush-left as in OE-core:
    BitBake concatenates the continuation lines into the string, so
    indenting them would inject spaces into the description.

  ## Validation

  Every commit that changes effective metadata was validated on
  `imx8mm-lpddr4-evk` with a before/after buildhistory comparison:
  **no change to any binary package, file list or dependency.** The only
  delta anywhere in the series is opencv's `-src` (debug source) PKGSIZE,
  which shifts by 6 bytes — the whitespace of the upstream-parity
  re-indent recorded in `/usr/src/debug`; no runtime package is affected.

  Suppression-only commits (qtbase, qt4-imx-support, testfloat,
  wireplumber) touch comments only and were confirmed to leave the
  effective metadata unchanged.

### DIFF
--- a/dynamic-layers/multimedia-layer/recipes-multimedia/wireplumber/wireplumber_%.bbappend
+++ b/dynamic-layers/multimedia-layer/recipes-multimedia/wireplumber/wireplumber_%.bbappend
@@ -3,10 +3,11 @@
 # nooelint: oelint.vars.noncoreoverride
 FILESEXTRAPATHS:prepend := "${THISDIR}/files:"
 
-SRC_URI:append:imx-nxp-bsp = " file://51-bluez-imx.conf \
-                              file://80-disable-logind.conf \
-                              file://0001-wpctl-fix-set-default-Segmentation-fault-on-32bit-pl.patch \
-                              "
+SRC_URI:append:imx-nxp-bsp = " \
+    file://51-bluez-imx.conf \
+    file://80-disable-logind.conf \
+    file://0001-wpctl-fix-set-default-Segmentation-fault-on-32bit-pl.patch \
+"
 
 do_install:append:imx-nxp-bsp () {
 
--- a/dynamic-layers/openembedded-layer/recipes-support/opencv/opencv_4.13.0.imx.bb
+++ b/dynamic-layers/openembedded-layer/recipes-support/opencv/opencv_4.13.0.imx.bb
@@ -79,14 +79,15 @@ addtask unpack_extra after do_unpack before do_patch
 
 CMAKE_VERBOSE = "VERBOSE=1"
 
-EXTRA_OECMAKE = "-DOPENCV_EXTRA_MODULES_PATH=${S}/contrib/modules \
-                 -DWITH_1394=OFF \
-                 -DENABLE_PRECOMPILED_HEADERS=OFF \
-                 -DCMAKE_SKIP_RPATH=ON \
-                 -DWITH_IPP=OFF \
-                 -DOPENCV_GENERATE_PKGCONFIG=ON \
-                 -DOPENCV_DOWNLOAD_PATH=${OPENCV_DLDIR} \
-                 -DOPENCV_ALLOW_DOWNLOADS=OFF \
+EXTRA_OECMAKE = " \
+    -DOPENCV_EXTRA_MODULES_PATH=${S}/contrib/modules \
+    -DWITH_1394=OFF \
+    -DENABLE_PRECOMPILED_HEADERS=OFF \
+    -DCMAKE_SKIP_RPATH=ON \
+    -DWITH_IPP=OFF \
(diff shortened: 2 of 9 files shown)


otavio (approved):

dynamic-layers/multimedia-layer/recipes-multimedia/wireplumber/wireplumber_%.bbappend
thochstein: FWIW, I find this alignment with the quotation mark to be archaic, hard to read and inefficient. Can you fix the problem thus?

```suggestion
SRC_URI:append:imx-nxp-bsp = " \
    file://51-bluez-imx.conf \
    file://80-disable-logind.conf \
    file://0001-wpctl-fix-set-default-Segmentation-fault-on-32bit-pl.patch"
```

dynamic-layers/openembedded-layer/recipes-support/opencv/opencv_4.13.0.imx.bb
```diff
@@ -79,14 +79,17 @@ addtask unpack_extra after do_unpack before do_patch
 
 CMAKE_VERBOSE = "VERBOSE=1"
 
+# Continuation indent kept as in the verbatim meta-openembedded copy (see header).
+# UPSTREAM-PARITY.
+# nooelint: oelint.vars.multilineident
 EXTRA_OECMAKE = "-DOPENCV_EXTRA_MODULES_PATH=${S}/contrib/modules \
```
thochstein: Also, FWIW, I find this easier to read if the first option is treated like the other options and not special cased for the first line.
```suggestion
EXTRA_OECMAKE = " \
    -DOPENCV_EXTRA_MODULES_PATH=${S}/contrib/modules \
```

otavio: @lucianogdittgen I agree with Tom here. I think it also avoids the linting override.

dynamic-layers/openembedded-layer/recipes-support/opencv/opencv_4.13.0.imx.bb
```diff
@@ -98,8 +99,14 @@ LDFLAGS:append:riscv32 = " -Wl,--no-as-needed -latomic -Wl,--as-needed"
 EXTRA_OECMAKE:append:x86 = " -DX86=ON"
 EXTRA_OECMAKE:append:aarch64 = " -DKLEIDICV_SOURCE_PATH=${S}/3rdparty/kleidicv"
 
+# Continuation indent kept as in the verbatim meta-openembedded copy (see header).
+# UPSTREAM-PARITY.
+# nooelint: oelint.vars.multilineident
```
otavio: Please do the same here, as well as the others that you need to have the override.

PR #2649, opened 2026-08-25
wrynose update

mkrahl-emlix:
This is a workaround to support our legacy U-Boot. It must be reverted once we update to a supported version.


otavio:
I'm not sure whether this is specific to your use case or a regression?

mkrahl-emlix:
> I'm not sure whether this is specific to your use case or a regression?

Sorry I selected the wrong target repository. I will close this pr. No effort for you.

PR #2650, opened 2026-08-25
oelint: Resolve vars.specific on dynamic split packages via PACKAGES_DYNAMIC

lucianogdittgen:
## What

Clears all `oelint.vars.specific` findings in meta-freescale: **36 → 0**.
The overwhelming majority are fixed **structurally** by declaring the
dynamically-created split packages; only the packages that are genuinely
resolved at build time by an inherited class are left as documented
suppressions.

## Why

`oelint.vars.specific` flags a variable set on a package override (e.g.
`FILES:${PN}-xwayland`, `RDEPENDS:${PN}-bash-completion`) when the standalone
linter cannot find that package in `PACKAGES`. In every case here the package
*is* real — but it enters `PACKAGES` through a mechanism the static parser
cannot resolve without running BitBake.

Where the recipe can honestly tell the linter about the package, it now does
so with a real `PACKAGES_DYNAMIC` declaration instead of a suppression:

| Approach | Recipes | Mechanism made visible |
| --- | --- | --- |
| **`PACKAGES_DYNAMIC` (structural fix)** | gstreamer1.0-plugins-{bad,good,ugly,base}, opencv, weston, imx-gpu-viv, restool, jailhouse-imx | dynamic `${PN}-*` splits, computed `bb.utils.contains(...)` PACKAGECONFIG packages, override-conditional intermediates, and class-created split packages (`^${PN}-bash-completion$`, `^kernel-module-.*`) |
| **Documented suppression** | linux-qoriq | `${KERNEL_PACKAGE_NAME}-image` / `-base` come from `kernel.bbclass`; the token is fully templated on a class-defined variable the linter cannot expand, and `${...}` are regex metacharacters, so no honest `PACKAGES_DYNAMIC` pattern can match. An inline `# nooelint: oelint.vars.specific` with the reason is the correct exception. |

Making `PACKAGES` statically literal to satisfy the parser would break the
machine-conditional / PACKAGECONFIG-conditional packaging, so that is never
done.

## Verification

Each change was checked against the authoritative whole-layer driver,
`contrib/oelint/run-oelint.sh --mode all`:

- `oelint.vars.specific`: **0** remaining.
- `oelint.file.inlinesuppress_na`: **0** — the remaining suppression is not
  reported as unneeded, so this does **not** trade one rule for another.

The `PACKAGES_DYNAMIC` changes are metadata changes, so each was
**buildhistory-validated output-neutral** on a compatible machine, with the
affected `do_package` force-rebuilt so sstate could not mask a difference:

- restool on `ls1046ardb` (qoriq-arm64): zero buildhistory delta;
  `restool-bash-completion` RDEPENDS/FILELIST/PACKAGES byte-identical.
- jailhouse-imx on `imx8mp-lpddr4-evk` (mx8m-nxp-bsp): zero buildhistory
  delta; the `kernel-module-jailhouse-*` package and all `jailhouse-imx`
  sub-packages byte-identical.
- The gstreamer / opencv / weston / imx-gpu-viv declarations were likewise
  verified inert (each commit body records the machine and result).

`PACKAGES_DYNAMIC` only whitelists package-name patterns the inherited
classes already create; it produces no new package and no downstream
packaging effect, which the buildhistory diffs confirm.

## Fork note

The opencv, weston and gstreamer blocks are verbatim copies of
OE-core / meta-openembedded recipes. The `PACKAGES_DYNAMIC` restatements are
an **intentional, lint-only divergence** from those upstream copies; each
commit body records it so a future re-sync preserves the lines.


otavio:
I'm not sure I like this. I think this is so confusing that I would rather ignore the rule completely than escape it in so many recipes for no good benefit and no good way. What do you think?

otavio (approved):

recipes-extended/jailhouse/jailhouse-imx_git.bb
```diff
@@ -120,6 +120,10 @@ INSANE_SKIP:${PN}-dbg = "buildpaths"
 # The QA error in package kernel-module-${KERNEL_VERSION} cannot be skipped with
 # INSANE_SKIP, so adjust at the ERROR_QA level
 ERROR_QA:remove = "buildpaths"
+# kernel-module-${KERNEL_VERSION} is generated by the kernel-module-split class
+# (inherit module, above); declare it so its INSANE_SKIP is recognised by
+# standalone lint. ${KERNEL_VERSION} is build-time only, so match the prefix.
+PACKAGES_DYNAMIC += "^kernel-module-.*"
```
zelan-nxp: Hi @lucianogdittgen 
The jailhouse-imx change uses too broad a pattern:

PACKAGES_DYNAMIC += "^kernel-module-.*"

^kernel-module-.* matches every kernel module name, so jailhouse-imx becomes a candidate provider for all kernel-module-* virtuals layer-wide. Recipes like iptables RRECOMMENDS those (e.g. kernel-module-x-tables); via do_package_qa[rdeptask] = "do_packagedata" they resolve to a provider, and since BitBake sorts providers alphabetically, jailhouse-imx wins over linux-imx:

$ oe-depends-dot -k jailhouse-imx -w task-depends.dot
core-image-minimal -> iproute2 -> iptables -> jailhouse-imx

So jailhouse-imx gets dragged into core-image-minimal's task graph even when jailhouse isn't installed. The buildhistory check doesn't catch this — the regression is in other recipes' task graphs, not jailhouse-imx's own packaging.

Suggested fix — scope it to this recipe's own module: https://github.com/Freescale/meta-freescale/pull/2674 

PACKAGES_DYNAMIC += "^kernel-module-jailhouse.*"

This still clears the vars.specific finding; afterwards kernel-module-x-tables etc. fall back to linux-imx and oe-depends-dot -k jailhouse-imx returns no path (verified on imx8mp-lpddr4-evk, mx8m-nxp-bsp, core-image-minimal).

CC： @thochstein @otavio

PR #2660, opened 2026-09-02
Set FILESEXTRAPATHS only where it is needed

lucianogdittgen:
oelint-adv reports 10 `oelint.vars.outofcontext` findings on this layer.
Every one is the same thing: `FILESEXTRAPATHS` set in a `.bb` or a `.inc`
instead of a `.bbappend`. This series takes them one at a time and brings
the rule to zero.

`FILESEXTRAPATHS` exists so a `.bbappend` can add its own files directory
to a recipe it does not own. Inside a recipe the default `FILESPATH`
normally already covers the recipe's own directory, so an entry there is
usually either a duplicate of what BitBake supplies or a leftover.

That is what six of the seven turned out to be. The method for each was
the same: run `bitbake -e` on both sides and compare the de-duplicated
`FILESPATH`. In every case the search reached **the same 60 unique
directories before and after** — the entry only added copies of paths the
default `${FILE_DIRNAME}/${BPN}` already supplies, so `FILESPATH` shrinks
from 84 entries to 63 while reaching exactly the same set.

Three entries were fully dead: the resolved `SRC_URI` of those recipes
carries no `file://` entry at all, so nothing was ever looked up through
the search path.

| Recipe / include | Why the entry could go |
| --- | --- |
| `imx-gpu-viv-6.inc` | Added in 2023 for `imx_icd.json`; commit 610fda22 deleted that file and its directory. No `file://` remains. |
| `optee-os-common-fslc-imx.inc` | `optee-os` and `optee-os-tadevkit` both resolve an empty set of `file://` entries. |
| `optee-os.nxp.inc` | Same, for `optee-os-qoriq` and `optee-os-qoriq-tadevkit`. |
| `optee-client-fslc-imx.inc` | Fetches `tee-supplicant.sh`, but `BPN` is `optee-client` in that same directory, so the default already reaches it. |
| `optee-test-fslc-imx.inc` | Fetches `run-ptest`; same reasoning. |
| `gstreamer1.0-rtsp-server_1.28.1.bb` | Fetches a patch; `PN` equals `BPN` here, so the entry duplicated the default. The default is also safer, since `PN` gains a `lib32-` prefix under multilib while `BPN` does not. |

The seventh is genuinely load-bearing and stays. `optee-client-qoriq`
shares `tee-supplicant.sh` with the i.MX client instead of carrying a
second copy, and its `BPN` is `optee-client-qoriq`, so the default
`FILESPATH` never reaches `recipes-security/optee-imx/optee-client`.
Removing the line makes BitBake halt while parsing:

    ERROR: optee-client-qoriq_4.8.0.bb: Unable to get checksum for
    optee-client-qoriq SRC_URI entry tee-supplicant.sh: file could not be found
    ERROR: Parsing halted due to errors

That commit adds the reason next to the line and a
`# nooelint: oelint.vars.outofcontext` directive, and changes nothing else.

Three of the touched files (`optee-*-fslc-imx.inc`) are verbatim copies
of meta-imx includes, as their headers record. Each commit body notes the
intentional divergence so a later re-sync knows the line is now local.

### Validation

Every commit that changes effective metadata was validated with
buildhistory: a baseline build and a candidate build with `do_unpack`
forced, on `imx8mp-lpddr4-evk` or `ls1046ardb` depending on the recipe.
Each recipe ran `do_unpack` through `do_package` on both sides, and every
buildhistory diff was empty apart from the `metadata-revs` "-- modified"
marker. The one comment-only commit needs no build.

The full survey goes from 251 findings to 234: 10 `outofcontext`, 6
`vars.fileextrapaths` cleared by the same edits, and 1 `vars.homepageping`
that is network noise on an untouched file. **Nothing was introduced.**

### DIFF
--- a/recipes-security/optee-qoriq/optee-client.nxp.inc
+++ b/recipes-security/optee-qoriq/optee-client.nxp.inc
@@ -4,6 +4,12 @@ require recipes-security/optee-imx/optee-client-fslc.inc
 
 # The patch same as imx-optee, so point FILESEXTRAPATHS to optee-imx/optee-client,
 # avoid duplicate copy files to optee-qoriq
+# The entry cannot be dropped: this recipe's BPN is optee-client-qoriq, so the
+# default FILESPATH only searches directories named after it, and the shared
+# i.MX directory that holds tee-supplicant.sh is unreachable without this line.
+# The entry also keeps the files this include adds to SRC_URI reachable when a
+# recipe in another layer requires the include.
+# nooelint: oelint.vars.outofcontext
 FILESEXTRAPATHS:prepend := "${THISDIR}/../optee-imx/optee-client:"
 
 DEPENDS = "util-linux-libuuid"
--- a/recipes-graphics/imx-gpu-viv/imx-gpu-viv-6.inc
+++ b/recipes-graphics/imx-gpu-viv/imx-gpu-viv-6.inc
@@ -66,7 +66,6 @@ PE = "1"
 
 inherit fsl-eula-unpack
 
-FILESEXTRAPATHS:append := "${THISDIR}/imx-gpu-viv:"
 SRC_URI = "${FSL_MIRROR}/${BPN}-${PV}-${IMX_SRCREV_ABBREV}.bin;fsl-eula=true"
 
 S = "${UNPACKDIR}/${BPN}-${PV}-${IMX_SRCREV_ABBREV}"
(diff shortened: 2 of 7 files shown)


otavio (approved):

recipes-security/optee-qoriq/optee-client.nxp.inc
```diff
@@ -4,6 +4,10 @@ require recipes-security/optee-imx/optee-client-fslc.inc
 
 # The patch same as imx-optee, so point FILESEXTRAPATHS to optee-imx/optee-client,
 # avoid duplicate copy files to optee-qoriq
+# The entry cannot be dropped: this recipe's BPN is optee-client-qoriq, so the
+# default FILESPATH only searches directories named after it, and the shared
+# i.MX directory that holds tee-supplicant.sh is unreachable without this line.
+# nooelint: oelint.vars.outofcontext
```
thochstein, line 12: There is another reason for doing this: it allows a user of the .inc in another layer to access any files added to SRC_URI by the .inc.

thochstein (approved):

PR #2664, opened 2026-09-03
oelint: Resolve oelint.var.licenseremotefile findings (10 -> 0)

lucianogdittgen:
## Summary

Fixes all 10 oelint.var.licenseremotefile findings in the layer. Each
recipe checksummed either oe-core's generic per-license text (which
never changes, so it cannot detect an upstream license change) or a
path with an unresolved variable (which a standalone analyzer cannot
follow). Every commit points LIC_FILES_CHKSUM at the real license text
the source ships, or fixes the path so it resolves.

11 commits, 11 STRUCTURAL FIX, 0 new suppressions:

- devil: Checksum the license the source ships
- gpp-aioptool: Checksum the license the source ships
- imx-boot: Checksum the license the source ships
- imx-mkimage: Checksum the license the source ships
- kernel-module-isp-vvcam: Drop the variable from the license path
- libxdg-shell: Drop the variables from the license path
- kernel-module-imx-gpu-viv: Checksum the license header in the source
- mesa-etnaviv-env: Checksum the license header in the shipped script
- udev-rules-imx: Checksum the license header in the shipped rules
- udev-rules-qoriq: Checksum the license header in the shipped rules
- uuu-bin: Checksum the real license files, not the generic text

Two old `# nooelint: oelint.var.licenseremotefile` suppressions were
removed and replaced with real fixes. One of them (imx-mkimage) had
suppressed the finding on a wrong claim -- the source does ship a
LICENSE file, it was just never referenced.

The 6 remaining inline suppressions for this rule (merge-files,
imx-atf, imx-test x2, packagegroup-imx-tools-audio,
fsl-eula-unpack.bbclass) were reviewed and are genuinely justified:
each recipe's source ships no license file of its own, so
LIC_FILES_CHKSUM correctly points at the shared common-licenses text
or a layer-provided EULA file.

## Test plan

- Every unit validated with `do_populate_lic` re-run plus a
  buildhistory `packages/` comparison: byte-identical, except where a
  commit explains an intended PKGSIZE change (an SPDX line now shipped
  in the package).
- Full-layer survey after the branch: oelint.var.licenseremotefile no
  longer appears; nothing else the branch touched regressed under a
  different rule.
- `git diff master..HEAD -U0 | grep nooelint:` -- zero new suppressions
  added, two removed.

### DIFF
--- a/recipes-bsp/imx-mkimage/imx-boot_1.0.bb
+++ b/recipes-bsp/imx-mkimage/imx-boot_1.0.bb
@@ -7,7 +7,7 @@ DESCRIPTION = "Generate Boot Loader for i.MX 8 device"
 HOMEPAGE = "https://github.com/nxp-imx/imx-mkimage"
 SECTION = "bsp"
 LICENSE = "GPL-2.0-only"
-LIC_FILES_CHKSUM = "file://${COREBASE}/meta/files/common-licenses/GPL-2.0-only;md5=801f80980d171dd6425610833a22dbe6"
+LIC_FILES_CHKSUM = "file://LICENSE;md5=59530bdf33659b29e73d4adb9f9f6552"
 
 DEPENDS:append = " xxd-native"
 DEPENDS:append:mx8m-generic-bsp = " u-boot-mkimage-native dtc-native u-boot-mkeficapsule-native"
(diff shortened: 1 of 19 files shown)


otavio (approved):

thochstein (approved):

otavio:
@lucianogdittgen Can you do the backport, please?

PR #2674, opened 2026-09-08
jailhouse-imx: Scope kernel-module PACKAGES_DYNAMIC to own module

zelan-nxp:
PACKAGES_DYNAMIC += "^kernel-module-.*" is too broad: it makes jailhouse-imx a candidate runtime provider for every kernel-module-* virtual package in the layer. Recipes such as iptables RRECOMMENDS kernel-module-* names (e.g. kernel-module-x-tables); via do_package_qa[rdeptask] = "do_packagedata" these resolve to a provider, and because BitBake sorts providers alphabetically jailhouse-imx is picked ahead of linux-imx. This pulls jailhouse-imx into the task graph of unrelated recipes and hence into core-image-minimal even when jailhouse is not installed.

Narrow the pattern to this recipe's own module so the vars.specific lint finding is still resolved without advertising foreign kernel-module-* names. After the change kernel-module-x-tables etc. fall back to linux-imx and 'oe-depends-dot -k jailhouse-imx' returns no path.

Verified on imx8mp-lpddr4-evk (mx8m-nxp-bsp) with bitbake -g core-image-minimal: the iptables do_package_qa/do_package_write_deb edges now point at linux-imx.do_packagedata and jailhouse no longer appears anywhere in task-depends.dot.


recipes-extended/jailhouse/jailhouse-imx_git.bb
```diff
@@ -122,8 +122,12 @@ INSANE_SKIP:${PN}-dbg = "buildpaths"
...
 # kernel-module-${KERNEL_VERSION} is generated by the kernel-module-split class
 # (inherit module, above); declare it so its INSANE_SKIP is recognised by
-# standalone lint. ${KERNEL_VERSION} is build-time only, so match the prefix.
-PACKAGES_DYNAMIC += "^kernel-module-.*"
+# standalone lint. Scope the pattern to this recipe's own module only: a broad
+# "^kernel-module-.*" makes jailhouse-imx a candidate provider for every
+# kernel-module-* virtual (e.g. iptables' RRECOMMENDS kernel-module-x-tables),
+# and since providers are sorted alphabetically it is picked ahead of linux-imx,
+# pulling jailhouse-imx into the task graph of unrelated recipes.
+PACKAGES_DYNAMIC += "^kernel-module-jailhouse.*"
```
thochstein, line 130: I don't understand this variable very well, but isn't there an obvious discrepancy between the original comment and this new implementation? The original comment says that `kernel-module-${KERNEL_VERSION}` needs to be declared, but now we are declaring `kernel-module-jailhouse.*`? Doesn't this change break the original oelint fix?

otavio: @lucianogdittgen Please give this a check.

lucianogdittgen: @thochstein Yes, good eye. The finding does come back, but it's fallout from my
#2650, not from this change. oelint keeps ${KERNEL_VERSION} unexpanded, so the
literal token kernel-module-${KERNEL_VERSION} matched the broad regex and doesn't
match the scoped one.

The scoping is the right call and should stay. What my broad pattern was hiding is a
stale name I should have caught: kernel-module-split builds names as
kernel-module-<module>-${KERNEL_VERSION}, so the package is really
kernel-module-jailhouse-${KERNEL_VERSION}, which is exactly what @zelan-nxp's
pattern matches. The old INSANE_SKIP never matched any package this recipe produces,
and the ERROR_QA:remove above it was the workaround for that.

I have a follow-up ready, validated on imx8mp-lpddr4-evk with cleansstate rebuilds.
do_package_qa now reports "Package kernel-module-jailhouse-6.18.24-... skipping QA
tests: {'buildpaths'}", so the skip finally applies and relaxing ERROR_QA for the
whole recipe is no longer needed. As a control, dropping the INSANE_SKIP fails with
"jailhouse.ko ... contains reference to TMPDIR [buildpaths]" on that package, so the
exception is still load-bearing. The change is output-neutral (zero buildhistory
delta), and oelint.vars.specific is back to 0 layer-wide with no suppression.

Sending it as a PR shortly.

PR #2678, opened 2026-09-08
oelint: Split the i.MX overrides into version-specific includes (var.order 43 -> 0)

lucianogdittgen:
Reworked per review. @thochstein asked for the i.MX overrides to stay
consolidated rather than interleaved into the copied body; @otavio preferred a
version-specific `.inc`. That is what this now is — the interleaving is gone.

Each fork recipe is once again a **verbatim copy** of its upstream original.
Its i.MX customization moved to a sibling `<recipe>.inc`, `require`d at exactly
the position the old `########## i.MX overrides ##########` block occupied.

### Why `.inc` and not `.bbappend`

Both were offered. A `require` at the block's former position preserves parse
order *by construction*, so effective metadata is identical with no per-recipe
argument needed — including for the `:remove` / `:prepend` / `+=` entries and
the `inherit use-imx-headers` that live inside the moved blocks. A `.bbappend`
parses after the whole recipe and would need that argument made case by case.

### What this buys at re-sync

Checking a fork against upstream is now a plain diff:

```
diff <(git -C openembedded-core show fc108ddb:meta/recipes-graphics/wayland/weston_14.0.2.bb) \
     recipes-graphics/wayland/weston_14.0.2.imx.bb
```

Each copy still records the exact `# Upstream hash:` it was taken from.

### `oelint.var.order`, measured against `7aa40fe03`: 43 → 0

| | count | |
|---|---|---|
| structural, outside the forks | 20 | imx-gpu-viv 10, mali-imx 5, qemu-qoriq 5 |
| cleared by the split | 8 | the offending pair straddled the copy/overrides boundary |
| structural, inside an i.MX block | 2 | gst-plugins-good, gst-plugins-bad — verbatim copies untouched |
| suppressed, UPSTREAM-PARITY | 13 | present verbatim in the upstream recipe at the recorded hash |

The 13 are upstream's own ordering — confirmed present in meta-oe
`opencv_4.13.0.bb`, OE-core `weston_15.0.0.bb`, OE-core
`gstreamer1.0-plugins-good_1.28.2.bb` and meta-oe `libcamera_0.6.0.bb`. Fixing
them in place would diverge the copy, so each carries an inline directive naming
where the order comes from. `oelint.file.inlinesuppress_na` will flag any of them
the moment upstream fixes it.

Full layer: **116 → 72**, 44 removed, **0 introduced**.

### One cost worth naming

oelint analyses per file, so the split loses a little cross-file visibility:
opencv's `oelint.task.multifragments` disappears because its two
`do_install:append` fragments now sit in different files. That is a reporting
artefact, not a fix.

### Validation

Every split is a **pure relocation** — strip comments, blanks and the new
`require`, and the result is byte-identical to the base:

```
diff <(git show upstream/master:$BB | grep -v '^[[:space:]]*#' | grep -v '^[[:space:]]*$') \
     <(cat $BB $INC | grep -v '^[[:space:]]*#' | grep -v '^[[:space:]]*$' | grep -vFx "require $(basename $INC)")
```

Each recipe was then built `-c package -f` and compared through buildhistory —
opencv and the gstreamer four on imx8mp-lpddr4-evk, weston 14.0.2 and libcamera
on imx8mm-lpddr4-evk, weston 10.0.5 on imx6qdlsabresd, qemu-qoriq on ls1046ardb.
**Every delta is empty.** For opencv the expanded datastore was compared as well:
2835 variables identical, the only change being `BBINCLUDED` gaining the new
`.inc`.

### Out of scope

`gst-devtools_1.28.1.bb` and `gstreamer1.0-rtsp-server_1.28.1.bb` have i.MX
blocks but no verbatim copy to protect, and no findings. The split exists to
protect a copy; they have none.

### Note for reviewers

Validating this on a current tree needs the build fixes on
`buildfix/wayland-protocols` — `wayland-protocols` 1.47 and `vulkan-loader`
1.4.341.0 both fail `do_patch` at `master` today, independently of this series.

### DIFF
--- a/dynamic-layers/openembedded-layer/recipes-support/opencv/opencv_4.13.0.imx.bb
+++ b/dynamic-layers/openembedded-layer/recipes-support/opencv/opencv_4.13.0.imx.bb
@@ -1,7 +1,7 @@
-# This recipe is modified for i.MX.
-# For ease of maintenance, the top section is a verbatim copy
-# of an OE-core recipe, and the second section customizes the
-# recipe for i.MX.
+# This recipe is modified for i.MX. For ease of maintenance the recipe itself
+# is a verbatim copy of a meta-openembedded recipe, so it can be diffed
+# against upstream directly; the i.MX customization lives in the required
+# opencv_4.13.0.imx.inc.
 
 ########## meta-openembedded copy ###########
 # Upstream hash: 9b77eae6988e98adbec5323d2491afc6b327c91a
@@ -12,7 +12,7 @@ HOMEPAGE = "http://opencv.org/"
 SECTION = "libs"
 
 LICENSE = "Apache-2.0"
-# Re-set in the i.MX overrides section below; kept here to preserve the
+# Re-set in the required opencv_4.13.0.imx.inc; kept here to preserve the
 # verbatim meta-openembedded copy (see header). UPSTREAM-PARITY.
 # nooelint: oelint.var.override
 LIC_FILES_CHKSUM = "file://LICENSE;md5=3b83ef96387f14655fc854ddc3c6bd57"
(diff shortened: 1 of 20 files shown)


dynamic-layers/openembedded-layer/recipes-support/opencv/opencv_4.13.0.imx.bb
```diff
@@ -1,20 +1,18 @@
...
+# This recipe is modified for i.MX. For ease of maintenance, the bulk of
+# the file is a copy of a meta-openembedded recipe. The i.MX overrides sit
+# beside the values they replace, so the file as a whole keeps the variable
+# order of the recipe style guide.
 
 ########## meta-openembedded copy ###########
 # Upstream hash: 9b77eae6988e98adbec5323d2491afc6b327c91a
 
 SUMMARY = "Opencv : The Open Computer Vision Library"
+SUMMARY:python3-opencv = "Python bindings to opencv"
```
thochstein: This is going to make things harder. With the separate section, there was never the question of whether a change like this line was added locally or removed upstream. Now there will be ambiguity that will make merging a new version difficult.

Can we use a .bbappend or a .inc to keep the overrides consolidated?

otavio: How do you envision doing that?

thochstein: Nothing fancy, just move the override section at the bottom of the file recipe to a .bbappend file or a .inc file.

otavio: so the inc seems to be better, don't you think? But we should keep it specific for the version.

thochstein: Maybe marginally.

lucianogdittgen: Done — the interleaving is reverted and each fork's i.MX overrides now live in a
version-specific sibling `.inc`, e.g. `opencv_4.13.0.imx.inc`. The `.bb` is once
again a verbatim copy of the upstream recipe, so checking it against the
`# Upstream hash:` in its header is a plain diff again.

On `.inc` vs `.bbappend`: `.inc` is the better of the two here for a concrete
reason. The `require` sits at exactly the position the old override block
occupied, so parse order is preserved *by construction* — no case-by-case
argument needed for the `:remove` / `:prepend` / `+=` entries or the
`inherit use-imx-headers` inside the moved blocks. A `.bbappend` parses after the
whole recipe, so each of those would need that argument made individually.

Every split is a pure relocation: strip comments, blank lines and the new
`require`, and the result is byte-identical to the current master. Each recipe
was then built `-c package -f` and compared through buildhistory — opencv and the
gstreamer four on imx8mp, weston 14.0.2 and libcamera on imx8mm, weston 10.0.5 on
imx6qdlsabresd. Every delta is empty.

Two things worth flagging rather than burying:

- 13 findings remain inside the copied bodies and are suppressed inline rather
  than fixed, because they are upstream's own ordering — present verbatim in
  meta-oe `opencv_4.13.0.bb`, oe-core `weston_15.0.0.bb`,
  `gstreamer1.0-plugins-good_1.28.2.bb` and meta-oe `libcamera_0.6.0.bb`. Fixing
  them in place would diverge the copies, which is the thing this rework exists
  to avoid. Happy to drop those commits if you would rather the copies carry no
  added comments at all.
- The split costs a little cross-file visibility: opencv's
  `oelint.task.multifragments` disappears only because its two
  `do_install:append` fragments now sit in separate files. That is a reporting
  artefact, not a fix.

`gst-devtools` and `gstreamer1.0-rtsp-server` also have i.MX blocks but no
verbatim copy to protect, so they are left alone.

otavio (approved):

thochstein (approved):

PR #2679, opened 2026-09-08
Remove or document the layer's anonymous python functions

lucianogdittgen:
## What this does

`oelint-adv` reports 9 `oelint.task.noanonpython` findings in this layer. This
series takes each one in turn. Where BitBake offers a declarative form, the
series removes the anonymous python. Where BitBake offers none, it documents the
block in place and suppresses the finding with the reason beside it.

The layer now reports 0 findings for this rule.

One packaging defect surfaced during the work. It is fixed in its own commit.

Following review, the Debian-rename migration metadata is dropped rather than
carried forward. See "Migration metadata removed" below for what went and what
stayed.

## Replaced with static metadata

| Commit | Recipe | Change |
| --- | --- | --- |
| `03fd74d` | `imx-gpu-viv-6.inc` | Both loops ran over a fixed set. `insane.bbclass` unions `INSANE_SKIP` with `INSANE_SKIP:<pkg>`, so one global `ldflags` replaces the per-package appends exactly. |
| `f8c58a0` | `mali-imx.inc` | The loop ran over a fixed set of five packages. The file already wrote these assignments statically for `${PN}-libgles-dev`. |
| `d9a0a1b` | `imx-boot-firmware-files` | A varflag takes inline python, so the SoC function names reach `do_deploy[vardeps]` without an anonymous python. |

## Migration metadata removed

`b8fc74e` and `9584ba2` drop the `RREPLACES` and `RCONFLICTS` entries for the
generic GL package names, and the Debian soname aliases (`libegl1`,
`libglesv1-cm1`, `libglesv2-2`, `libgbm1`) everywhere they appeared. These
existed so package feeds could upgrade deployed devices across the rename to the
`-imx` / `${PN}-` suffixes. That transition is long past.

Two things stayed, deliberately:

* **`DEBIAN_NOAUTONAME`.** Not migration metadata. It is what stops
  `debian.bbclass` renaming `libegl-imx` to `libegl1` and colliding with mesa.
* **`RPROVIDES` of the bare generic names** (`libegl`, `libgles1`, `libgles2`,
  `libgles3`, `libgbm`, and the `-dev` forms). These are a live provider
  interface, not an alias. `cogl-1.0` dlopens its GL backend, so it cannot pick
  the dependency up through shlibs and instead carries
  `ERDEPENDS_GLES2 = "libgles2"` (`meta-gnome/recipes-graphics/cogl/cogl-1.0.inc:30`),
  wired into `PACKAGECONFIG[gles2]`, which is on by default. On an i.MX GPU
  system that resolves only through `RPROVIDES:libgles2-imx`. oe-core's
  `mesa.inc` still declares the same triple over the same name pairs, so keeping
  the provides also keeps mesa's `RCONFLICTS` exclusion working against these
  packages.

## Reduced, with the remainder documented

| Commit | Recipe | Change |
| --- | --- | --- |
| `58b4236` | `qemu-qoriq` | `BPN` is always `qemu-qoriq` here, so `PROVIDES` needed no parse-time code and is now static. The per-package rename stays: inline python can compute a value, but it cannot compute a variable *name*, and the rename generates `RPROVIDES:<pkg>`. |

## Documented in place

BitBake has no declarative form for any of these. Each keeps its anonymous
python and carries the reason next to it.

| Commit | File | Why no declarative form exists |
| --- | --- | --- |
| `3123aad` | `imx-gpu-viv-6.inc` | The guard must skip the recipe during parse. `COMPATIBLE_MACHINE` cannot express it: `MACHINE_HAS_VIVANTE_KERNEL_DRIVER_SUPPORT` is a weak default over `SOC_HAS_VIVANTE_KERNEL_DRIVER_SUPPORT:imxgpu` that a machine conf may set on its own, so an `(imxgpu)` regex would ignore that choice. |
| `7b45fea` | `image_populate_mfgtool.bbclass` | The loop enumerates the varflags of `MFGTOOLCONFIG`, and it feeds `do_populate_mfgtool[depends]`, which must exist before the task graph. OE-core parses `PACKAGECONFIG` the same way. |
| `95c228d` | `linux-imx_6.18.bb` | A migration guard that aborts the parse when a machine still sets the removed `DELTA_KERNEL_DEFCONFIG`. It tests whether the key is defined at all, and it raises an error. Neither has a declarative form. |
| `cb22df8` | `qoriq_build_64bit_kernel.bbclass` | The block conditionally replaces `KERNEL_CC`, `KERNEL_LD` and `KERNEL_AR`. A declarative form would restate the `kernel.bbclass` defaults inside this class, and `DEPENDS` is read before the task graph exists. |
| `ace8929` | `u-boot-qoriq_2025.04.bb` | The block raises `SkipPackage`/`SkipRecipe` when the machine override needs a multilib cross toolchain that `MULTILIB_VARIANTS` does not build. Only a parse-time raise can drop a recipe, and `COMPATIBLE_MACHINE` matches machine tokens, not multilib configuration. |

## Packaging defect found along the way

`4fa9dbf` fixes a real bug that the Debian-rename loop hid.

The loop appended the `-dev` name with no separator. `RPROVIDES:libgles1-imx-dev`
is the only one of these with a prior value, `libgles-imx-dev`, so it resolved to
the single token `libgles-imx-devlibgles1-dev`. The package provided neither
name. The fix is committed before the rewrite, so the rewrite stays inert.

The same loop drops the `libgles3-imx` names. Only `libgles3-imx-dev` is a
package, so those names were never read.

## Validation

Every commit that changes effective metadata was built from a clean
`cleansstate` and compared with buildhistory. A green build alone was not
treated as evidence.

| Recipe | Build | Result |
| --- | --- | --- |
| `imx-gpu-viv` | `bitbake imx-gpu-viv` on `imx8mp-lpddr4-evk` | de-looping commit: buildhistory reports no change |
| `imx-gpu-viv` (defect fix) | same | the buildhistory delta is that one `RPROVIDES` |
| `imx-gpu-viv` (removal) | same | 10 files, 5 insertions, 14 deletions — every line an `RPROVIDES` or `RCONFLICTS`. No file lists, sizes, or package additions/removals. |
| `mali-imx` | `bitbake mali-imx` on `imx95-19x19-verdin` | de-looping commit: every value token-identical, buildhistory reports no change |
| `mali-imx` (removal) | same | 11 files, 5 insertions, 15 deletions — same shape. `${PN}-libgles-dev` untouched. |
| `imx-boot-firmware-files` | `bitbake imx-boot-firmware-files -c deploy -f` on `imx8mp-lpddr4-evk` | `do_deploy` hash unchanged, sigdata still records `deploy_for_mx8m` alone, deployed firmware byte-identical |
| `qemu-qoriq` | `bitbake qemu-qoriq -C package` on `ls1046ardb` | `bitbake -e` byte-identical, buildhistory reports no change |

The five documentation commits move comments only, so they cannot change the
generated output.

`mali-imx` cannot currently be built from clean in this tree: it `DEPENDS` on
`vulkan-loader` and `wayland-protocols`, and both fail `do_patch` because the
layer's downstream patches were cut against an older oe-core (documented in
`OELINT-TODO.md`). Its two builds therefore ran with `069e7cef1` and `99a2c4499`
from `fix-downstream-patches-oe-core-bump` cherry-picked onto **both** sides, so
baseline and candidate differ only by this series. Those two commits are not
part of this PR.

Full-layer survey, both sides scanned serially with an identical invocation:
**113 to 103 findings** (error 16 to 16, warning 74 to 65, info 23 to 22).
**10 removed, 0 introduced.** The tenth is an `oelint.task.multifragments`
finding on `imx-gpu-viv-6.inc`, which the same structural fix retired.


recipes-graphics/imx-gpu-viv/imx-gpu-viv-6.inc
```diff
@@ -432,3 +410,45 @@ FILES:libnn-imx += "${libdir}/libNN*${SOLIBS}"
...
+
+RCONFLICTS:libegl-imx += "libegl libegl1"
+RCONFLICTS:libegl-imx-dev += "libegl-dev"
+RCONFLICTS:libgles1-imx += "libgles1 libglesv1-cm1"
+RCONFLICTS:libgles1-imx-dev += "libgles1-dev"
+RCONFLICTS:libgles2-imx += "libgles2 libglesv2-2"
+RCONFLICTS:libgles2-imx-dev += "libgles2-dev"
+RCONFLICTS:libgles3-imx-dev += "libgles3-dev"
+RCONFLICTS:libgbm-imx += "libgbm"
+RCONFLICTS:libgbm-imx-dev += "libgbm-dev"
```
otavio: This has been maintained for several years, so I think we can remove those migration packages to reduce the metadata we need to port. @thochstein, do you agree?

recipes-graphics/mali/mali-imx.inc
```diff
@@ -90,30 +90,58 @@ RDEPENDS:${PN}-libgles3-dev = "${PN}-libgles-dev"
...
+RCONFLICTS:${PN}-libegl += "libegl libegl1"
+RCONFLICTS:${PN}-libegl-dev += "libegl-dev"
+RCONFLICTS:${PN}-libgbm += "libgbm libgbm1"
+RCONFLICTS:${PN}-libgbm-dev += "libgbm-dev"
+RCONFLICTS:${PN}-libgles1 += "libgles1 libglesv1-cm1"
+RCONFLICTS:${PN}-libgles1-dev += "libgles1-dev"
+RCONFLICTS:${PN}-libgles2 += "libgles2 libglesv2-2"
+RCONFLICTS:${PN}-libgles2-dev += "libgles2-dev"
+RCONFLICTS:${PN}-libgles3 += "libgles3"
+RCONFLICTS:${PN}-libgles3-dev += "libgles3-dev"
```
otavio: same.

recipes-graphics/imx-gpu-viv/imx-gpu-viv-6.inc
thochstein: I'll trust you, I may not understand the original intent. Did it have something to do with the debian naming?

otavio: Not really; it was about package feeds that could automatically upgrade packages on remote devices.

recipes-graphics/imx-gpu-viv/imx-gpu-viv-6.inc
```diff
@@ -432,3 +413,27 @@ FILES:libnn-imx += "${libdir}/libNN*${SOLIBS}"
 # nooelint: oelint.vars.insaneskip
 INSANE_SKIP:append:libc-musl = " file-rdeps"
 
+# The OpenGL interface packages keep their -imx suffix: Debian renaming would
+# drop it and collide with the generic mesa package names.
+
```
otavio:
```suggestion
```

otavio (changes requested):

PR #2680, opened 2026-09-09
jailhouse-imx: Fix kernel module package name in INSANE_SKIP

lucianogdittgen:
Follow-up to #2674, which correctly scoped `PACKAGES_DYNAMIC` to this recipe's
own module but left `oelint.vars.specific` firing again on the line below it.

## Cause

`kernel-module-split` names split packages `kernel-module-<module>-${KERNEL_VERSION}`,
so the real package here is `kernel-module-jailhouse-${KERNEL_VERSION}` (buildhistory
on imx8mp-lpddr4-evk: `kernel-module-jailhouse-6.18.24-lf-6.18.y-g1d08c2fc5b37`).

`INSANE_SKIP:kernel-module-${KERNEL_VERSION}` therefore never matched a package this
recipe produces. Two things followed from that:

* `ERROR_QA:remove = "buildpaths"` was added as a recipe-wide workaround, per the
  comment that the error "cannot be skipped with INSANE_SKIP".
* The broad `^kernel-module-.*` added in #2650 happened to cover the wrong literal
  token, so lint passed for the wrong reason. Scoping the pattern in #2674 removed
  that accident and exposed the stale name.

## Change

Name the real package. The skip then applies and the recipe-wide `ERROR_QA`
relaxation is no longer needed, so it goes too. The `PACKAGES_DYNAMIC` comment is
trimmed to the part a reader cannot derive: why it must not be broadened again.

## Verification

Three `cleansstate` rebuilds on imx8mp-lpddr4-evk (mx8m-nxp-bsp, fslc-framebuffer):

| State | Result |
| --- | --- |
| fixed `INSANE_SKIP`, `ERROR_QA:remove` present | passes |
| fixed `INSANE_SKIP`, `ERROR_QA:remove` removed | passes; `do_package_qa` logs `Package kernel-module-jailhouse-6.18.24-... skipping QA tests: {'buildpaths'}` |
| `INSANE_SKIP` removed, `ERROR_QA:remove` removed | fails: `jailhouse.ko ... contains reference to TMPDIR [buildpaths]` |

The third is the control: the exception is still load-bearing, and it is now the
`INSANE_SKIP` that carries it.

Buildhistory is byte-identical under `packages/` against the pre-change build.
Layer-wide `oelint.vars.specific` goes 1 → 0 with no suppression added
(full scan: removed 1, introduced 0).


otavio (approved):

thochstein:
Nice job fixing this, Luciano and Zelan.

Just wanted to mention that Jun is looking at splitting this recipe into two, one for userspace and one for module. Does it seem reasonable?

PR #2682, opened 2026-09-09
Fix oelint.vars.srcurimutualex in the i.MX fork recipes

lucianogdittgen:
## Summary

Clears `oelint.vars.srcurimutualex` across the i.MX fork recipes.

Every flagged recipe reports the same thing: `SRCREV` and `SRC_URI[sha256sum]` set for the
same item. The two never actually apply together — the i.MX overrides `SRC_URI:remove` the
release tarball and fetch the NXP git fork instead, so the checksum belongs to a URI that is
no longer in `SRC_URI`. oelint reports it because it does not evaluate `SRC_URI:remove`.

Note that the `.inc` split merged in #2678 does not clear this on its own: oelint parses the
`require`d `.inc` along with the `.bb`, so it still sees `SRCREV` and the checksum together.

## What changed

The unused checksum is annotated in place rather than deleted, so the copied OE-core body
stays byte-identical to upstream and remains diffable against it:

```bitbake
# The required .inc removes this tarball from SRC_URI and fetches the NXP git
# fork instead, so this checksum and SRCREV never apply to the same item.
# UPSTREAM-PARITY.
# nooelint: oelint.vars.srcurimutualex
SRC_URI[sha256sum] = "..."
```

One commit per recipe: `weston` (10.0.5 and 14.0.2 together), `gstreamer1.0`,
`gstreamer1.0-plugins-base`, `gstreamer1.0-plugins-bad`, `gstreamer1.0-plugins-good`.

## Validation

Comments only — no assignment is added, removed or changed, so generated output cannot move.
The full diff against master is six comment blocks and nothing else.

`oelint-adv` over the six recipes (each with its required `.inc`):

| | before | after |
|---|---|---|
| `oelint.vars.srcurimutualex` | 6 | 0 |
| all findings | 62 | 56 |

The rule is cleared with no collateral: the totals differ by exactly the six findings this
PR targets, so nothing else was masked or shifted.

## Note on scope

This branch originally also carried the `.inc` split for these recipes. That work landed
independently in #2678, so those commits were dropped in the rebase — upstream's versions
supersede them (identical for three recipes; for `-plugins-bad` and `-plugins-good` upstream
additionally reorders `PACKAGE_ARCH` / `PACKAGECONFIG`). What remains here is only the
parity annotation.

### DIFF
--- a/recipes-graphics/wayland/weston_10.0.5.imx.bb
+++ b/recipes-graphics/wayland/weston_10.0.5.imx.bb
@@ -23,6 +23,10 @@ SRC_URI = "https://gitlab.freedesktop.org/wayland/weston/-/releases/${PV}/downlo
            file://systemd-notify.weston-start \
            "
 
+# The required .inc removes this tarball from SRC_URI and fetches the NXP git
+# fork instead, so this checksum and SRCREV never apply to the same item.
+# UPSTREAM-PARITY.
+# nooelint: oelint.vars.srcurimutualex
 SRC_URI[sha256sum] = "89646ca0d9f8d413c2767e5c3828eaa3fa149c2a105b3729a6894fa7cf1549e7"
 
 UPSTREAM_CHECK_URI = "https://wayland.freedesktop.org/releases.html"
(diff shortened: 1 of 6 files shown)


otavio:
@lucianogdittgen rebase

PR #2713, opened 2026-09-17
Fixes for CVE_PRODUCT on optee and atf 

hiagofranco:
This fixes false-positives and missing CVEs on TFA and OPTEE recipes from meta-freescale.


hiagofranco:
I'd like to backport this to wrynose and scarthgap, once this is merged. Thanks!

PR #2730, opened 2026-09-18
testfloat: Drop the 2a pin and use meta-oe's TestFloat 3e

lucianogdittgen:
Removes `testfloat_2a`, which cannot build on any current toolchain, and drops the
`PREFERRED_VERSION` pin that was shadowing meta-openembedded's portable
`testfloat_3e`. Layer survey **9 -> 8**; clears `oelint.vars.downloadfilename`
(**1 finding -> 0**) as a side effect rather than as the goal. One commit.

This started as an oelint fix for that finding. The finding's recipe turned out to
be dead code, so the honest fix was to delete it.

### `testfloat_2a` has not been buildable for years

`recipes-extended/testfloat/files/TestFloat-powerpc-E500v2-SPE-1.patch` generates
sources against a toolchain that no longer exists. Measured against the actual
cross compiler in the build (**GCC 15.3.0**):

    __builtin_spe_mfspefscr()   error: implicit declaration of function
                                '__builtin_spe_mfspefscr'; did you mean
                                '__builtin_vec_mfvscr'?      (SPE dropped in GCC 9)

    -mfloat-gprs=double         error: unrecognized command-line option

    #include <bits/nan.h>       fatal error: No such file or directory
                                                       (dropped in glibc 2.26)

    main( int argc, ... )       error: return type defaults to 'int'
                                                       (an error since GCC 14)

oe-core already reflects the SPE removal — `tune-ppce500v2.inc` has

    TUNE_FEATURES:tune-ppce500v2 = "m32 ppce500v2 bigendian"

with no `spe`, so the `-mabi=spe -mspe -mfloat-gprs=double` line in that file is
dead metadata that now fails outright if selected. Plain `-mcpu=8548 -O2` on
`float add(float,float)` emits `fadds`, the classic-FPU instruction e500v2 has no
hardware for. There is no spelling of this recipe's SPE test that GCC 15 can still
express.

### The pin reached further than the recipe could

`conf/machine/include/qoriq-base.inc` is included by `qoriq-arm.inc`,
`qoriq-arm64.inc` **and** `qoriq-ppc.inc`, so

    PREFERRED_VERSION_testfloat = "2a"

applied to every QorIQ machine. But `testfloat_2a` is `COMPATIBLE_HOST:e500v2`
only, with `COMPATIBLE_HOST ?= "(none)"` elsewhere — so on QorIQ ARM and ARM64 the
pin selected a PowerPC-only recipe that could never build there, while blocking
meta-oe's `testfloat_3e`, which can. `testfloat` is in `packagegroup-meta-oe`.

### What replaces it

meta-openembedded's `meta-oe/recipes-test/testfloat/testfloat_3e.bb` — Berkeley
TestFloat 3e, BSD-3-Clause, architecture-neutral, already in the layer set.

### Validation

`build-oelint-p2020`, MACHINE `p2020rdb` (ppce500v2), DISTRO `fslc-framebuffer` —
the only machine class `testfloat_2a` could ever have built on. MACHINE, DISTRO and
layer set are identical on both sides.

Resolution proof, `bitbake -e testfloat` on the candidate:

    FILE="…/meta-openembedded/meta-oe/recipes-test/testfloat/testfloat_3e.bb"
    PN="testfloat"
    PV="3e"

- **Baseline**: buildhistory tagged at `9844203` — **no testfloat entry at all**,
  consistent with 2a never having built here. A baseline `bitbake testfloat` fails
  in `do_compile` with the errors above.
- **Candidate**: **PASS**, `testfloat-3e-r0 do_populate_sysroot: Succeeded`.
- **Buildhistory delta**: 7 packages appear (`testfloat`, `-dbg`, `-dev`, `-doc`,
  `-locale`, `-src`, `-staticdev`). The main package ships **five** binaries —
  `testfloat`, `testfloat_gen`, `testfloat_ver`, `testsoftfloat`, `timesoftfloat` —
  against the two `testfloat_2a` claimed to install. `RDEPENDS = glibc`,
  `LICENSE = BSD-3-Clause`. Incidentally pulls `unzip-native` (3e fetches `.zip`).
  No packages removed, no regressions.

This is not a behaviour-preserving cleanup, so the predicted delta was not "nothing"
but "testfloat starts existing" — which is what was measured.

### Not included

`meta-freescale-baseline` carries its own identical copy of both the pin and
`recipes-extended/testfloat/testfloat_2a.bb`. It is a separate repository and needs
the same change in its own PR.

Counts are from `contrib/oelint/run-oelint.sh --mode all`.


thochstein (commented):
I'm no expert here. @zelan-nxp, can you ask the QorIQ owners for their input?

zelan-nxp:
It's ok to remove that. NXP LF bsp has removed testfloat since 2018, Even more added testfloat at the begining for e500v, I assume we didn't support them now.

